PR #42, opened 2026-04-14
docs: refine README and release polish

NikolayS:
## Summary
- refine README positioning and keep the front page more useful
- split some long-form material into docs/ while keeping core value visible
- clarify batch semantics and remove weak/overstated claims
- polish packaging, grants, and tests for the modern API layer

## Included follow-up fixes
- assemble roles after public API functions so install works cleanly
- remove low-value roadmap page
- add/shared public type definition cleanup
- add tests for partial batch ack semantics and modern nack->retry cycle

## Notes
- benchmark material is marked as quick'n'dirty preliminary
- managed Postgres/platform positioning kept explicit
- this supersedes the earlier fragmented doc PR state


NikolayS:
Closing this branch/PR because the history got messy during consolidation/rebase. Replacing it with a fresh clean PR from current main.

PR #46, opened 2026-04-15
fix: xid8 type mismatch breaks rotation and batch event SQL

NikolayS:
## Problem

Table rotation never completed during benchmarks. Event tables grew to 185 GiB
on AWS i4g.2xlarge. Throughput dropped from 44k to 29k ev/s.

Original PgQ did NOT have this issue.

## Root cause

`build/transform.sh` converted `queue_switch_step1/step2` from `bigint` to
`xid8`, but PostgreSQL lacks operators between `xid8` and `bigint`/`integer`:

```sql
SELECT 1::int8 <= pg_current_xact_id();
-- ERROR: operator does not exist: bigint <= xid8

SELECT pg_current_xact_id() - 100;
-- ERROR: operator does not exist: xid8 - integer
```

This breaks `maint_rotate_tables_step1` (consumer check comparison) and
`batch_event_sql` (txid arithmetic for batch range optimization).

PgQ works because `txid_current()` returns `bigint` — everything is bigint.

## Fix

All changes in `build/transform.sh` only — no PgQ source modifications:

- `queue_switch_step1/step2`: keep as **bigint** (preserves PgQ arithmetic)
- `ev_txid`: keep as **xid8** (required by `pg_visible_in_snapshot()`)
- `pg_current_xact_id()`: cast to `::text::bigint` in PL/pgSQL code
- `pg_snapshot_xmin/xmax/xip()`: wrap with `::text::bigint` (return xid8, PgQ expects bigint)
- Remove xid8 cast patches in batch_event_sql (not needed with bigint columns)
- Fix `sed -i` for cross-platform (macOS BSD + Linux GNU)

Also adds:

- Clear section boundaries in `pgque.sql` — sections 1-5 marked "(derived from PgQ)" with origin file, sections 6-7 marked "(NEW)"
- Inline `-- PgQue transformation:` comments at each change point in the output SQL

## Verified

- Rotation works: `queue_cur_table` advances 0 → 1 after rotation period
- Build passes on macOS (BSD sed) 
- All assembly checks pass

## Test plan

- [x] Local rotation test (create queue → insert → ticker → maint step1/step2 → verify cur_table advances)
- [x] Build on macOS
- [x] CI on Linux (this PR)
- [x] Rerun 30-min benchmark with rotation working

### DIFF
--- a/build/transform.sh
+++ b/build/transform.sh
@@ -122,12 +122,15 @@ apply_txid_function_renames() {
   # Replace legacy txid_* functions with PG14+ equivalents.
   local content="$1"
 
+  # Rename txid_* functions to pg_* equivalents.
+  # pg_snapshot_xmin/xmax return xid8 (not bigint like the old txid_ versions).
+  # Wrap them with ::text::bigint to preserve PgQ's bigint arithmetic.
   content=$(echo "$content" | sed \
     -e 's/txid_current_snapshot()/pg_current_snapshot()/g' \
     -e 's/txid_current()/pg_current_xact_id()/g' \
-    -e 's/txid_snapshot_xmax(/pg_snapshot_xmax(/g' \
-    -e 's/txid_snapshot_xmin(/pg_snapshot_xmin(/g' \
-    -e 's/txid_snapshot_xip(/pg_snapshot_xip(/g' \
+    -e 's/txid_snapshot_xmax(\([^)]*\))/pg_snapshot_xmax(\1)::text::bigint/g' \
+    -e 's/txid_snapshot_xmin(\([^)]*\))/pg_snapshot_xmin(\1)::text::bigint/g' \
+    -e 's/txid_snapshot_xip(\([^)]*\))/pg_snapshot_xip(\1)/g' \
     -e 's/txid_visible_in_snapshot(/pg_visible_in_snapshot(/g')
 
   echo "$content"
@@ -141,21 +144,37 @@ apply_txid_snapshot_type_rename() {
 }
 
 apply_bigint_to_xid8() {
-  # Replace bigint -> xid8 ONLY for txid-related columns:
-  #   queue_switch_step1, queue_switch_step2, ev_txid
-  # These are the columns whose defaults reference txid_current()
-  # (now pg_current_xact_id() after the function rename).
+  # DISABLED: Keep columns as bigint (same as PgQ) instead of converting to xid8.
+  #
+  # Rationale: pg_current_xact_id() returns xid8, but xid8 lacks comparison
+  # and arithmetic operators that PgQ's rotation/batching code relies on:
+  #   - int8 <= xid8  (no operator — breaks maint_rotate_tables_step1)
+  #   - xid8 - integer (no operator — breaks batch_event_sql)
+  #   - xid8 -> int8   (no implicit cast — breaks variable assignment)
+  #
+  # Instead, we add ::text::bigint casts where pg_current_xact_id() is used
+  # as a column default. The pg_snapshot_xmin/xmax functions return xid8 but
+  # PL/pgSQL record fields auto-cast to the comparison type, and the batch
+  # SQL is built as text with ::text coercions.
+  #
+  # This preserves ALL PgQ's arithmetic and comparison logic unchanged.
   local content="$1"
 
-  # queue_switch_step1 and queue_switch_step2 column definitions
+  # Keep ev_txid as xid8 (required for pg_visible_in_snapshot).
   content=$(echo "$content" | sed -E \
-    's/(queue_switch_step1[[:space:]]+)bigint/\1xid8/g')
-  content=$(echo "$content" | sed -E \
-    's/(queue_switch_step2[[:space:]]+)bigint/\1xid8/g')
+    's/(ev_txid[[:space:]]+)bigint([[:space:]]+not null default)/\1xid8\2/g')
 
-  # ev_txid column definition in event_template
+  # Cast pg_current_xact_id() to ::text::bigint ONLY in switch_step columns
+  # and PL/pgSQL code (comparisons, assignments). ev_txid default stays as
+  # raw xid8 since the column is xid8.
+  # Strategy: cast ALL calls, then UN-cast the ev_txid default.
   content=$(echo "$content" | sed -E \
-    's/(ev_txid[[:space:]]+)bigint/\1xid8/g')
+    's/pg_current_xact_id\(\)([^:])/pg_current_xact_id()::text::bigint\1/g' \
+    | sed -E 's/pg_current_xact_id\(\)$/pg_current_xact_id()::text::bigint/g')
+
+  # Undo the cast for ev_txid default (needs raw xid8, not bigint):
+  content=$(echo "$content" | sed -E \
+    's/(ev_txid[[:space:]]+xid8[[:space:]]+not null default )pg_current_xact_id\(\)::text::bigint/\1pg_current_xact_id()/g')
 
   echo "$content"
 }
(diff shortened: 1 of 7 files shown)


NikolayS:
## REV Code Review Report

- **PR:** NikolayS/pgque#46 — fix: xid8 type mismatch breaks rotation and batch event SQL
- **Author:** @NikolayS
- **AI-Assisted:** Yes (Claude, Co-Authored-By)
- **CI Status:** ✅ All 5/5 checks pass (test PG14-17 + verify)

---

### BLOCKING ISSUES (2)

Issues that must be addressed before merge (high-confidence CRITICAL, HIGH, MEDIUM severity).

**HIGH** `sql/experimental/delayed.sql` — maint() override does NOT skip step2, defeating the rotation fix
> The core fix in `maint.sql` adds `if f.func_name = 'pgque.maint_rotate_tables_step2' then continue` to prevent step2 from running inside the maint transaction. But `delayed.sql` provides its own `CREATE OR REPLACE FUNCTION pgque.maint()` that iterates the same `pgque.maint_operations()` loop and only skips `vacuum` — it does **not** skip step2. Since `delayed.sql` replaces the function definition entirely, when the delayed module is loaded, step2 runs inside the maint transaction again — exactly the condition this PR is supposed to fix.
> 
> This was independently flagged by 3 review agents (security, bugs, docs).
> 
> **Fix:** Add the same step2 skip guard to `delayed.sql`:
> ```sql
> if f.func_name = 'pgque.maint_rotate_tables_step2' then
>     continue;
> elsif f.func_name = 'vacuum' then
>     continue;
> ```

**MEDIUM** No automated regression test for the critical rotation fix
> This PR fixes a production bug (185 GiB unbounded table growth) but adds no automated test for rotation. The test plan is manual only. If the step2-skip logic regresses (e.g., `func_name` from `maint_operations()` changes), rotation silently breaks again with no CI signal.
> **Fix:** Add a test that creates a queue, runs ticker + maint step1 in one TX, step2 in another, and verifies `queue_cur_table` advances. This is the exact scenario that caused the 185 GiB growth.

---

### NON-BLOCKING (1)

**INFO** Commit messages — All 3 subjects exceed 50-char guideline (59, 53, 52 chars)
> **Suggestion:** "fix: xid8 type mismatch in rotation SQL" (39), "fix: run rotation step2 in separate TX" (38), "fix: handle missing step2 job in stop()" (39)

---

### POTENTIAL ISSUES (4)

Issues with moderate confidence (4-7/10). Review manually — may be false positives.

**MEDIUM** `sql/pgque-api/maint.sql` — `f.func_name` not quoted in dynamic SQL execution *(confidence: 7/10)*
> ```sql
> execute 'select ' || f.func_name || '(' || quote_literal(f.func_arg) || ')' into r;
> ```
> `f.func_arg` is protected by `quote_literal()`, but `f.func_name` is concatenated without `quote_ident()` or allowlist validation. Combined with `security definer`, if the source table (`maint_operations()`) is writable by non-owners, this is a privilege escalation path.
> **Suggestion:** Add `quote_ident()` around `f.func_name`, or validate against an allowlist. This is pre-existing code but worth hardening.

**MEDIUM** `sql/experimental/delayed.sql` — Comment says "rotation step1/step2 changes" would be rolled back, but only vacuum is skipped *(confidence: 6/10)*
> The IMPORTANT comment implies both step1 and step2 are protected, but the function only skips vacuum. Misleading for future maintainers.
> **Suggestion:** Update comment to clarify scope, or (better) also skip step2 here per the BLOCKING fix above.

**LOW** `build/transform.sh` — `::text::bigint` cast will hard-fail at xid8 > 2^63 *(confidence: 5/10)*
> `xid8` is unsigned 64-bit, `bigint` is signed 64-bit. Values exceeding 2^63-1 will throw a runtime error. Unreachable in practice but a latent failure mode.
> **Suggestion:** Document the assumption. Consider whether PgQ's bigint arithmetic can eventually be restructured to use xid8 natively.

**LOW** `sql/pgque-additions/lifecycle.sql` — `exception when others then null` in stop() hides real errors *(confidence: 5/10)*
> Intended to handle "job doesn't exist" but catches permission errors, connection failures, etc.
> **Suggestion:** Catch specific exception (e.g., `undefined_object`) or at minimum `raise notice` on unexpected errors.

---

### Summary

| CI/Pipeline | 0 | 0 | 0 |
| Security | 0 | 2 | 0 |
| Bugs | 1 | 0 | 0 |
| Tests | 1 | 0 | 0 |
| Guidelines | 1 | 0 | 0 |
| Docs | 0 | 2 | 0 |

- **Findings**: High-confidence issues (8-10/10) — blocking or non-blocking per severity
- **Potential**: Medium-confidence issues (4-7/10) — review manually, may be false positives
- **Filtered**: Low-confidence issues (0-3/10) — excluded as likely false positives

> **Note:** The xid8 fix in `build/transform.sh` and the rotation step2 separation in `lifecycle.sql` are well-designed. The core approach is sound — the blocking issue is purely about the `delayed.sql` override not being updated to match.

---

NikolayS:
## REV Code Review Report

- **PR:** NikolayS/pgque#46 — fix: xid8 type mismatch breaks rotation and batch event SQL
- **Author:** NikolayS
- **CI Status:** ✅ All checks pass (test PG14-17, verify)

---

### BLOCKING ISSUES (0)

No blocking issues found. The core fix is correct: keeping `queue_switch_step1/step2` as `bigint` and casting `pg_current_xact_id()::text::bigint` resolves the operator-not-found errors without breaking any of PgQ's arithmetic.

---

### POTENTIAL ISSUES (4)

**MEDIUM** `sql/pgque-additions/lifecycle.sql:~87` — `exception when others` in `stop()` is too broad *(confidence: 6/10)*
> ```sql
> exception when others then
>     raise notice 'pgque.stop: rotate_step2 job not found (OK on first install)';
> end;
> ```
> Catches ALL exceptions including pg_cron connection errors, permission errors, internal panics — not just "job not found." A real failure would be silently downgraded to a NOTICE.
> **Suggestion:** Use a specific exception class: `exception when undefined_object or no_data_found` (or check `cron.job` table existence before calling `cron.unschedule`).

**MEDIUM** `sql/pgque-additions/lifecycle.sql:~37` — `step2_job_id` not persisted in config *(confidence: 5/10)*
> `start()` calls `cron.schedule_in_database(...)` and stores the result in `v_step2_id`, but never writes it to `pgque.config`. `stop()` then unschedules by name (`'pgque_rotate_step2'`) rather than by ID. This is inconsistent with `ticker_job_id`/`maint_job_id` and means: (a) the job ID is unrecoverable if the session crashes after scheduling, (b) any future `ALTER CRON JOB` or admin queries that join on job IDs won't find `step2`.
> **Suggestion:** Either add a `step2_job_id` column to `pgque.config` and persist it, or document the intentional name-based approach with a comment.

**LOW** `tests/test_core_rotation.sql:~36` — test does not isolate rotation failures *(confidence: 5/10)*
> If the `FAIL: expected queue_cur_table = 1` assertion in rotation 1's `step2` block raises an exception, the test aborts and the cleanup `drop_queue('test_rotation')` in the final DO block never runs. Subsequent test runs on the same database will find the stale queue and fail setup.
> **Suggestion:** Add `BEGIN; ... ROLLBACK;` wrapper around the full test, or add a `DO $$ BEGIN perform pgque.drop_queue('test_rotation'); EXCEPTION WHEN others THEN null; END; $$;` guard at the top as a pre-cleanup step.

**LOW** `sql/experimental/delayed.sql:~89` vs `sql/pgque-api/maint.sql:~18` — VACUUM behavior diverges between the two `maint()` implementations *(confidence: 4/10)*
> Non-experimental `pgque-api/maint.sql`: skips step2, still runs VACUUM.
> Experimental `delayed.sql`: skips BOTH step2 AND VACUUM (with rationale: autovacuum handles it).
> Users switching between the two get different VACUUM semantics. If the rationale is correct, the non-experimental version should also skip VACUUM. If not, the experimental one is regressing.
> **Suggestion:** Align the two implementations or add a clear comment in `maint.sql` explaining why it retains VACUUM.

---

### Summary

| CI/Pipeline | 0 | 0 | 0 |
| Security | 0 | 0 | 0 |
| Bugs | 0 | 2 | 0 |
| Tests | 0 | 1 | 0 |
| Guidelines | 0 | 1 | 0 |
| Docs | 0 | 0 | 0 |

**Notes on the fix correctness:**
- `::text::bigint` cast for `pg_current_xact_id()` is safe — xid8 values fit in int8 for any realistic transaction count.
- `pg_snapshot_xip()` returning `xid8` is fine in context — used as a table function in SQL JOINs (not PL/pgSQL bigint variable assignment); CI confirms no runtime errors.
- The "cast-all, then un-cast ev_txid" sed strategy is fragile but deterministic for the known PgQ source format.
- `sedi()` cross-platform fix is correct: `sed -i ''` (BSD) vs `sed -i` (GNU).
- awk-based ticker patching is more portable than the previous `sed -i` multiline approach.

---

**Overall: PASSED** — Root cause analysis is solid, fix is correct, CI passes on all supported PG versions.

*REV-assisted review (AI analysis)*

PR #52, opened 2026-04-17
docs(readme): clarify example client libraries

NikolayS:
## Summary
- clarify that Go, Python, and TypeScript client libraries are examples for now
- note that they are unpublished and still evolving
- explicitly invite community help improving them

## Why
This sets expectations correctly for early client libraries without overselling SDK maturity.


NikolayS:
Superseded by #53, which includes the same README clarification plus the actual TypeScript example client implementation.

PR #55, opened 2026-04-17
feat: add text overloads for pgque.send() and pgque.send_batch()

NikolayS:
## Summary

- Add `text` overloads alongside the existing `jsonb` overloads for `pgque.send(queue, payload)`, `pgque.send(queue, type, payload)`, and `pgque.send_batch(queue, type, payloads[])`.
- The jsonb overloads stay the ergonomic default (PG validates JSON at parse time, clear error on malformed payloads). The text overloads skip the jsonb parse + canonical reserialization round-trip, store bytes verbatim (preserving key order), and unlock non-JSON encodings: protobuf, msgpack, Avro, XML.
- Storage is unchanged — `ev_data` is still `TEXT`. No changes to rotation, triggers, or any PgQ engine internals.
- Resolves the send/receive type-asymmetry concern: a symmetric `text`-in / `text`-out path is now available without forcing `jsonb` on consumers.

## Changes

- `sql/pgque-api/send.sql`: add text overloads for `send` (2-arg + 3-arg) and `send_batch`.
- `sql/pgque.sql`: mirror the same additions in the assembled single-file install.
- `blueprints/SPECx.md` §4.1 + mapping table: document both overloads and the rationale.
- `README.md` function reference: note both overloads.
- `tests/test_api_send.sql`: new fast-path cases (default type, explicit type with non-JSON payload, `send_batch(text[])`).
- `tests/test_api_receive.sql`: round-trip through ticker + receive to assert byte-for-byte verbatim preservation (unsorted JSON keys stay unsorted; binary bytes stay binary).

## Test plan

- [x] `test_api_send.sql` — fast-path overloads return event ids, accept non-JSON bytes
- [x] `test_api_receive.sql` — send(text) → ticker → receive preserves payload byte-for-byte (verifies no jsonb canonicalization sneaks in)
- [x] Full `tests/run_all.sql` suite — all 16 files pass on PG 16
- [ ] CI to confirm on PG 14, 15, 17, 18

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,7 +12,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        pg_version: ['14', '15', '16', '17', '18']
+        pg_version: ['18', '17', '16', '15', '14']
 
     steps:
       - uses: actions/checkout@v4
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,7 +12,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        pg_version: ['14', '15', '16', '17']
+        pg_version: ['14', '15', '16', '17', '18']
 
     steps:
       - uses: actions/checkout@v4
--- a/README.md
+++ b/README.md
@@ -353,9 +353,9 @@ select pgque.ack(batch_id);
 
 | Function | Returns | Description |
 |---|---|---|
-| `pgque.send(queue, payload)` | `bigint` | Send with default type; `payload` may be `jsonb` (validated, canonicalized) or `text` (fast path, opaque bytes) |
-| `pgque.send(queue, type, payload)` | `bigint` | Send with explicit event type; `jsonb` and `text` overloads |
-| `pgque.send_batch(queue, type, payloads[])` | `bigint[]` | Batch send in a single transaction; `jsonb[]` and `text[]` overloads |
+| `pgque.send(queue, payload)` | `bigint` | Send with default type. Untyped literals resolve to the `text` overload (fast path, bytes verbatim); use `::jsonb` cast to opt into validation + canonicalization |
+| `pgque.send(queue, type, payload)` | `bigint` | Send with explicit event type. Same `text` / `jsonb` overload rules as above |
+| `pgque.send_batch(queue, type, payloads[])` | `bigint[]` | Batch send in a single transaction. `text[]` by default, `jsonb[]` opt-in via cast |
 | `pgque.send_at(queue, type, payload, deliver_at)` | `bigint` | Delayed/scheduled delivery |
 
 ### Consuming
--- a/README.md
+++ b/README.md
@@ -124,7 +124,6 @@ sustained load this causes documented, reproducible production failures:
   autovacuum starvation documented at Heroku
 - Oban Pro shipped table partitioning specifically to mitigate bloat
 - PGMQ ships aggressive autovacuum settings in some deployment setups
-  image
 
 PgQue's TRUNCATE rotation creates zero dead tuples in event tables by
 construction. No tuning required, immune to xmin horizon pinning. This
@@ -195,6 +194,40 @@ To uninstall:
 \i sql/pgque_uninstall.sql
 ```
 
+## Roles and grants
+
+The install creates three roles with explicit grants for the operational
+API. Application users do not need superuser; grant them whichever role
+matches their access pattern.
+
+| Role | Purpose | Granted access |
+|---|---|---|
+| `pgque_reader` | Dashboards, metrics, debugging | `get_queue_info`, `get_consumer_info`, `get_batch_info`, `version`, plus `select` on all tables |
+| `pgque_writer` | Producers and consumers (most apps) | inherits `pgque_reader` + the modern API (`send`, `send_batch`, `subscribe`, `unsubscribe`, `receive`, `ack`, `nack`) and the underlying PgQ primitives (`insert_event`, `next_batch`, `get_batch_events`, `finish_batch`, `event_retry`, `register_consumer`, `unregister_consumer`) |
+| `pgque_admin`  | Operators, migrations | inherits `pgque_writer` + full schema/table/sequence access. Excludes `uninstall()` (superuser-only on purpose: it drops the schema). |
+
+Typical app setup:
+
+```sql
+-- once, as superuser or schema owner
+\i sql/pgque.sql
+select pgque.start();           -- optional: pg_cron ticker + maint
+
+-- for the application connection
+create user app_orders with password '...';
+grant pgque_writer to app_orders;
+
+-- for a read-only metrics user
+create user metrics with password '...';
+grant pgque_reader to metrics;
+```
+
+DDL-class operations (`create_queue`, `drop_queue`, `start`, `stop`,
+`maint`, `ticker`, `force_tick`) are not granted to `pgque_writer` and
+should be performed by an admin / migration role. They run on PUBLIC
+default for now; revoking from PUBLIC and granting only to
+`pgque_admin` is on the roadmap.
+
 ## Project status
 
 PgQue is **early-stage** as a product and API layer.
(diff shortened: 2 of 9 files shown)


NikolayS:
@claude review

NikolayS:
@claude review again (use approach from https://gitlab.com/postgres-ai/rev; skip SOC2 stuff)

NikolayS:
## REV Code Review — #55

**Summary:** Adds `text` overloads for `pgque.send()` and `pgque.send_batch()` so callers can skip the jsonb parse/canonicalize round-trip and push non-JSON payloads. Untyped string literals now resolve to the new `text` path; `::jsonb` cast opts into validation. Also adds PG18 to CI and fixes a subtle bug in `send_batch` empty-array handling.

---

### BLOCKING — 0

No blockers. Logic looks correct and internally consistent.

### NON-BLOCKING

**MEDIUM** `blueprints/SPECx.md` / `README.md` — "protobuf, msgpack, Avro" claim is misleading for binary encodings
> PostgreSQL `text` **cannot store NUL (`\x00`) bytes** — it rejects them at input. Real protobuf/msgpack/Avro wire payloads routinely contain NULs (varints with high bit clear, tag 0, zero fields). The test fixtures sidestep this by using *textual* backslash-escape strings (`E'\\x01\\x02opaque'`), which are just literal `\`, `x`, `0`, `1` characters — not binary.
> **Fix:** Either (a) add an explicit note under §4.1 that the `text` overload is for textual payloads (JSON/XML/base64/hex-encoded binary); true binary requires `bytea` or caller-side encoding, or (b) add a `bytea` overload. Otherwise the first user who pipes raw protobuf in will get a confusing `invalid byte sequence` error.

**LOW** `sql/pgque-api/send.sql` / `sql/pgque.sql` — new `send(text, text)` / `send(text, text, text)` / `send_batch(text, text, text[])` overloads have no explicit `grant execute … to pgque_writer`
> The jsonb overloads don't either — they ride on PUBLIC default execute. Not a regression, but `sql/pgque-additions/roles.sql` grants every other writer-facing primitive explicitly (`insert_event`, `next_batch`, `get_batch_events`, …). The `send*` family is the odd one out, and `tests/test_pgque_roles.sql:13` only asserts `send(text, jsonb)`.
> **Fix:** add explicit grants for all four `send*` overloads + `subscribe`/`unsubscribe`, then revoke from `public`. Extend `test_pgque_roles.sql` to cover the text overloads.

**LOW** `tests/test_api_send.sql` Test 3a uses `'{"raw":"text"}'::text` to force the text overload
> Fine for the test's intent, but it doesn't exercise the headline claim that *untyped* literals resolve to `send(text, text)`. A reader will want one assertion that `select pgque.send(q, '{"k":1}')` (no cast) actually picks the text path — e.g. send a non-canonicalizable payload like `'{"b":2,"a":1}'` without a cast and assert key order is preserved on receive.
> **Fix:** add a no-cast case in `test_api_send.sql` or extend the round-trip test in `test_api_receive.sql` step 7 (currently both sends there use explicit `::text`).

**INFO** `sql/pgque-api/send.sql` / `sql/pgque.sql` — nice bug fix worth calling out
> `ids bigint[] := '{}';` initialization fixes a pre-existing bug: `send_batch(_, _, array[]::jsonb[])` previously returned `NULL` (uninitialized array), not an empty array. Test 3d covers both the jsonb and text empty-array paths. Good catch.

**INFO** `.github/workflows/ci.yml` — matrix reorder to `['18','17','16','15','14']`
> Adds PG18. Descending order is unusual but harmless; `fail-fast: false` means ordering doesn't affect results. Fine as-is.

### POTENTIAL ISSUES

**MEDIUM** `sql/pgque-api/send.sql` / `sql/pgque.sql` — `send_batch(jsonb[])` still does `p jsonb` + `p::text` per row
> Minor, but the PR description promotes `send_batch(text[])` as a fast path. If a high-volume caller already has `jsonb` in hand and batches 1000 items, the per-row `jsonb_out` cost is real. Not a regression — just noting the text batch path will look disproportionately faster in any benchmark. Probably fine; flagging so you don't get surprised by microbench numbers.

**LOW** `blueprints/SPECx.md` — protobuf example `E'\\x08\\x2a\\x10\\x63'`
> Same NUL caveat as above. Also worth noting that a real protobuf `0x08 0x2a 0x10 0x63` is 4 bytes; the `E'...'` literal here is 16 characters of backslash-escape text. Readers copying this as a template for "send real protobuf" will hit NUL errors quickly. See fix for the medium finding above.

---

### Summary

|------|----------|-----------|
| Security | 0 | 0 |
| Bugs | 0 (1 fix acknowledged) | 0 |
| Tests | 1 | 0 |
| Guidelines/Grants | 1 | 0 |
| Docs | 1 | 2 |

**Result: APPROVE with docs fix recommended.** The NUL-byte limitation on the `text` overload needs to be called out in SPECx.md §4.1 and the README before users try to shove raw binary through it. Everything else is green — the overload resolution reasoning is correct, the tests cover the verbatim round-trip, and the empty-array bug fix is a nice bonus.

---

NikolayS:
## REV Code Review — #55 (round 2)

Re-review after `0325ba8` ("fix: address REV review for PR #55"). Diff since last review: +121 / -29 across 6 files.

---

### Previous findings — status

| Finding | Status |
|---|---|
| MEDIUM: misleading "protobuf/msgpack/Avro" claim — `text` can't hold NUL | **Fixed** in SPECx.md §4.1 (new "NUL bytes" paragraph + encode/decode guidance; code examples switched to XML and base64-encoded protobuf) |
| LOW: no explicit `grant execute` for `send*` / `subscribe` / `unsubscribe` | **Fixed** — 8 grants added in `sql/pgque-api/send.sql`, with a co-location comment in `sql/pgque-additions/roles.sql` explaining the build-order reason. `tests/test_pgque_roles.sql` extended to assert all 8 |
| LOW: no test exercising the *untyped* literal path | **Fixed** — `tests/test_api_receive.sql` step 7 adds `json.nocast` case with `'{"z":9,"m":5,"a":1}'` (no cast) and asserts key order stays unsorted. Clear comment calling out the invariant |

All three addressed cleanly. Nice turnaround.

---

### New findings

**MEDIUM** `blueprints/SPECx.md:646` — the replacement protobuf example is syntactically invalid
> ```sql
> select pgque.send('orders', 'order.proto_b64', encode('\x08\x2a\x10\x63'::bytea, 'base64'));
> ```
> This errors at parse time. Reproduced on PG17:
> ```
> ERROR:  invalid hexadecimal digit: "\"
> LINE 1: select encode('\x08\x2a\x10\x63'::bytea, 'base64');
>                       ^
> ```
> PostgreSQL's bytea hex-input format is a single `\x` prefix followed by hex digits only — embedded `\x` separators are not allowed. Correct form: `'\x082a1063'::bytea` (produces `CCoQYw==`).
> **Fix:** `encode('\x082a1063'::bytea, 'base64')` — or use the escape format `E'\\010\\052\\020\\143'::bytea` if the per-byte visual grouping matters.

**LOW** `README.md:356` — NUL-byte caveat didn't make it to the README
> SPECx.md got the new "NUL bytes" paragraph, but the README function-reference table still says "Untyped literals resolve to the `text` overload (fast path, bytes verbatim)" with no mention of the encoding requirement for binary. The README is the entry point most users hit first.
> **Fix:** either add a one-liner to the table ("text payloads only — NUL bytes rejected; encode binary first") or add a brief note + link under §Quickstart pointing to SPECx.md §4.1.

**INFO** `tests/test_pgque_roles.sql:34-36` — follow-up gap is explicitly tracked (good)
> The new comment "receive/ack/nack (grants still inherited via PUBLIC; tracked separately)" is exactly the right level of disclosure — it's an acknowledged remaining gap rather than a silent oversight. No action required here, but worth filing a separate issue so it doesn't get lost. Same four grants that `send_batch` needed (`receive`, `ack`, `nack`, and the modern alias `receive(text, text, integer)` implicitly) still ride on PUBLIC.

---

### Still-green

- Build-order reasoning for co-locating grants in `pgque-api/send.sql` instead of `pgque-additions/roles.sql` is correct (`build/transform.sh:685` concatenates additions before api). The comment in `roles.sql` explaining this is exactly the context a future reader will want.
- `json.nocast` test is the right minimal assertion: if the jsonb overload ever silently wins, the sorted-keys output fails the equality check with a clear error message.
- Grants ordering in the generated `sql/pgque.sql:4524-4531` is after all `create or replace function` calls — correct.

---

### Summary

|---|---|---|
| Security | 0 | 0 |
| Bugs | 0 | 0 |
| Tests | 0 | 0 |
| Guidelines/Grants | 0 | 0 |
| Docs | 2 (1 MEDIUM, 1 LOW) | 0 |

**Result: APPROVE after the SPECx.md bytea literal is fixed.** The bad example will confuse anyone who copy-pastes it — small one-character fix (`\x082a1063`) but worth catching before merge. README NUL-byte note is nice-to-have.

---

PR #56, opened 2026-04-17
fix: promote DLQ from experimental to default install

NikolayS:
## Problem

`pgque.nack()` in the default install calls `pgque.event_dead()`, which is only defined in `sql/experimental/dlq.sql`. On any install that doesn't explicitly load the experimental DLQ, `nack()` raises `function pgque.event_dead(...) does not exist` the moment a message hits `max_retries`. The `pgque.dead_letter` table is similarly experimental-only.

Found during the v0.1 docs audit: PHASES.md v0.1 list didn't include `nack()`, but the shipped install does — so default-install users get a runtime crash on the retry-exhaustion path.

## Fix

Promote the DLQ pieces into the default install.

### Assembly and ordering
- move `sql/experimental/dlq.sql` → `sql/pgque-additions/dlq.sql`
- add `dlq.sql` to the additions list in `build/transform.sh`, **ordered after `roles.sql`** so the explicit grants below can reference `pgque_reader` / `pgque_writer` / `pgque_admin` (which `roles.sql` creates)
- remove `dlq_inspect` from the "experimental APIs must not leak" blocklist in `build/transform.sh`; add a comment explaining the promotion

### Explicit grants on the new surface
Because `dlq.sql` now runs after `roles.sql`'s blanket `grant select on all tables / execute on all functions` passes, the DLQ table and functions need explicit grants:

| Object | Grant |
|---|---|
| `pgque.dead_letter` (table) | `select` → `pgque_reader`, `all` → `pgque_admin` |
| `pgque.dead_letter_dl_id_seq` | `all` → `pgque_admin` |
| `pgque.dlq_inspect(text, int)` | `pgque_reader` |
| `pgque.dlq_replay(bigint)` | `pgque_writer` |
| `pgque.dlq_replay_all(text)` | `pgque_writer` |
| `pgque.event_dead(...)` | `pgque_admin` |
| `pgque.dlq_purge(text, interval)` | `pgque_admin` |

### Tests and docs
- move `test_api_dlq` from `tests/run_experimental.sql` → `tests/run_all.sql` (now covered by CI against PG 14–18 by default)
- `blueprints/PHASES.md`: DLQ moves out of the experimental section into the v0.1 default install list (adds `nack`, `dead_letter` table, `event_dead`, `dlq_*` helpers)
- `README.md`: drop "dead letter queue tooling" from the experimental bullet list under §Usage examples

## Test plan

- [x] `build/transform.sh` rebuilds `sql/pgque.sql` cleanly (all assembly verification passes)
- [x] DLQ symbols land in `sql/pgque.sql` after `roles.sql` content
- [ ] CI on PG 14, 15, 16, 17 (pending)
- [ ] `tests/test_api_dlq.sql` passes as part of the default `run_all.sql` suite (covered by CI)

### DIFF
--- a/README.md
+++ b/README.md
@@ -290,7 +290,6 @@ The repo also contains additional SQL under `sql/experimental/` for features
 that are not part of the default v0.1 install yet:
 
 - delayed delivery
-- dead letter queue tooling
 - observability helpers
 
 Those pieces are being kept out of the default install until the API surface settles.
--- a/blueprints/PHASES.md
+++ b/blueprints/PHASES.md
@@ -25,6 +25,13 @@ useful, and stable enough to promote into the default install.
 - `pgque.subscribe(queue, consumer)`
 - `pgque.receive(queue, consumer, max_return)`
 - `pgque.ack(batch_id)`
+- `pgque.nack(batch_id, msg, retry_after, reason)`
+
+### Dead letter queue
+- `pgque.dead_letter` table
+- `pgque.event_dead()` — called by `nack()` when `retry_count >= max_retries`
+- `pgque.dlq_inspect()`, `pgque.dlq_replay()`, `pgque.dlq_replay_all()`,
+  `pgque.dlq_purge()`
 
 ## Experimental SQL (`sql/experimental/`)
 
@@ -34,11 +41,6 @@ These files are not installed by default in v0.1.
 - delayed delivery table
 - `pgque.send_at()`
 - delayed-delivery maintenance hook
-
-### `sql/experimental/dlq.sql`
-- dead letter queue tables and helpers
-- replay / inspect / purge workflows
-- retry-to-DLQ API surface
 
 ### `sql/experimental/observability.sql`
 - queue / consumer stats
--- a/build/transform.sh
+++ b/build/transform.sh
@@ -681,7 +681,7 @@ echo "-- Section 6: pgque additions (NEW — not derived from PgQ)" >> "${INSTAL
 echo "-- ======================================================================" >> "${INSTALL_FILE}"
 echo "" >> "${INSTALL_FILE}"
 
-for addition_file in config.sql queue_max_retries.sql lifecycle.sql roles.sql; do
+for addition_file in config.sql queue_max_retries.sql lifecycle.sql roles.sql dlq.sql; do
   echo "-- pgque-additions/${addition_file}" >> "${INSTALL_FILE}"
   cat "${ADDITIONS_DIR}/${addition_file}" >> "${INSTALL_FILE}"
   echo "" >> "${INSTALL_FILE}"
@@ -819,8 +819,10 @@ else
   asm_errors=$((asm_errors + 1))
 fi
 
-# Verify experimental APIs are NOT in the default install script
-if grep -q 'maint_deliver_delayed\|send_at\|delayed_events\|dlq_inspect\|otel_metrics\|queue_stats' "${INSTALL_FILE}"; then
+# Verify experimental APIs are NOT in the default install script.
+# DLQ (dlq_inspect / event_dead / pgque.dead_letter) was promoted to the default
+# install — nack() depends on event_dead, so the DLQ has to ship by default.
+if grep -q 'maint_deliver_delayed\|send_at\|delayed_events\|otel_metrics\|queue_stats' "${INSTALL_FILE}"; then
   echo "FAIL: experimental APIs leaked into default install script"
   asm_errors=$((asm_errors + 1))
 else
--- a/sql/pgque-additions/dlq.sql
+++ b/sql/pgque-additions/dlq.sql
@@ -5,10 +5,23 @@
 -- See SPECx.md section 4.5.
 
 -- pgque.dead_letter table
+--
+-- FK behavior: both dl_queue_id and dl_consumer_id use `on delete cascade`.
+-- Rationale:
+--   - `pgque.drop_queue()` deletes the pgque.queue row unconditionally. DLQ
+--     entries for a dropped queue are meaningless, so cascading them is
+--     correct (users who want to preserve the audit trail should call
+--     `pgque.dlq_purge` or copy the rows out before dropping the queue).
+--   - `pgque.unregister_consumer()` deletes the pgque.consumer row only when
+--     the consumer has no other subscriptions. That is an explicit,
+--     user-initiated action (not routine maintenance), so cascading the
+--     historical DLQ rows tied to that (now-removed) consumer id is the
+--     least-surprising default. Same escape hatch: purge/copy first if the
+--     audit trail matters.
 create table if not exists pgque.dead_letter (
     dl_id           bigserial primary key,
-    dl_queue_id     int4 not null references pgque.queue(queue_id),
-    dl_consumer_id  int4 not null references pgque.consumer(co_id),
+    dl_queue_id     int4 not null references pgque.queue(queue_id)    on delete cascade,
+    dl_consumer_id  int4 not null references pgque.consumer(co_id)    on delete cascade,
     dl_time         timestamptz not null default now(),
     dl_reason       text,
 
@@ -153,15 +166,20 @@ $$ language plpgsql security definer set search_path = pgque, pg_catalog;
 -- ---------------------------------------------------------------------------
 -- Grants
 -- ---------------------------------------------------------------------------
--- dlq.sql runs after roles.sql in transform.sh, so role names already exist
--- and roles.sql's blanket "grant select on all tables / execute on all
--- functions" passes have already run — the explicit grants below cover the
--- DLQ table and functions created here.
+-- dlq.sql runs after roles.sql in transform.sh, so role names already exist.
+-- However, roles.sql's blanket "grant select on all tables / execute on all
+-- functions" passes ran *before* dlq.sql created these objects, so they do
+-- NOT cover the DLQ table and functions. The explicit grants below are
+-- therefore required (not redundant) for every role mentioned, including
+-- pgque_admin.
 --
 -- dlq_inspect is read-only — available to pgque_reader and above.
 -- dlq_replay / dlq_replay_all re-insert events — writer-level.
--- dlq_purge and event_dead stay on PUBLIC default; pgque_admin already has
--- blanket execute via roles.sql.
+-- dlq_purge / event_dead: admin-level operations (purge = data loss,
+-- event_dead = internal DLQ hook called from nack()). Granted to pgque_admin
+-- explicitly for the reason above. Left on PUBLIC default EXECUTE for now;
+-- consider revoke-from-public if the codebase adopts that convention more
+-- broadly.
 
 grant select on pgque.dead_letter                           to pgque_reader;
 grant all    on pgque.dead_letter                           to pgque_admin;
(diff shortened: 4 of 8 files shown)


NikolayS:
## REV Code Review — #56

**Summary:** Promotes the DLQ from `sql/experimental/` to the default install. Fixes a real crash: `pgque.nack()` in the default install calls `pgque.event_dead()`, which only existed in the experimental DLQ file — so any install without explicit experimental load hit `function pgque.event_dead(...) does not exist` on the retry-exhaustion path. Moves the file, wires it into `build/transform.sh` (ordered after `roles.sql`), adds explicit grants for the DLQ table/sequence/functions, updates the experimental-leak guard, moves `test_api_dlq` into `run_all.sql`, and updates PHASES.md + README.md to reflect the new default-install surface. +224/-12.

The core fix is correct and the ordering decision (dlq.sql after roles.sql) is the right call because the grants need to reference the role names. One real interaction bug surfaces from the promotion; see below.

---

### BLOCKING — 0

### NON-BLOCKING

**MEDIUM** `sql/pgque-additions/dlq.sql:11-13` — FK constraint breaks `drop_queue` on any queue that has DLQ entries
> The DLQ table declares:
> ```sql
> dl_queue_id int4 not null references pgque.queue(queue_id)
> ```
> with no `on delete` clause, so the default `no action`/`restrict` applies. `pgque.drop_queue()` (`sql/pgque.sql:1561`) does a plain `delete from pgque.queue where queue_name = x_queue_name` and does **not** clean up DLQ rows first. Result: any queue that has ever hit max-retries cannot be dropped until the caller manually purges its DLQ entries.
>
> This was tolerable while DLQ was opt-in experimental. Now that every default-install user gets the DLQ, this is a foot-gun for every user.
>
> **Evidence that the author is already aware of the workaround:** `tests/test_api_dlq.sql:105-107` explicitly hacks around it in test-2 cleanup:
> ```sql
> -- Delete DLQ entries before unregistering consumer (FK constraint)
> delete from pgque.dead_letter
> where dl_queue_id = (select queue_id from pgque.queue where queue_name = 'test_dlq2');
> ```
> If the test has to do this, real users will too — they just won't know why `drop_queue` is failing.
>
> **Fix options (pick one):**
> 1. `references pgque.queue(queue_id) on delete cascade` — simplest; drop_queue nukes DLQ history with the queue. Defensible because a dropped queue's DLQ is meaningless anyway.
> 2. In `drop_queue()`, add `delete from pgque.dead_letter where dl_queue_id = q.queue_id;` before the `delete from pgque.queue` (both in the force and non-force paths).
> 3. At minimum: document the requirement in README and have `drop_queue` raise a clearer error message that points to `dlq_purge`.
>
> Same concern applies to `dl_consumer_id int4 not null references pgque.consumer(co_id)` — `unregister_consumer` may or may not remove the consumer row; worth double-checking that codepath while you're at it.

**LOW** `sql/pgque-additions/dlq.sql:163-167` / `sql/pgque.sql:4406-4410` — grant-block comment is self-contradictory
> ```sql
> -- dlq_purge and event_dead stay on PUBLIC default; pgque_admin already has
> -- blanket execute via roles.sql.
> ```
> This reads as "no explicit grant needed," but the very next lines do grant explicitly:
> ```sql
> grant execute on function pgque.event_dead(...) to pgque_admin;
> grant execute on function pgque.dlq_purge(text, interval) to pgque_admin;
> ```
> The explicit grant is actually **required**, because `roles.sql` runs *before* `dlq.sql` in the assembled install, so its `grant execute on all functions in schema pgque to pgque_admin` pass (`sql/pgque-additions/roles.sql:78`) runs before `event_dead` / `dlq_purge` exist and therefore cannot cover them. The top comment on the block already says this correctly; the second half contradicts it.
> **Fix:** rewrite as:
> ```sql
> -- dlq_purge / event_dead: admin-level operations (purge data loss,
> -- internal DLQ hook). Granted to pgque_admin explicitly because
> -- roles.sql's blanket pass ran before these functions existed. Left on
> -- PUBLIC default EXECUTE for now; consider revoke-from-public if the
> -- codebase adopts that convention more broadly.
> ```

### POTENTIAL ISSUES

**LOW** `sql/pgque-additions/dlq.sql` — `event_dead` and `dlq_purge` remain callable by `PUBLIC`
> Default PG behavior grants EXECUTE on functions to `PUBLIC`. These two aren't revoked, so any authenticated role can call them directly. `event_dead`'s `SECURITY DEFINER` body writes to `pgque.dead_letter` and requires a valid `sub_batch` lookup (raises `batch not found` otherwise) so it's constrained, but `dlq_purge(queue_name, '0 seconds')` from `PUBLIC` is a data-loss vector — any role that can open a connection can purge any queue's DLQ.
> Consistent with the rest of the codebase (nothing else revokes from PUBLIC), so probably fine for this PR. Worth a separate cleanup pass once the revoke-from-public convention is adopted anywhere.

---

### Verified (spot-checked)

- `pgque.nack()` *does* call `pgque.event_dead()` on this branch (`sql/pgque-api/receive.sql:89`), and `event_dead` lives in `dlq.sql` — the stated crash is real and the fix is necessary for default-install users.
- `event_dead` grant signature matches its 12-arg definition (4 args have defaults; PG resolves by base types, so listing all 12 is correct).
- `bigserial` convention confirms the sequence name: `pgque.dead_letter_dl_id_seq`. ✓
- `transform.sh:684` additions list order is `config → queue_max_retries → lifecycle → roles → dlq` — roles before dlq is exactly what the grant-reference comment needs.
- `transform.sh:825` experimental-leak guard: `dlq_inspect` correctly removed, other experimental markers (`maint_deliver_delayed`, `send_at`, `delayed_events`, `otel_metrics`, `queue_stats`) still guarded.
- `tests/run_all.sql:61` correctly adds `test_api_dlq`; `tests/run_experimental.sql:12` correctly drops it.
- `PHASES.md` and `README.md` updates are accurate.
- `dlq_replay` uses the 7-arg `insert_event` (confirmed exists at `sql/pgque.sql:1665`). ✓

### Summary

|---|---|---|
| Security | 0 | 1 LOW |
| Bugs | 1 MEDIUM | 0 |
| Tests | 0 | 0 |
| Guidelines/Grants | 1 LOW | 0 |
| Docs | 0 | 0 |

**Result: APPROVE after addressing the `drop_queue` / DLQ FK interaction (MEDIUM).** The core promotion is clean and necessary — `nack()` is broken without it. The comment cleanup is trivial. The FK-constraint interaction is the one real design question and the test's explicit workaround is the smoking gun: please pick option 1 (`on delete cascade`) or option 2 (explicit cleanup in `drop_queue`) before merge, otherwise you're shipping the same foot-gun to every default-install user.

---

NikolayS:
Thanks for the review — all three points addressed in NikolayS/pgque@9b8af9a.

**MEDIUM — FK cascade on `dead_letter`.** Went with `on delete cascade` on *both* `dl_queue_id` and `dl_consumer_id` after checking `unregister_consumer` (sql/pgque.sql L1869): it deletes the `pgque.consumer` row only when no other subscriptions remain for that consumer, which makes it an explicit user-initiated action rather than routine maintenance. Cascading historical DLQ rows tied to that (now-removed) id is the least-surprising default; users who want the audit trail should call `dlq_purge` (or copy rows out) before dropping the queue / unregistering the consumer. I didn't need `set null` on `dl_consumer_id` — the column stays `not null`. Rationale is in the commit message and in a new comment block above the `create table`.

**LOW — grant-block comment.** Rewritten to correctly state that `roles.sql`'s blanket pass ran *before* `dlq.sql` created these objects, so the explicit grants are required (not redundant). Kept the existing convention on PUBLIC default execute.

**LOW / POTENTIAL — PUBLIC execute on `event_dead` / `dlq_purge`.** Deferred as agreed — belongs in a separate convention-change PR.

FK-constraint workaround in `tests/test_api_dlq.sql` cleanup is gone; full `run_all.sql` suite passes locally on PG 18.3.

NikolayS:
## REV Code Review — #56 (round 2)

Re-review after `9b8af9a` ("fix: address PR #56 review feedback"). Diff since last review: +54 / -19 across 3 files — all in direct response to the previous round.

---

### Previous findings — status

| Finding | Status |
|---|---|
| MEDIUM: `drop_queue` + DLQ FK foot-gun (no cascade, test had to manually purge) | **Fixed** — both `dl_queue_id` and `dl_consumer_id` now have `on delete cascade`. Rationale captured in a 10-line comment on the table DDL. `tests/test_api_dlq.sql:105-107` workaround removed (replaced with a one-line comment noting the cascade makes it automatic) — the test now exercises the same cleanup path real users will hit |
| LOW: grant-block comment self-contradiction | **Fixed** — rewritten to correctly say the blanket grants from `roles.sql` ran *before* `dlq.sql` created these objects and therefore the explicit grants are **required**, not redundant. The `dlq_purge` / `event_dead` admin-level section was also rewritten along the exact lines I'd sketched |
| LOW (potential): `dlq_purge` / `event_dead` still on PUBLIC default EXECUTE | **Deferred with explicit note** — the comment now says "consider revoke-from-public if the codebase adopts that convention more broadly," which is the right stance for this patch |

All three addressed. No new findings introduced.

---

### Verification of the cascade fix

Spot-checked the cleanup path in `tests/test_api_dlq.sql:100-108`:

1. Test 2 inserts an event, forces `retry_count >= max_retries`, nacks it → row lands in `pgque.dead_letter` (asserted at line 98).
2. Cleanup now calls `unregister_consumer` then `drop_queue` with **no manual DLQ delete**.
3. First cascade trigger: if `unregister_consumer` removes the `pgque.consumer` row (which it does for single-subscription consumers), the DLQ row cascades via `dl_consumer_id`.
4. Second cascade trigger: `drop_queue` deletes `pgque.queue` row; any remaining DLQ rows cascade via `dl_queue_id`.
5. Test passes ⇒ at least one of the two cascades fires, and neither raises FK errors.

The cascade decision itself is the right call:
- Alternative `on delete restrict` would keep `drop_queue` foot-gun forever.
- Alternative `on delete set null` isn't available (columns are `not null`, and changing them would lose consumer attribution on surviving DLQ rows).
- Cascade + explicit "copy rows out first if you want the audit trail" escape hatch in the comment is the most defensible default for an embedded library.

One very minor consideration (not blocking, not even worth filing): the test doesn't explicitly assert `count(dead_letter where dl_queue_id = …) = 0` post-drop. It relies on "drop_queue didn't error" as proof. Since the cascade is a PG-enforced behavior (not application logic), testing it more deeply is low-value — but a single `assert` line would turn an implicit contract into an explicit one. Entirely optional.

### Regression spot-checks

- `sql/pgque.sql` lines match the committed `dlq.sql` (transform.sh rebuild passed per the PR description). ✓
- No other call sites of the DLQ API were touched — the fix is surgical. ✓
- `tests/run_all.sql` still includes `test_api_dlq`; `tests/run_experimental.sql` no longer does. ✓
- Nothing else in the tree references the deleted `sql/experimental/dlq.sql` path. ✓

### Summary

|---|---|---|
| Security | 0 | 1 LOW (carried, deferred by design) |
| Bugs | 0 | 0 |
| Tests | 0 | 0 |
| Guidelines/Grants | 0 | 0 |
| Docs | 0 | 0 |

**Result: APPROVE.** Clean resolution of both findings. The cascade fix is the right shape (simplest semantics, foot-gun removed, audit-trail escape hatch documented), and the test now mirrors the cleanup path real users will take rather than papering over it.

---

PR #57, opened 2026-04-17
fix: grant execute on pgque.receive/ack/nack to pgque_writer

NikolayS:
## Problem

`README.md` §"Roles and grants" claims that `pgque_writer` has execute privileges on `receive`, `ack`, and `nack`. `tests/test_pgque_roles.sql` also asserts it. But the grants were never explicit — those functions rely on PostgreSQL's default `PUBLIC execute`.

The role test passes today only because `pgque_writer` inherits `PUBLIC` privileges, not because of an explicit grant. Found during the v0.1 docs audit as a secondary issue behind the `nack()` / DLQ bug.

## Fix

Add three `grant execute` statements at the bottom of `sql/pgque-api/receive.sql`, matching the convention already used by `sql/pgque-api/send.sql` (grants colocated with function definitions, since the `pgque-api/` section is assembled after `roles.sql`):

\`\`\`sql
grant execute on function pgque.receive(text, text, int)                      to pgque_writer;
grant execute on function pgque.ack(bigint)                                   to pgque_writer;
grant execute on function pgque.nack(bigint, pgque.message, interval, text)   to pgque_writer;
\`\`\`

Also update the stale comment in \`tests/test_pgque_roles.sql\` — it previously said the grants were "inherited via PUBLIC; tracked separately", which is no longer true.

## Scope

- Does not revoke \`PUBLIC\` execute (consistent with the rest of the codebase — no function in \`pgque-additions/\` or \`pgque-api/\` revokes from PUBLIC explicitly).
- Does not change the role test's assertions; they already cover the intended surface.

## Test plan

- [x] \`build/transform.sh\` rebuilds \`sql/pgque.sql\` cleanly
- [x] Explicit grants appear at the bottom of \`sql/pgque.sql\`
- [ ] \`tests/test_pgque_roles.sql\` passes on PG 14–18 via CI

### DIFF
--- a/sql/pgque-api/receive.sql
+++ b/sql/pgque-api/receive.sql
@@ -99,3 +99,14 @@ begin
     return 1;
 end;
 $$ language plpgsql security definer set search_path = pgque, pg_catalog;
+
+-- ---------------------------------------------------------------------------
+-- Grants
+-- ---------------------------------------------------------------------------
+-- Colocated here (not in pgque-additions/roles.sql) because roles.sql is
+-- assembled before pgque-api/, so these functions do not yet exist when
+-- roles.sql runs. Same convention as sql/pgque-api/send.sql.
+
+grant execute on function pgque.receive(text, text, int)                      to pgque_writer;
+grant execute on function pgque.ack(bigint)                                   to pgque_writer;
+grant execute on function pgque.nack(bigint, pgque.message, interval, text)   to pgque_writer;
--- a/sql/pgque.sql
+++ b/sql/pgque.sql
@@ -4225,6 +4225,12 @@ grant execute on function pgque.finish_batch(bigint) to pgque_writer;
 grant execute on function pgque.event_retry(bigint, bigint, timestamptz) to pgque_writer;
 grant execute on function pgque.event_retry(bigint, bigint, integer) to pgque_writer;
 
+-- Note: grants for the modern API wrappers (send*, subscribe, unsubscribe,
+-- receive, ack, nack) live colocated with their definitions in
+-- sql/pgque-api/*.sql. transform.sh appends pgque-additions/ before
+-- pgque-api/, so API-layer grants cannot reference their functions from
+-- this file.
+
 -- ---------------------------------------------------------------------------
 -- Admin: full access to everything in the pgque schema
 -- ---------------------------------------------------------------------------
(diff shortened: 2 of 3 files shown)


NikolayS:
## REV Code Review — #57

**Summary:** Adds three explicit `grant execute … to pgque_writer` statements for `pgque.receive(text, text, int)`, `pgque.ack(bigint)`, and `pgque.nack(bigint, pgque.message, interval, text)` at the bottom of `sql/pgque-api/receive.sql`, matching the convention `sql/pgque-api/send.sql` established in #55. Also updates the stale "tracked separately" comment in `tests/test_pgque_roles.sql`. +30/-1, three files.

---

### BLOCKING — 0

### NON-BLOCKING — 0

### POTENTIAL ISSUES — 0

---

### Verification

Checked each grant's signature against the actual function definitions on this branch:

| Grant | Actual definition | Match |
|---|---|---|
| `pgque.receive(text, text, int)` | `sql/pgque-api/receive.sql:27-28` — `i_queue text, i_consumer text, i_max_return int default 100` | ✓ |
| `pgque.ack(bigint)` | `sql/pgque-api/receive.sql:60` — `i_batch_id bigint` | ✓ |
| `pgque.nack(bigint, pgque.message, interval, text)` | `sql/pgque-api/receive.sql:68-72` — `i_batch_id bigint, i_msg pgque.message, i_retry_after interval default '60 seconds', i_reason text default null` | ✓ |

Signatures use the base types (`int` is an alias for `integer`, both resolve to the same function) — `has_function_privilege('pgque_writer', 'pgque.receive(text, text, integer)', …)` in the test resolves to the same function object as the `int` grant, so the existing test assertions cover it.

Also verified:
- Generated `sql/pgque.sql` places grants at lines 4390-4402, after the function defs at 4317/4350/4358. Correct order.
- `pgque-additions/roles.sql` got the cross-reference comment (lines 4228-4232 in the generated file) explaining the split.
- `tests/test_pgque_roles.sql` comment change accurately describes the new state.
- No revoke-from-PUBLIC, which matches the stated scope and the rest of the codebase.

### Summary

|---|---|---|
| Security | 0 | 0 |
| Bugs | 0 | 0 |
| Tests | 0 | 0 |
| Guidelines/Grants | 0 | 0 |
| Docs | 0 | 0 |

**Result: APPROVE.** Clean follow-up to #55. Exactly the right scope — three grants, one comment, no drive-by changes. The pre-existing `PUBLIC` execute hold-over is called out in the PR description as intentional and consistent with the rest of the codebase, which is the correct stance for this patch.

---

PR #62, opened 2026-04-18
Rotate subscription and tick tables to avoid held-xmin bloat (#61)

### LINKED ISSUE #61: Metadata-table bloat under held xmin — subscription/tick need rotation
## Problem

PgQue bypasses MVCC pathology on **event tables** via rotation (TRUNCATE). But the **small metadata tables** — specifically `pgque.subscription` and `pgque.tick` — still depend on UPDATE/INSERT and plain autovacuum. When the xmin horizon is held (idle-in-transaction, long-running writer, stale logical replication slot, physical standby with `hot_standby_feedback=on`), those tiny tables accumulate unvacuumable dead tuples and their indexes bloat. Consumer throughput degrades ~5× even though the event tables themselves stay clean.

**R4 benchmark evidence** (PG18, 1000 ev/s producer, 1h held xmin):

| Table | Peak dead tuples | Size semantic |
|---|---:|---|
| `pgque.subscription` | 14 312 | ~5 rows (one per consumer) |
| `pgque.tick` | 7 154 | tick heartbeats |
| `pgque.queue` | 238 | per-queue counters |
| `pgque.event_*` | **0** | rotation via TRUNCATE ✓ |

`pgque.subscription` at ~14k dead rows on a 5-row logical table = index bloat that makes `next_batch` / `finish_batch` index scans 5× slower. Consumer throughput drops from ~4.5k iter/s steady to ~900 iter/s during the held-xmin window. Not catastrophic (compared to DELETE-based queues that collapse to 5-25 TPS), but it's residual MVCC exposure in an extension whose whole selling point is MVCC immunity.

## PoC — 4 mitigation candidates tested, 90s workload at 5000 UPDATE/s under held xmin

([Full PoC thread + chart](https://gitlab.com/postgres-ai/postgresql-consulting/tests-and-benchmarks/-/issues/77#note_3263417609))

| Scenario | Approach | Achieved TPS | Avg latency | Heap after 90s | Verdict |
|---|---|---:|---:|---:|---|
| A | UPDATE baseline | 2878 | 9001 ms | 9.0 MB | ❌ collapsed |
| B | UPDATE + `fillfactor=50` (HOT-friendly) | 2867 | 9054 ms | 9.0 MB | ❌ HOT doesn't help |
| C | **two-table rotation + rename-swap + TRUNCATE** | **5020** | **1.0 ms** | 1.8 MB (saw-tooth) | ✅ works |
| D | append-only INSERT + TRUNCATE | 5001 | 0.38 ms | 2.1 MB | ✅ works, but state-loss risk |
| E | VACUUM FULL every N seconds | unchanged | N/A | **unchanged** | ❌ useless under held xmin |

Key findings:

### HOT updates do not help
99.6% of the UPDATEs in scenario B took the HOT path (`n_tup_hot_upd / n_tup_upd = 0.996`). The heap still grew to 9 MB (1100× the clean single-page size of 8 KB). Reason: `heap_page_prune` requires `xmax < OldestXmin`; when xmin is held, recent dead HOT tuples can't be pruned either.

### VACUUM FULL does not help (and bloats pg_class)
Testing VACUUM FULL on the 1-row table under held xmin:
- Heap size before: 3472 kB. After VACUUM FULL: **3472 kB** (unchanged).
- VACUUM FULL uses the same MVCC visibility rules — dead-but-potentially-visible tuples get copied to the new relfilenode rather than dropped.
- Each VACUUM FULL adds ~4 dead tuples to `pg_class` (relfilenode UPDATE). 5 cycles = +20 dead pg_class tuples. Frequent use would degrade ALL query planning.
- Guarding VACUUM FULL on "high dead tup count" doesn't rescue this — the trigger condition is exactly when VACUUM FULL can't work.

### Rotation works — TRUNCATE bypasses MVCC
TRUNCATE doesn't check visibility rules; it drops the old file regardless of who holds xmin. Scenario C ran at target rate, kept latency at 1 ms, and heap stayed bounded at ~1.8 MB (saw-toothing around rotation points every 10 s).

### `pgque.subscription`
Implement two-table rotation (scenario C shape):
- `pgque.subscription_a` and `pgque.subscription_b` with identical schema
- `pgque.subscription_ctrl` pointer table (1 row, which side is active)
- `pgque.finish_batch(b)` routes UPDATE through a plpgsql wrapper that resolves the active side
- Rotation task (probably piggybacked on the existing ticker): `COPY latest state → inactive`, swap pointer, `TRUNCATE old`
- Rotation cadence: once per `queue_rotation_period` (align with event-table rotation) — already configurable per queue

### `pgque.tick`
Already insert-only. Even simpler: just TRUNCATE on rotation cadence. No two-table dance needed; consumers that depend on tick history can keep a longer-lived `tick_archive` if required.

### `pgque.queue`
Very low UPDATE rate — may not need intervention. Revisit after subscription/tick are fixed and re-bench.

## Acceptance criteria

After this change, under the R4/R5 held-xmin benchmark:
- `pgque.subscription*` peak dead tuples ≤ 500 (vs current 14 312)
- `pgque.tick*` peak dead tuples ≤ 200 (vs current 7 154)
- Consumer iteration rate within 20% of clean-baseline during held-xmin window (vs current 5× drop)
- Event-table rotation behavior unchanged

## Scope note

This is a pgque change, not a pgq upstream patch. The pgq C-free v3.5.1 base has the same pattern and would benefit, but that repo is upstream-maintained and out of scope here.

## References

- R4 full bench: https://gitlab.com/postgres-ai/postgresql-consulting/tests-and-benchmarks/-/issues/77
- PoC details: https://gitlab.com/postgres-ai/postgresql-consulting/tests-and-benchmarks/-/issues/77#note_3263417609
- PoC script: `/tmp/pgque_metadata_poc.sh` locally (will move to a subdir of this repo when the fix PR lands)

NikolayS:
## Summary

Fixes #61.

Applies PgQ's own 3-table rotation pattern to the two PgQ metadata tables that are NOT already rotated — `pgque.subscription` and `pgque.tick` — so that held-xmin bloat on those tables is bounded instead of unbounded.

The event tables (`pgque.event_<queue>_0/_1/_2`) already rotate, and R4/R5 showed their dead-tuple count stays at 0 even under a 10-minute idle-in-transaction. But the metadata tables don't rotate; under the same R5 run they peaked at ~14 312 dead tuples on `pgque.subscription` and ~7 154 on `pgque.tick`, which dragged pgque's consumer iter-TPS from ~3 500 down to ~1 540 during the held-xmin window.

This PR extends the rotation pattern to those two tables.

## Design

- `pgque.subscription` becomes a view (`UNION ALL` over three physical children `subscription_0 / _1 / _2`), filtered to the currently-active child via `pgque.meta_rotation.cur_subscription_table`. `INSTEAD OF` triggers route every `INSERT` / `UPDATE` / `DELETE` to the active child.
- `pgque.tick` becomes a view (`UNION ALL` over `tick_0 / _1 / _2`) with no active-child filter — a consumer's `sub_last_tick` can legitimately reference a tick that was inserted before the most recent rotation. `INSTEAD OF` triggers route `INSERT`s to the active child. `DELETE`s (as issued by the existing `maint_rotate_tables_step1`) fan out to all three children.
- `pgque.maint_rotate_metadata()` performs the rotation step1: `TRUNCATE` the `(cur + 1) % 3` slot, `INSERT … SELECT` live subscription rows from `cur` → the new slot, flip `cur`. For tick, truncation is gated on "no live `sub_last_tick` references rows on the target slot."
- `pgque.maint_rotate_metadata_step2()` is the step2 counterpart (same pattern PgQ uses for event rotation), scheduled by `pgque.start()` alongside the existing `pgque_rotate_step2` cron job.
- Default rotation period is 30 s, tunable via GUC `pgque.meta_rotation_period`.
- `maint_operations()` emits both rotation calls; `maint()` skips the step2 call (it needs its own transaction, same as the event `maint_rotate_tables_step2` case).

No external dependencies added. Install path stays `psql -f sql/pgque.sql`. Uninstall picks up the new children transitively via `DROP SCHEMA … CASCADE`. No C accelerator.

## Correctness trade-offs

The subscription view filters to `cur`, which means that any transaction whose MVCC snapshot was taken before a rotation committed will (via `pgque.meta_rotation`) see the *old* `cur`'s child. That matches how PgQ's event tables already work: long-running read snapshots that predate a rotation see the pre-rotation layout. pgque itself has no code path that reads `pgque.subscription` under a long-lived snapshot; `next_batch_custom` / `finish_batch` are short transactions.

Tick rotation is more conservative than subscription rotation: we only truncate the target tick slot if no live `sub_last_tick` currently resolves to a row in it. In the (rare) case where a consumer has lagged for longer than two rotation periods, the tick flip is deferred until the consumer catches up. Subscription rotation still happens in that case — which is the dominant bloat source.

## Acceptance criteria

- pgque consumer iter-TPS during held-xmin window within 20% of the clean baseline.
- `pgque.subscription_*` peak dead tuples ≤ 500 (was 14 312 on R5 without this patch).
- `pgque.tick_*` peak dead tuples ≤ 200 (was 7 154 on R5 without this patch).
- All existing `tests/*.sql` continue to pass.

## Test plan

- [x] Fresh install on a scratch database (`psql -f sql/pgque.sql`) produces no errors or warnings.
- [x] `pgque.create_queue`, `pgque.register_consumer`, send → ticker → next_batch → finish_batch happy path works across multiple forced rotations.
- [x] After 3 forced rotations (full cycle through `cur = 0 → 1 → 2 → 0`), `pgque.subscription` view returns exactly one row per `(sub_queue, sub_consumer)` and `sub_last_tick` resolves via the view.
- [x] Dead tuples on `subscription_<cur>` are cleared to 0 after rotation (heap-only — tuples on the previous slot stick around until that slot is truncated two rotations later).
- [ ] R6 smoke test (pgbench, 5 min clean + 10 min idle-in-tx + 5 min recovery) passes the acceptance criteria above.

R6 smoke results will be posted as a follow-up comment on this PR.

### DIFF
--- a/build/transform.sh
+++ b/build/transform.sh
@@ -789,7 +789,7 @@ echo "-- Section 6: pgque additions (NEW — not derived from PgQ)" >> "${INSTAL
 echo "-- ======================================================================" >> "${INSTALL_FILE}"
 echo "" >> "${INSTALL_FILE}"
 
-for addition_file in config.sql queue_max_retries.sql lifecycle.sql roles.sql dlq.sql; do
+for addition_file in config.sql queue_max_retries.sql lifecycle.sql roles.sql dlq.sql metadata_rotation.sql; do
   echo "-- pgque-additions/${addition_file}" >> "${INSTALL_FILE}"
   cat "${ADDITIONS_DIR}/${addition_file}" >> "${INSTALL_FILE}"
   echo "" >> "${INSTALL_FILE}"
--- a/sql/pgque-additions/lifecycle.sql
+++ b/sql/pgque-additions/lifecycle.sql
@@ -56,7 +56,8 @@ begin
     select cron.schedule_in_database(
         'pgque_rotate_step2',
         '10 seconds',
-        $sql$SELECT pgque.maint_rotate_tables_step2()$sql$,
+        $sql$SELECT pgque.maint_rotate_tables_step2();
+             SELECT pgque.maint_rotate_metadata_step2();$sql$,
         v_dbname
     ) into v_step2_id;
 
--- a/sql/pgque-additions/metadata_rotation.sql
+++ b/sql/pgque-additions/metadata_rotation.sql
@@ -1,0 +1,832 @@
+-- metadata_rotation.sql -- 3-table rotation for pgque.subscription and pgque.tick
+-- Copyright 2026 Nikolay Samokhvalov. Apache-2.0 license.
+--
+-- Converts the upstream PgQ single-table pgque.subscription and pgque.tick into
+-- 3-child UNION ALL views to cap held-xmin bloat, mirroring the event-table
+-- rotation pattern already used for pgque.event_<queue>_0/1/2.
+--
+-- Load order: AFTER config.sql (needs pgque.config table), AFTER all PgQ-derived
+-- function files (overrides maint_tables_to_vacuum, maint_operations,
+-- unregister_consumer).
+
+-- ======================================================================
+-- Step 1: Rotation-pointer singleton
+-- ======================================================================
+--
+-- Declared before the views so the subscription view's WHERE clause can
+-- reference it. Keep this table tiny and hot: it is updated on every
+-- rotation cycle and kept separate from pgque.config (which is updated
+-- only at start()/stop() time).
+
+create table if not exists pgque.meta_rotation (
+    singleton                bool      primary key default true check (singleton),
+    cur_subscription_table   smallint  not null default 0,
+    cur_tick_table           smallint  not null default 0,
+    last_rotation_time       timestamptz not null default now(),
+    last_rotation_step1_txid bigint    not null default pg_current_xact_id()::text::bigint,
+    last_rotation_step2_txid bigint             default pg_current_xact_id()::text::bigint
+);
+
+insert into pgque.meta_rotation (singleton) values (true)
+on conflict (singleton) do nothing;
+
+-- ======================================================================
+-- Step 2: Replace pgque.tick with 3-child rotation
+-- ======================================================================
+--
+-- The upstream PgQ tables.sql defines pgque.tick as a plain table.
+-- We need to replace it with a view over three child tables. On a fresh
+-- install the table exists but is empty, so we can drop it safely.
+-- On reinstall (idempotent) the table may already be a view; drop is a
+-- no-op in that case because we use CREATE OR REPLACE VIEW below.
+--
+-- Strategy: rename the base table to tick_tmpl (template/schema carrier),
+-- then build the three child tables from it, then create the view.
+-- We use DO blocks to make each step idempotent.
+
+do $$
+begin
+    -- Convert pgque.tick (base table) to pgque.tick_tmpl if not already done.
+    -- After first install: tick is a view, tick_tmpl already exists — skip.
+    if exists (
+        select 1 from pg_class c
+        join pg_namespace n on n.oid = c.relnamespace
+        where n.nspname = 'pgque' and c.relname = 'tick' and c.relkind = 'r'
+    ) then
+        -- Drop constraints that reference child tables won't need on template
+        alter table pgque.tick drop constraint if exists tick_pkey;
+        alter table pgque.tick drop constraint if exists tick_queue_fkey;
+        alter table pgque.tick rename to tick_tmpl;
+    end if;
+end;
+$$;
+
+-- Template table: holds the column layout for LIKE inheritance.
+-- Never holds rows; only the children do.
+create table if not exists pgque.tick_tmpl (
+    tick_queue      int4        not null,
+    tick_id         bigint      not null,
+    tick_time       timestamptz not null default now(),
+    tick_snapshot   pg_snapshot not null default pg_current_snapshot(),
+    tick_event_seq  bigint      not null,
+    constraint tick_tmpl_queue_fkey foreign key (tick_queue)
+                               references pgque.queue (queue_id)
+);
+
+-- Three physical children, one per rotation slot.
+create table if not exists pgque.tick_0 (
+    like pgque.tick_tmpl including defaults,
+    constraint tick_0_pkey primary key (tick_queue, tick_id),
+    constraint tick_0_queue_fkey foreign key (tick_queue)
+                               references pgque.queue (queue_id)
+);
+create table if not exists pgque.tick_1 (
+    like pgque.tick_tmpl including defaults,
+    constraint tick_1_pkey primary key (tick_queue, tick_id),
+    constraint tick_1_queue_fkey foreign key (tick_queue)
+                               references pgque.queue (queue_id)
+);
+create table if not exists pgque.tick_2 (
+    like pgque.tick_tmpl including defaults,
+    constraint tick_2_pkey primary key (tick_queue, tick_id),
+    constraint tick_2_queue_fkey foreign key (tick_queue)
+                               references pgque.queue (queue_id)
+);
+
+-- Migrate any rows from the template (on fresh install this is empty).
+do $$
+begin
+    if exists (
+        select 1 from pg_class c
+        join pg_namespace n on n.oid = c.relnamespace
+        where n.nspname = 'pgque' and c.relname = 'tick_tmpl' and c.relkind = 'r'
+    ) then
+        insert into pgque.tick_0
+            (tick_queue, tick_id, tick_time, tick_snapshot, tick_event_seq)
+        select tick_queue, tick_id, tick_time, tick_snapshot, tick_event_seq
+          from pgque.tick_tmpl
+        on conflict do nothing;
+        truncate pgque.tick_tmpl;
+    end if;
+end;
+$$;
+
+-- Drop the base table view if it already exists from a previous install,
+-- then create the UNION ALL view.
+-- (CREATE OR REPLACE VIEW handles the view case idempotently.)
+create or replace view pgque.tick as
+      select 0::smallint as tick_child_table, * from pgque.tick_0
+union all
+      select 1::smallint as tick_child_table, * from pgque.tick_1
+union all
+      select 2::smallint as tick_child_table, * from pgque.tick_2;
+
+-- ======================================================================
+-- Step 3: Replace pgque.subscription with 3-child rotation
+-- ======================================================================
+
+do $$
+begin
+    -- Convert pgque.subscription (base table) to pgque.subscription_tmpl.
+    -- After first install: subscription is a view — skip.
+    if exists (
+        select 1 from pg_class c
+        join pg_namespace n on n.oid = c.relnamespace
+        where n.nspname = 'pgque' and c.relname = 'subscription' and c.relkind = 'r'
+    ) then
+        alter table pgque.subscription drop constraint if exists subscription_pkey;
+        alter table pgque.subscription drop constraint if exists subscription_batch_idx;
+        alter table pgque.subscription drop constraint if exists sub_queue_fkey;
+        alter table pgque.subscription drop constraint if exists sub_consumer_fkey;
+        alter table pgque.subscription rename to subscription_tmpl;
+    end if;
+end;
+$$;
+
+-- Shared sequence for sub_id: survives rotation so sub_id values remain
+-- stable across rotations.
+create sequence if not exists pgque.subscription_sub_id_seq;
+
+-- Sync the sequence floor with any pre-existing max sub_id (idempotent).
+do $$
+declare
+    v_max bigint;
+begin
+    -- If the template table still holds rows (just-renamed), find the max.
+    if exists (
+        select 1 from pg_class c
+        join pg_namespace n on n.oid = c.relnamespace
+        where n.nspname = 'pgque' and c.relname = 'subscription_tmpl' and c.relkind = 'r'
+    ) then
+        select max(sub_id) into v_max from pgque.subscription_tmpl;
+        if v_max is not null then
+            perform setval('pgque.subscription_sub_id_seq', v_max, true);
+        end if;
+    end if;
+end;
+$$;
+
+-- Template table: column layout only, no PK/FK (those live on children).
+create table if not exists pgque.subscription_tmpl (
+    sub_id          int4        not null default nextval('pgque.subscription_sub_id_seq'),
+    sub_queue       int4        not null,
+    sub_consumer    int4        not null,
+    sub_last_tick   bigint,
+    sub_active      timestamptz not null default now(),
+    sub_batch       bigint,
+    sub_next_tick   bigint
+);
+
+create table if not exists pgque.subscription_0 (
+    like pgque.subscription_tmpl including defaults,
+    constraint subscription_0_pkey primary key (sub_queue, sub_consumer),
+    constraint subscription_0_batch_uq unique (sub_batch),
+    constraint sub_0_queue_fkey foreign key (sub_queue)
+                               references pgque.queue (queue_id),
+    constraint sub_0_consumer_fkey foreign key (sub_consumer)
+                               references pgque.consumer (co_id)
+);
+create table if not exists pgque.subscription_1 (
+    like pgque.subscription_tmpl including defaults,
+    constraint subscription_1_pkey primary key (sub_queue, sub_consumer),
+    constraint subscription_1_batch_uq unique (sub_batch),
+    constraint sub_1_queue_fkey foreign key (sub_queue)
+                               references pgque.queue (queue_id),
+    constraint sub_1_consumer_fkey foreign key (sub_consumer)
+                               references pgque.consumer (co_id)
+);
+create table if not exists pgque.subscription_2 (
+    like pgque.subscription_tmpl including defaults,
+    constraint subscription_2_pkey primary key (sub_queue, sub_consumer),
+    constraint subscription_2_batch_uq unique (sub_batch),
+    constraint sub_2_queue_fkey foreign key (sub_queue)
+                               references pgque.queue (queue_id),
+    constraint sub_2_consumer_fkey foreign key (sub_consumer)
+                               references pgque.consumer (co_id)
+);
+
+-- Migrate any pre-existing rows from the template into child_0.
+do $$
+begin
+    if exists (
+        select 1 from pg_class c
+        join pg_namespace n on n.oid = c.relnamespace
+        where n.nspname = 'pgque' and c.relname = 'subscription_tmpl' and c.relkind = 'r'
+    ) then
+        insert into pgque.subscription_0
+            (sub_id, sub_queue, sub_consumer, sub_last_tick,
+             sub_active, sub_batch, sub_next_tick)
+        select sub_id, sub_queue, sub_consumer, sub_last_tick,
+               sub_active, sub_batch, sub_next_tick
+          from pgque.subscription_tmpl
+        on conflict do nothing;
+        truncate pgque.subscription_tmpl;
+    end if;
+end;
+$$;
+
+-- Compatibility view named pgque.subscription — same shape as upstream PgQ's
+-- table so the rest of the SQL layer keeps working unchanged.
+-- Only the *active* child is exposed; the non-active children hold stale row
+-- versions from previous rotations and must not be visible to readers.
+create or replace view pgque.subscription as
+      select 0::smallint as sub_child_table, s.*
+        from pgque.subscription_0 s
+       where (select cur_subscription_table from pgque.meta_rotation) = 0
+union all
+      select 1::smallint, s.*
+        from pgque.subscription_1 s
+       where (select cur_subscription_table from pgque.meta_rotation) = 1
+union all
+      select 2::smallint, s.*
+        from pgque.subscription_2 s
+       where (select cur_subscription_table from pgque.meta_rotation) = 2;
+
+-- ======================================================================
+-- Step 4: Routing trigger functions for the views
+-- ======================================================================
+
+-- pgque._subscription_route() -- INSTEAD OF trigger for pgque.subscription view.
+-- Routes INSERT/UPDATE/DELETE to the currently-active child table.
+create or replace function pgque._subscription_route()
+returns trigger language plpgsql as $$
+declare
+    v_cur smallint;
+begin
+    select cur_subscription_table into v_cur from pgque.meta_rotation;
+    if tg_op = 'INSERT' then
+        if v_cur = 0 then
+            insert into pgque.subscription_0
+                (sub_id, sub_queue, sub_consumer, sub_last_tick, sub_active,
+                 sub_batch, sub_next_tick)
+            values (coalesce(new.sub_id, nextval('pgque.subscription_sub_id_seq')),
+                    new.sub_queue, new.sub_consumer, new.sub_last_tick,
+                    coalesce(new.sub_active, now()), new.sub_batch, new.sub_next_tick);
+        elsif v_cur = 1 then
+            insert into pgque.subscription_1
+                (sub_id, sub_queue, sub_consumer, sub_last_tick, sub_active,
+                 sub_batch, sub_next_tick)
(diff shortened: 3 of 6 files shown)


NikolayS:
### Follow-up suggestion: decouple metadata rotation cadence from event rotation

The PR currently ties subscription/tick rotation to the existing `queue_rotation_period` (same knob that drives event-table rotation). That works and passes the acceptance criteria in the [R6 combined smoke](https://gitlab.com/postgres-ai/postgresql-consulting/tests-and-benchmarks/-/issues/77#note_3263507358), but the two have very different write rates and different optimal rotation cadences.

**Observed in R6 smoke (benchmark `queue_rotation_period=30s`)**:

- Event tables: ~1000 INSERTs/s → at 30 s cadence each partition sees ~30 k rows → rotation at 30 s is appropriate.
- `pgque.subscription`: ~2 UPDATEs/s (only on `finish_batch` when a batch has events — not on every ticker call) → at 30 s cadence each rotation-side accumulates only ~60 dead tuples. Way under target (500).
- `pgque.tick`: rate depends on ticker cadence, in practice ~1 row/s → ~30 dead per 30 s rotation. Also way under target (200).

So metadata rotation at 30 s is 4-6× more frequent than it needs to be. Costs:

- **pg_class churn**: each rotation cycle calls TRUNCATE on 3 metadata tables → ~3 `relfilenode` UPDATEs on pg_class per cycle. At 30 s cadence = ~360 dead pg_class tuples/hour. Over a 24 h production workload = ~8 640 dead pg_class rows. pg_class is hit by every query planner lookup — its bloat affects ALL queries, not just pgque.
- **TRUNCATE AccessExclusive lock**: brief (small table) but non-zero. 3 locks per cycle × 120 cycles/hour = 360 brief blockers/hour.
- **WAL per TRUNCATE**: small but measurable over long runs.

### Proposal

Add a separate config field on `pgque.queue`:

```sql
ALTER TABLE pgque.queue
  ADD COLUMN IF NOT EXISTS queue_metadata_rotation_period interval NOT NULL DEFAULT '2 minutes';
```

Default to 2 min; the rotation trigger in the ticker uses this field for subscription/tick rotation specifically, while event-table rotation keeps using `queue_rotation_period`.

At 2 min cadence the expected peak bloat is:

| Table | Write rate | Peak dead per cycle | vs target |
|---|---:|---:|---|
| `subscription` | ~2 /s | ~240 | well under 500 ✓ |
| `tick` | ~1-2 /s | ~120-240 | under 200-500 depending on workload |

Still meets acceptance criteria while cutting pg_class bloat and TRUNCATE churn by 4×.

Benchmark override: `UPDATE pgque.queue SET queue_metadata_rotation_period='30 seconds'` — preserves stress-test behavior.

### Rationale for separation

Event tables are bulk-storage tables where rotation = capacity management (drop old events you no longer need). Metadata tables are tiny-hot tables where rotation = bloat mitigation. Same mechanism, different economics — deserve independent knobs.

### Scope for this PR vs follow-up

This is a design-polish suggestion, not a correctness blocker for landing the current PR. Options:

1. Add the `queue_metadata_rotation_period` field in this PR as a small addition (+10 LoC) with default `'2 minutes'`. Bench override sets it to 30 s explicitly.
2. Land this PR as-is and open a follow-up issue/PR specifically for the decoupled cadence.

I'd lean toward (1) since the acceptance criteria for the metadata-bloat fix ([pgque#61](https://github.com/NikolayS/pgque/issues/61)) arguably includes "reasonable default cadence for production" — and 30 s matching events is a debatable default.

Either way, the rotation design itself is validated by the R6 smoke; this is a tuning refinement.

NikolayS:
## REV Code Review — #62

**Summary:** Fixes #61 by extending PgQ's 3-table rotation pattern to `pgque.subscription` and `pgque.tick` — the two PgQ metadata tables that were never rotated — so held-xmin bloat on them is bounded instead of unbounded. Tables become `UNION ALL` views over three physical children (`*_0 / *_1 / *_2`), an `INSTEAD OF` trigger routes writes to the currently-active child (tracked in new singleton `pgque.meta_rotation`), and `pgque.maint_rotate_metadata` + its `_step2` counterpart do the rotate-truncate-copy cycle on the existing `pgque_rotate_step2` cron. 30 s default cadence, tunable via GUC `pgque.meta_rotation_period`. +552/-29 across two files.

Installed on PG18 locally — full `tests/run_all.sql` passes, and a forced-rotation smoke test (described below under **Verified**) walks the pointer `0 → 1 → 2 → 0` correctly and preserves consumer semantics.

---

### BLOCKING — 0

### NON-BLOCKING

**MEDIUM** No new regression tests for rotation logic
> The diff touches two files — `sql/pgque.sql` and `sql/pgque_uninstall.sql`. Nothing under `tests/`. The only item in the test plan that specifically exercises the new behavior (R6 smoke test) is unchecked, and its checking happens outside the repo's test suite. That means CI on PG 14–18 doesn't actually prove rotation is correct; it only proves the existing send/receive/dlq paths still work on top of the new storage layout — which isn't the same thing.
>
> Minimum regression coverage to add:
> 1. Force 3 full rotations (`cur = 0 → 1 → 2 → 0`) and assert `pgque.subscription` returns exactly one row per `(sub_queue, sub_consumer)` throughout.
> 2. Between rotations, run a full send → ticker → next_batch → finish_batch cycle to prove writes route to the active child and reads resolve via the view.
> 3. Create a consumer, let it lag so its `sub_last_tick` points at a specific tick slot, force rotation, assert the tick rotation defers (not truncates) and the lagged consumer can still `receive()`.
> 4. Call `maint_rotate_metadata()` twice without an intervening `_step2` and assert the second call returns `0` (the held-xmin gate).
>
> I ran a version of tests 1 and 2 manually on this branch and they passed — `cur_sub` cycled through 0/1/2/0 across three forced rotations, the view kept showing 2 rows with 2 consumers, and `c1` / `c2` both cycled receive+ack cleanly on each rotation. But "worked on Nik's laptop once" ≠ "covered in CI against PG 14–18 on every future PR." Given how central this change is, a `tests/test_metadata_rotation.sql` added to `run_all.sql` is load-bearing for future refactor safety, not optional polish.

**MEDIUM** `sql/pgque.sql` L2406-L2418 (`unregister_consumer`) — `FOR UPDATE OF s` silently dropped, correctness justification under-argued
> Original:
> ```sql
>    for update of s, c;
> ```
> New:
> ```sql
>    for update of c;
> ```
> with the comment: "Callers of unregister_consumer() are rare and single-threaded in practice; we rely on the subsequent DELETE to provide the necessary row lock."
>
> "Rare and single-threaded in practice" is a behavior claim, not a correctness proof. The DELETE does take a row lock — but it's a DELETE on the view, which fires the INSTEAD OF trigger, which DELETEs from all three children (`subscription_0/1/2`). If two concurrent `unregister_consumer` calls race on the same (queue, consumer), both pass the SELECT without blocking, both fire the DELETE trigger, and we rely on the DELETE being idempotent — which it is (the `where sub_queue=... and sub_consumer=...` is a no-op on the second one). So operationally it probably works out.
>
> But: `(s.sub_last_tick IS NULL AND s.sub_next_tick IS NULL) OR ...` is computed in the SELECT and used further down the function body. If one of the racing transactions sees that condition as `true` while the other sees `false` (because the row state is mid-flight from an unrelated `finish_batch`), downstream logic that branches on `_is_subconsumer` could take different paths, and the function body after the snippet cut off here might depend on that.
>
> **Ask:** either (a) add a test that runs two `unregister_consumer` calls against the same consumer concurrently and asserts end-state is clean, or (b) consider taking a short advisory lock on (sub_queue, sub_consumer) as a replacement for the dropped row lock. The comment needs to describe the correctness argument, not just the expected concurrency regime.

**LOW** `sql/pgque.sql` L1300-L1490 (`maint_rotate_metadata`) — `if mr.last_rotation_step2_txid is null then return 0 end if` isn't quite the "held-xmin gate" the comment describes
> The comment is 15 lines of careful reasoning about held-xmin trade-offs, ending with:
> ```sql
> if mr.last_rotation_step2_txid is null then
>     -- previous rotation's step2 has not yet been observed;
>     -- not safe to rotate again.
>     return 0;
> end if;
> ```
> But this only checks whether `_step2` has executed in some prior transaction — it does not inspect any backend's xmin. An external long transaction that predates the previous `_step2` and holds an old xmin will still see stale rows on whatever child gets truncated next cycle. That's a trade-off the comment acknowledges further up ("External queries that read pgque.subscription under a long snapshot may see stale rows or no rows after the prev-prev slot is truncated") — but the variable name `last_rotation_step2_txid` and the "held xmin" framing of the nearby code make it read as if that were being gated, when it isn't.
>
> **Fix:** rename the field to something like `last_step2_txid` (no "held xmin" framing) and shorten the gate comment to: "step2 must have run in a separate transaction so the pointer flip is visible to new snapshots — this is a sequencing gate, not a held-xmin gate. Truncating the prev-prev slot is intentionally not gated on backend_xmin (see file header)."

**LOW** `sql/pgque.sql` L165 — redundant FK constraint on `pgque.tick_tmpl`
> ```sql
> create table if not exists pgque.tick_tmpl (
>     ...
>     constraint tick_tmpl_queue_fkey foreign key (tick_queue)
>                                references pgque.queue (queue_id)
> );
> ```
> The children use `LIKE pgque.tick_tmpl INCLUDING DEFAULTS`, which does **not** copy constraints (would need `INCLUDING ALL` or `INCLUDING CONSTRAINTS`). Each child explicitly redeclares its own `tick_N_queue_fkey` FK, so `tick_tmpl`'s FK is dead weight — it constrains a table that by design will never hold rows.
> **Fix:** drop the constraint from `tick_tmpl`. Same harmless-but-dead pattern to check on `subscription_tmpl` — which already omits the FKs, FWIW, so the right call here was the inconsistent one.

**LOW** `sql/pgque.sql` L4552 — `pgque_rotate_step2` cron command concatenates two SELECTs; if the first fails the second never runs
> ```sql
> $sql$SELECT pgque.maint_rotate_tables_step2();
>      SELECT pgque.maint_rotate_metadata_step2();$sql$
> ```
> Each `_step2` is defensive (it updates a single config row), so an outright failure is unlikely. But if `maint_rotate_tables_step2()` ever raises (e.g. lock timeout from a concurrent `drop_queue`), `maint_rotate_metadata_step2()` gets skipped that cycle → next rotation hits the `step2_txid IS NULL` gate and is deferred. In pathological cases with recurring step-1 failures, metadata rotation could stall without any log-visible reason.
> **Fix:** either split into two cron jobs (`pgque_rotate_step2_events`, `pgque_rotate_step2_metadata`), or wrap each call in a `DO $$ BEGIN ... EXCEPTION WHEN OTHERS THEN ... END $$` block so one failure doesn't cascade. Two-cron jobs is simpler and cleaner.

### POTENTIAL ISSUES

**MEDIUM** Hot-path overhead: every `finish_batch` / `next_batch_custom` now goes through an `INSTEAD OF` PL/pgSQL trigger plus a UNION-ALL view with a scalar subquery reading `pgque.meta_rotation` on each of its three branches
> The acceptance criterion is "within 20% of clean baseline," which is honest about the cost. But the per-statement breakdown is:
> 1. View resolution: `(select cur_subscription_table from pgque.meta_rotation)` evaluated once per UNION branch (3×) per query.
> 2. Predicate `= 0` / `= 1` / `= 2` filters two of three branches to empty — but only after the subquery evaluates.
> 3. INSTEAD OF trigger fires per row, runs PL/pgSQL branching on `v_cur`, then does an UPDATE on the child.
>
> For `finish_batch`, that's one trigger fire + one underlying UPDATE per ack. For `next_batch_custom`, same pattern on the UPDATE side. At 3 500 iter-TPS (the baseline in the PR description), the PL/pgSQL overhead is real. `R6 smoke test` results will say how much of it materializes.
>
> **Possible follow-ups** (not for this PR): (a) materialize `cur_subscription_table` into a session GUC that rotation bumps and cron refreshes — avoids the per-query scalar subquery; (b) rewrite the INSTEAD OF trigger in C via PL/v8 or an extension (doesn't fit "no C accelerator" principle, noted for completeness); (c) use a partitioned table with a rotating partition-pruned predicate instead of a view.

**LOW** The subscription view's `where (select cur_subscription_table from pgque.meta_rotation) = N` predicate is re-evaluated per transaction snapshot, which means READ COMMITTED callers see rotation happen mid-transaction
> In RC mode, each statement gets a fresh snapshot, so back-to-back statements within one transaction can read different `cur` values. Operationally this appears correct: `finish_batch` finds the row by `sub_batch`, which is unique across children at any instant (because rotation copies and flips atomically under EXCLUSIVE lock). But if a caller ever does `SELECT ... FROM pgque.subscription` and uses the result across statements, they might see ghosts. Not this PR's bug; flagging because the commit note says "pgque itself has no code path that reads pgque.subscription under a long-lived snapshot" — which is a statement that needs a watch going forward (new code that reads subscription across statements + relies on row identity would silently break).

**INFO** The tick rotation's "defer if any consumer's sub_last_tick lives in the target slot" rule means a single stuck consumer can block tick rotation indefinitely while subscription rotation keeps happening
> Author explicitly accepts this ("Subscription rotation still happens in that case — which is the dominant bloat source"). Worth instrumenting: a metric / log line on consecutive deferrals would make the operational picture observable. The `pgque.meta_rotation` singleton could grow a `tick_rotation_deferred_cycles bigint` counter. Not this PR's scope; just noting so it doesn't fall off.

---

### Verified on a local PG18 install

- Full `tests/run_all.sql` passes (send, receive, dlq, roles) — so the view+trigger layer is transparent to every test that already exists. ✓
- Forced 3 rotations (`update pgque.meta_rotation set last_rotation_time = now() - interval '1 hour'` + call `maint_rotate_metadata` + `maint_rotate_metadata_step2`). `cur_subscription_table` cycles through `0 → 1 → 2 → 0`. ✓
- After each rotation, view row count stays at the right number (2 consumers → 2 rows in the view), while the prev children retain their stale copies until they become the target of the next rotation. ✓
- Inter-rotation `pgque.receive()` + `pgque.ack()` on both consumers works — writes route correctly to whichever child is currently active. ✓
- Post-full-cycle (`cur=0` again after 3 rotations), insert + ticker + receive + ack on `c2` returns the fresh event. ✓
- `pgque.subscription_0/1/2` dead-tuple count sanity: on this scratch instance with 2 consumers it stays at 0/2/2 shape as expected, matching the "active child is fresh" claim.

No crashes, no FK violations, no grant errors on install, no issues with the INSTEAD OF triggers. The design holds together in practice.

### Summary

|---|---|---|
| Security | 0 | 0 |
| Bugs | 0 | 0 |
| Tests | 1 MEDIUM | 0 |
| Correctness/Concurrency | 1 MEDIUM (unregister lock), 1 LOW (gate framing) | 1 LOW (RC snapshot) |
| Performance | 0 | 1 MEDIUM (hot-path trigger overhead) |
| Guidelines | 0 | 0 |
| Docs | 0 | 0 |
| Operational | 0 | 1 INFO (deferred-tick observability) |
| Cleanup | 2 LOW (redundant FK, cron command split) | 0 |

**Result: REQUEST CHANGES (non-blocking).** The design is correct, the manual validation holds, and the existing test suite keeps passing. But shipping a change of this scope without a dedicated `tests/test_metadata_rotation.sql` in the `run_all.sql` CI path is a maintainability risk the merge shouldn't absorb — the 3-rotation cycle test I ran by hand in 2 minutes belongs in the tree. The `unregister_consumer` `FOR UPDATE` removal wants an explicit concurrency argument (or a test) rather than a "rare in practice" note. And the cron-command concatenation at L4552 is one-character's worth of risk (split into two jobs) that would've caught a dependent-failure bug long before anyone noticed. R6 smoke results will close out the perf question independently.

---

NikolayS:
## Rebase complete — `8acc53f` on current main

**Strategy:** Could not cleanly rebase `a9e992f` onto main (55-commit divergence, 4 000-line generated-file conflict). Instead: re-ported the logic from `/tmp/pr62.diff` directly into source-of-truth files and regenerated, per CLAUDE.md "Keep generated files and source changes consistent."

---

### Files changed

| File | Role |
|------|------|
| `sql/pgque-additions/metadata_rotation.sql` | **New.** All rotation DDL and functions live here — this is the canonical source. |
| `sql/pgque-additions/lifecycle.sql` | `pgque_rotate_step2` cron job now also calls `maint_rotate_metadata_step2()`. |
| `sql/pgque-api/maint.sql` | `maint()` skips `maint_rotate_metadata_step2` (needs separate tx, same pattern as `maint_rotate_tables_step2`). |
| `build/transform.sh` | Adds `metadata_rotation.sql` to the assembly loop (after `dlq.sql`). |
| `sql/pgque.sql`, `sql/pgque-tle.sql` | Regenerated — not hand-edited. |

`sql/pgque.sql` is now derived, not the source of truth for this feature.

---

### What `metadata_rotation.sql` contains
- `pgque.meta_rotation` singleton table (rotation pointer)
- `pgque.tick_tmpl`, `tick_0/1/2` child tables + `pgque.tick` UNION ALL view
- `pgque.subscription_tmpl`, `subscription_0/1/2` child tables + `pgque.subscription` UNION ALL view
- `pgque._subscription_route()` and `pgque._tick_route()` INSTEAD OF trigger functions
- `pgque.maint_rotate_metadata()` and `pgque.maint_rotate_metadata_step2()` — both SECURITY DEFINER with `set search_path = pgque, pg_catalog`
- Override of `maint_tables_to_vacuum()` (lists child tables, not views)
- Override of `maint_operations()` (emits both metadata rotation entries)
- Override of `unregister_consumer()` (drops `FOR UPDATE OF s` — view can't be locked; `FOR UPDATE OF c` kept)
- Explicit `SELECT` grants on the new views to `public` (replacing the OID-renamed base-table grants from `grants.sql`)

---

### Test results (PG 16, local)
- **Regression (`tests/run_all.sql`):** `=== ALL TESTS PASSED ===`
- **Acceptance (`tests/acceptance/run_acceptance.sql`):** `=== ALL ACCEPTANCE TESTS PASSED ===`
- **Idempotent reinstall:** no ERROR/FATAL output on second `\i sql/pgque.sql`
- **Functional smoke:** rotation returns 0 when too-soon, returns 1 after backdating `last_rotation_time`; pointer flips from `cur=0` to `cur=1` correctly

---

### Caveats / things to review

1. **`subscription_tmpl` / `tick_tmpl` tables persist** after install — they are empty but sit in the schema. This is intentional (column-schema carrier for `LIKE … INCLUDING DEFAULTS`), but could be dropped after child tables are created if preferred.
2. **`FOR UPDATE OF s` removal in `unregister_consumer`** is safe for the single-consumer unregister path (DELETE provides the lock), but worth a second pair of eyes given PgQ's subconsumer logic.
3. **R6-style benchmark (dead-tuple bound under held xmin) not yet run** — acceptance criteria from #61 (`subscription dead-tuples ≤ 500`, `tick dead-tuples ≤ 200`, iter-TPS within 20%) require a long-running soak test, not included here.
4. **Tick view `tick_child_table` extra column** is visible to code that does `SELECT *` from `pgque.tick`. PgQ code only ever selects named columns, but worth noting.

---

NikolayS:
## samorev Code Review Report

- **PR:** NikolayS/PgQue#62 - Rotate subscription and tick tables to avoid held-xmin bloat (#61)
- **Author:** @NikolayS
- **AI-Assisted:** Unknown

| Pipeline | Coverage |
|----------|----------|
| PASS | Not reported |

---

### BLOCKING ISSUES (1)

---

### Summary

| CI/Pipeline | 0 | 0 | 0 |
| Security | 0 | 0 | 0 |
| Bugs | 0 | 0 | 0 |
| Tests | 0 | 0 | 0 |
| Guidelines | 0 | 0 | 0 |
| Docs | 0 | 0 | 0 |
| Metadata | 1 | 0 | 0 |

Note:

<details>

```text
provider=github
kind=pr
project=NikolayS/PgQue
number=62
target=github:NikolayS/PgQue#62
state=OPEN
draft=true
diff_lines=2615
diff_added=2519
diff_removed=7
diff_bytes=109899
comments_count=3
commits_count=1
ci_status=success
ci_summary=total=10 success=10 failure=0 pending=0 other=0
blocking=false
posted_by=gh
no_comment=false
live_posting=posted
```

</details>

---

PR #63, opened 2026-04-18
fix: schedule pgque.maint_retry_events() from pgque.start()

NikolayS:
## Problem

\`pgque.start()\` schedules three cron jobs: \`pgque_ticker\`, \`pgque_maint\`, \`pgque_rotate_step2\`. But \`pgque.maint()\` delegates to \`maint_operations()\`, and \`maint_operations()\` does **not** include \`pgque.maint_retry_events\`. So events a consumer nacks get inserted into \`pgque.retry_queue\` and then sit there until someone manually calls \`pgque.maint_retry_events()\` — they are never moved back into the main event stream for redelivery.

On any default-install deployment using \`nack()\` (the documented retry path), **retries are silently broken**. The DLQ routing still works because \`nack()\` calls \`event_dead\` directly inside its body; but anything that does not exceed \`max_retries\` is stranded.

Kukushkin's 2026 "Rediscovering PgQ" deck shows the pgqd schedule cleanly: ticker 1s, retries 30s, maintenance 120s. We mirror the first two today but fused retries into the wrong function.

Found during the docs restructure (PR #59) — the tutorial had to call \`maint_retry_events()\` explicitly to show a retry loop. That was the smoking gun.

## Fix

Add a fourth cron job in \`pgque.start()\`:

| Job | Cadence | Function |
|---|---|---|
| \`pgque_ticker\` | 2 s | \`pgque.ticker()\` |
| \`pgque_retry_events\` | **30 s (new)** | \`pgque.maint_retry_events()\` |
| \`pgque_maint\` | 30 s | \`pgque.maint()\` (rotation step 1 + vacuum) |
| \`pgque_rotate_step2\` | 10 s | \`pgque.maint_rotate_tables_step2()\` |

Cadence matches pgqd for retries (30 s). We keep \`pgque_maint\` at 30 s for now; whether to drop it to 60 or 120 s (pgqd default) is a separate design question, out of scope for this PR.

### Changes

- \`sql/pgque-additions/lifecycle.sql\`:
  - \`pgque.start()\` schedules \`pgque_retry_events\` alongside the other three jobs. Job id is not stored in \`pgque.config\`; unscheduled by name in \`pgque.stop()\` like \`pgque_rotate_step2\`.
  - \`pgque.stop()\` unschedules \`pgque_retry_events\` before \`pgque_rotate_step2\`.
  - Start notice lists all four jobs.
- \`tests/test_pgcron_lifecycle.sql\`:
  - Test 1 now asserts all four \`pgque_*\` jobs exist in \`cron.job\` after \`start()\`.
  - New Test 1b asserts all four are gone after \`stop()\`.
- \`sql/pgque.sql\` rebuilt via \`build/transform.sh\`.

## Test plan

- [x] \`build/transform.sh\` rebuilds cleanly.
- [x] Four \`pgque_*\` cron jobs after \`start()\`; zero after \`stop()\` (assertions added).
- [ ] CI on PG 14–18 confirms.
- [ ] Manual verification: \`nack\` event with \`0 seconds\` retry delay, let \`pgque_retry_events\` fire, confirm event reappears on the next tick.

## Docs impact

PR #59 (docs restructure) documents \`pgque_maint\` as "(rotation step 1 and vacuum)" — accurate against both the pre- and post-fix surface. Once this merges and #59 rebases, the tutorial's Production cadence subsection will get a one-line update to mention the fourth job. No wording change to this PR's description needed.

### DIFF
--- a/sql/pgque-additions/lifecycle.sql
+++ b/sql/pgque-additions/lifecycle.sql
@@ -5,6 +5,7 @@ create or replace function pgque.start()
 returns void as $$
 declare
     v_ticker_id bigint;
+    v_retry_id bigint;
     v_maint_id bigint;
     v_step2_id bigint;
     v_dbname text;
@@ -29,7 +30,18 @@ begin
         v_dbname
     ) into v_ticker_id;
 
-    -- Maintenance: every 30 seconds (rotation step1, retry, vacuum)
+    -- Retry events: every 30 seconds (move nack'd events from the retry
+    -- queue back into the main event stream for the next tick).
+    -- pgque.maint() / maint_operations() does NOT include retry handling,
+    -- so this has to be scheduled separately — matches pgqd cadence.
+    select cron.schedule_in_database(
+        'pgque_retry_events',
+        '30 seconds',
+        $sql$SET statement_timeout = '25s'; SELECT pgque.maint_retry_events()$sql$,
+        v_dbname
+    ) into v_retry_id;
+
+    -- Maintenance: every 30 seconds (rotation step 1 and vacuum).
     select cron.schedule_in_database(
         'pgque_maint',
         '30 seconds',
@@ -47,13 +59,13 @@ begin
         v_dbname
     ) into v_step2_id;
 
-    -- Store job IDs in config
+    -- Store job IDs in config (retry + rotate_step2 unscheduled by name)
     update pgque.config
     set ticker_job_id = v_ticker_id,
         maint_job_id = v_maint_id;
 
-    raise notice 'pgque started: ticker job=%, maint job=%, rotate_step2 job=%',
-        v_ticker_id, v_maint_id, v_step2_id;
+    raise notice 'pgque started: ticker=%, retry_events=%, maint=%, rotate_step2=%',
+        v_ticker_id, v_retry_id, v_maint_id, v_step2_id;
 end;
 $$ language plpgsql security definer set search_path = pgque, pg_catalog;
 
@@ -83,6 +95,14 @@ begin
         if v_maint_id is not null then
             perform cron.unschedule(v_maint_id);
         end if;
+
+        -- Unschedule retry_events by name (job ID not stored in config).
+        -- Ignore if job doesn't exist (first run or already removed).
+        begin
+            perform cron.unschedule('pgque_retry_events');
+        exception when others then
+            raise notice 'pgque.stop: retry_events job not found (OK on first install)';
+        end;
 
         -- Unschedule rotate_step2 by name (job ID not stored in config)
         -- Ignore if job doesn't exist (first run or already removed)
--- a/sql/pgque-additions/lifecycle.sql
+++ b/sql/pgque-additions/lifecycle.sql
@@ -37,7 +37,7 @@ begin
     select cron.schedule_in_database(
         'pgque_retry_events',
         '30 seconds',
-        $sql$SET statement_timeout = '25s'; SELECT pgque.maint_retry_events()$sql$,
+        $sql$set statement_timeout = '25s'; select pgque.maint_retry_events()$sql$,
         v_dbname
     ) into v_retry_id;
 
--- a/sql/pgque.sql
+++ b/sql/pgque.sql
@@ -4028,7 +4028,7 @@ begin
     select cron.schedule_in_database(
         'pgque_retry_events',
         '30 seconds',
-        $sql$SET statement_timeout = '25s'; SELECT pgque.maint_retry_events()$sql$,
+        $sql$set statement_timeout = '25s'; select pgque.maint_retry_events()$sql$,
         v_dbname
     ) into v_retry_id;
 
--- a/sql/pgque.sql
+++ b/sql/pgque.sql
@@ -3996,6 +3996,7 @@ create or replace function pgque.start()
 returns void as $$
 declare
     v_ticker_id bigint;
+    v_retry_id bigint;
     v_maint_id bigint;
     v_step2_id bigint;
     v_dbname text;
@@ -4020,7 +4021,18 @@ begin
         v_dbname
     ) into v_ticker_id;
 
-    -- Maintenance: every 30 seconds (rotation step1, retry, vacuum)
+    -- Retry events: every 30 seconds (move nack'd events from the retry
+    -- queue back into the main event stream for the next tick).
+    -- pgque.maint() / maint_operations() does NOT include retry handling,
+    -- so this has to be scheduled separately — matches pgqd cadence.
+    select cron.schedule_in_database(
+        'pgque_retry_events',
+        '30 seconds',
+        $sql$SET statement_timeout = '25s'; SELECT pgque.maint_retry_events()$sql$,
+        v_dbname
+    ) into v_retry_id;
+
+    -- Maintenance: every 30 seconds (rotation step 1 and vacuum).
     select cron.schedule_in_database(
         'pgque_maint',
         '30 seconds',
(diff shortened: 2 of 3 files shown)


tests/test_pgcron_lifecycle.sql
```diff
@@ -23,10 +24,39 @@ begin
...
-  raise notice 'PASS: start() sets job IDs (ticker=%, maint=%)', v_ticker_id, v_maint_id;
+  raise notice 'PASS: start() schedules four jobs (ticker, retry_events, maint, rotate_step2)';
+end $$;
+
+-- Test 1b: stop() unschedules every pgque_* job (ticker, retry_events, maint, rotate_step2)
+do $$
+declare
+  v_job_count int;
+begin
+  if not exists (select 1 from pg_extension where extname = 'pg_cron') then
```
NikolayS:
```suggestion
  if not exists (select from pg_extension where extname = 'pg_cron') then
```

sql/pgque.sql
```diff
@@ -4020,7 +4021,18 @@ begin
...
 
-    -- Maintenance: every 30 seconds (rotation step1, retry, vacuum)
+    -- Retry events: every 30 seconds (move nack'd events from the retry
+    -- queue back into the main event stream for the next tick).
+    -- pgque.maint() / maint_operations() does NOT include retry handling,
+    -- so this has to be scheduled separately — matches pgqd cadence.
+    select cron.schedule_in_database(
+        'pgque_retry_events',
+        '30 seconds',
+        $sql$SET statement_timeout = '25s'; SELECT pgque.maint_retry_events()$sql$,
```
NikolayS:
```suggestion
        $sql$set statement_timeout = '25s'; select pgque.maint_retry_events()$sql$,
```

NikolayS:
## REV Code Review — #63

**Summary:** Fixes a real bug in the default install: `pgque.start()` schedules `pgque_ticker`, `pgque_maint`, `pgque_rotate_step2`, but `pgque.maint()` (via `maint_operations()`) does not include `maint_retry_events`. Nack'd events land in `pgque.retry_queue` and never move back to the main stream — retries silently broken for every `nack()` caller under `max_retries`. PR adds a fourth `pgque_retry_events` cron job at 30s cadence (matching pgqd), updates `stop()` to unschedule it, updates the start notice, and asserts all four jobs in `test_pgcron_lifecycle.sql`. +80/-10 across three files.

---

### BLOCKING — 0

### NON-BLOCKING — 0

### POTENTIAL ISSUES

**LOW** `sql/pgque.sql:823-866` (pre-existing, surfaced by the promotion) — `maint_retry_events()` LIMIT 10 + 30s cadence caps retry throughput at ~20/min per instance
> The underlying PgQ function processes at most 10 rows per invocation (`sql/pgque.sql:853`), and the function's own docstring says "It should be called until it returns 0" (line 831). This cron wires it up as a single-shot call per period, not a drain-until-zero loop. At 30s cadence, max steady-state retry throughput is ~20 events/minute. Matches pgqd's default `retry_period=30` behavior so it's not a regression, but for the first time this behavior is now on by default for every install.
>
> For high-throughput workloads that rely on `nack()` + retry (e.g. transient external-API failures with heavy fan-out), this ceiling will be felt. Not this PR's problem to solve — the bug is "retries are silently dropped entirely," which this fixes — but worth filing as a follow-up.
>
> **Possible follow-up:** either (a) drain-in-loop inside the cron command with a time budget (`while pgque.maint_retry_events() > 0 and clock_timestamp() - start < '20s' loop … end loop`), or (b) document the LIMIT and let operators lower the cron period for high-retry workloads, or (c) parameterize the LIMIT in `maint_retry_events()`. All three are design decisions beyond this patch's scope.

**LOW** `sql/pgque-additions/lifecycle.sql:61-65` — half the jobs track by `job_id`, half by `jobname`, and the split is load-bearing
> `ticker_job_id` and `maint_job_id` are stored in `pgque.config` and unscheduled by numeric id in `stop()`. `pgque_retry_events` (new) and `pgque_rotate_step2` are unscheduled by name (`cron.unschedule('pgque_retry_events')`). The comment even calls this out: `-- retry + rotate_step2 unscheduled by name`. Two consequences:
> 1. If someone later swaps which jobs get stored in config, `stop()` could orphan jobs silently (the current exception-swallowing in the name-based paths would hide it).
> 2. Any external tooling introspecting `pgque.config` to find the cron schedule sees two of four jobs.
>
> Not a bug here — matches the established convention that `rotate_step2` uses — but noting that now that *two* of the four jobs use the name-based path, the inconsistency is harder to dismiss as a one-off. A small cleanup PR moving all four to name-based unschedule (or storing all four ids in config) would remove a category of future foot-gun.

---

### Verified against `sql/pgque.sql`

- `maint_operations()` at `sql/pgque.sql:1043-1120` indeed does not yield `pgque.maint_retry_events` — only step1/step2 rotation, `queue_extra_maint` hooks, and vacuum. The stated bug is real. ✓
- `pgque.maint()` at `sql/pgque.sql:4479` iterates `maint_operations()` without a retry hook, confirming the gap. ✓
- `pgque.maint_retry_events()` exists at `sql/pgque.sql:823-866` with signature `returns integer`, callable with no args — matches the `SELECT pgque.maint_retry_events()` in the cron command. ✓
- `SET statement_timeout = '25s'` in the command is a sensible 5s safety margin under the 30s cadence — protects against lock contention with `pgque_maint` (which also fires at 30s and takes vacuum locks). ✓
- `cron.schedule_in_database(job_name, schedule, command, database)` — 4-arg form; username defaults to the caller's `CURRENT_USER`. Since `pgque.start()` is SECURITY DEFINER, `CURRENT_USER` at schedule time is the function owner (the install principal). That owner has DELETE on `pgque.retry_queue` and LOCK privileges, so the scheduled job will have the privileges it needs. ✓
- Stop-by-name block (lines 100-105 / 4091-4096) correctly mirrors the existing `pgque_rotate_step2` pattern, including the `exception when others` swallow for first-run safety. ✓
- Notice message now lists all four jobs (`ticker=%, retry_events=%, maint=%, rotate_step2=%`). ✓
- Tests: Test 1 checks job count = 4 post-`start()` by `jobname in (…)`; Test 1b checks count = 0 post-`stop()`. Both gate on `pg_extension where extname = 'pg_cron'` so they auto-skip when pg_cron isn't installed. Complete coverage of the new behavior. ✓
- `sql/pgque.sql` hunks mirror `sql/pgque-additions/lifecycle.sql` line-for-line → transform.sh rebuild was clean. ✓

### Summary

|---|---|---|
| Security | 0 | 0 |
| Bugs | 0 | 1 LOW (pre-existing throughput ceiling) |
| Tests | 0 | 0 |
| Guidelines | 0 | 1 LOW (job id/name split) |
| Docs | 0 | 0 |

**Result: APPROVE.** Correct, surgical fix for a real retries-silently-broken bug. The 30s cadence matches the upstream pgqd reference, the statement_timeout is appropriately defensive, and the lifecycle test assertions cover both start and stop paths (including auto-skip when pg_cron isn't present). The two potential items are follow-up territory, not merge-blockers.

---

sql/pgque-additions/lifecycle.sql
```diff
@@ -29,7 +30,18 @@ begin
...
 
-    -- Maintenance: every 30 seconds (rotation step1, retry, vacuum)
+    -- Retry events: every 30 seconds (move nack'd events from the retry
+    -- queue back into the main event stream for the next tick).
+    -- pgque.maint() / maint_operations() does NOT include retry handling,
+    -- so this has to be scheduled separately — matches pgqd cadence.
+    select cron.schedule_in_database(
+        'pgque_retry_events',
+        '30 seconds',
+        $sql$SET statement_timeout = '25s'; SELECT pgque.maint_retry_events()$sql$,
```
NikolayS:
```suggestion
        $sql$set statement_timeout = '25s'; select pgque.maint_retry_events()$sql$,
```

PR #75, opened 2026-04-19
fix: multiple bug fixes and dead code cleanup

hobostay:
## Summary

- **Go consumer fixes**: Add missing `Nack()` method; use per-message nack for handler failures instead of skipping the entire batch; log warning for unregistered event types instead of silently dropping messages
- **Security hardening**: Use `quote_ident()` in `maint()` for VACUUM target; add explicit queue existence check in `insert_event_raw()` with clear error message
- **Atomicity fix**: `dlq_replay_all()` now wraps per-event replay in exception blocks so a single failure doesn't abort the entire operation
- **Role hierarchy fix**: `upgrade_schema()` had inverted role membership — `pgque_admin` was created as member of reader/writer instead of the reverse; also fixed return value (was hardcoded `0`, now returns actual count)
- **Dead code cleanup**: Remove PG 8.3 `pg_autovacuum` code path (PgQue requires PG14+); remove references to non-existent schemas (`pgq_ext`, `pgq_node`, `londiste`, `txid`) from `maint_tables_to_vacuum()`
- **Typo fixes**: `consubers`→`consumers`, `poin`→`point`, `oldes`→`oldest`, `wuith`→`with`, `accitentally`→`accidentally`, `Quete`→`Quote`

## Test plan

- [x] Go code compiles and passes `go vet`
- [ ] Run existing SQL test suite against modified `pgque.sql`
- [ ] Verify `upgrade_schema()` role hierarchy is correct (pgque_reader → pgque_writer → pgque_admin)
- [ ] Test `dlq_replay_all()` with a concurrent queue drop to verify exception handling
- [ ] Verify `insert_event_raw()` raises clear error for non-existent queue

### DIFF
--- a/sql/pgque-api/maint.sql
+++ b/sql/pgque-api/maint.sql
@@ -22,7 +22,7 @@ begin
         if f.func_name = 'pgque.maint_rotate_tables_step2' then
             continue;
         elsif f.func_name = 'vacuum' then
-            sql := 'vacuum ' || f.func_arg;
+            sql := 'vacuum ' || quote_ident(f.func_arg);
             execute sql;
             total := total + 1;
         elsif f.func_arg is not null then
--- a/sql/pgque.sql
+++ b/sql/pgque.sql
@@ -37,11 +37,11 @@ create schema if not exists pgque;
 -- 
 -- Standard triggers store events in the pgque.event_* data tables
 -- There is one top event table pgque.event_<queue_id> for each queue
--- inherited from pgque.event_template with three tables for actual data
+-- inherited from pgque.event_template wuith three tables for actual data
 -- pgque.event_<queue_id>_0 to pgque.event_<queue_id>_2.
 --
--- The active table is rotated at interval, so that if all the consumers
--- have passed some point the oldest one can be emptied using TRUNCATE command
+-- The active table is rotated at interval, so that if all the consubers
+-- have passed some poin the oldes one can be emptied using TRUNCATE command
 -- for efficiency
 -- 
 -- 
@@ -957,7 +957,7 @@ begin
     -- That should keep all ticks for all batches that are completely
     -- in old table.  This keeps them for longer than needed, but:
     -- 1. we want the pgque.tick table to be big, to avoid Postgres
-    --    accidentally switching to seqscans on that.
+    --    accitentally switching to seqscans on that.
     -- 2. that way we guarantee to consumers that they an be moved
     --    back on the queue at least for one rotation_period.
     --    (may help in disaster recovery)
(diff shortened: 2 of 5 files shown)


NikolayS:
@hobostay thank you

It would be great to split this PR into pieces - I see two different classes things here:

1. Typos in the original PgQ code, as well as "dead code" (for pre-14) – we aim to keep the original code as unchanged as possible, incorporating it into a single .sql.  I think it would be great to move typo fixes upstream: https://github.com/pgq/pgq
2. Everything else – I'm going to process separately

hobostay:
@NikolayS Thank you for the feedback! I've updated the PR to remove all typo fixes and dead code cleanup (PG 8.3 `pg_autovacuum` code path, non-existent schema references in `maint_tables_to_vacuum()`). These changes have been reverted in commit 1830ca2.

This PR now only contains the substantive bug fixes:

- **Go consumer**: per-message nack for handler failures, log warning for unregistered event types
- **Go client**: add missing `Nack()` method
- **`dlq_replay_all()`**: wrap per-event replay in exception blocks
- **`maint()`**: use `quote_ident()` for VACUUM target
- **`upgrade_schema()`**: fix inverted role hierarchy, fix return value
- **`insert_event_raw()`**: add queue existence check with clear error message

I'll look into submitting the typo fixes upstream to pgq/pgq separately.

sql/pgque-api/maint.sql
```diff
@@ -22,7 +22,7 @@ begin
         if f.func_name = 'pgque.maint_rotate_tables_step2' then
             continue;
         elsif f.func_name = 'vacuum' then
-            sql := 'vacuum ' || f.func_arg;
+            sql := 'vacuum ' || quote_ident(f.func_arg);
```
NikolayS, line 25: This is wrong for schema-qualified names like `pgque.event_0`; it becomes:

```
vacuum "pgque.event_0"
```

while we need:
```
vacuum "pgque"."event_0"
```

(I wonder if tests could catch this - maybe no at current state, because `maint_tables_to_vacuum()` returns nothing when `autovacuum = on`)

sql/pgque.sql
```diff
@@ -4407,10 +4413,15 @@ begin
...
-            v_dl.ev_extra1, v_dl.ev_extra2, v_dl.ev_extra3, v_dl.ev_extra4);
-        delete from pgque.dead_letter where dl_id = v_dl.dl_id;
-        v_cnt := v_cnt + 1;
+        begin
+            perform pgque.insert_event(v_dl.queue_name, v_dl.ev_type, v_dl.ev_data,
+                v_dl.ev_extra1, v_dl.ev_extra2, v_dl.ev_extra3, v_dl.ev_extra4);
+            delete from pgque.dead_letter where dl_id = v_dl.dl_id;
+            v_cnt := v_cnt + 1;
+        exception when others then
+            raise notice 'dlq_replay_all: failed to replay dl_id=%, error: %',
```
NikolayS, line 4422: I'm thinking, wo do we use NOTICE in cause of a failure? This can be overlooked if `log_min_messages` has a higher value value and hides NOTICEs

NikolayS:
@hobostay thank you!

a few things that come to mind before merging:

1. we need to fix/remove the quote_ident() VACUUM change - see a separate comment
2. I think we need tests for Go Nack() / handler failure behavior
3. Also need to clarify on dlq_replay_all() partial-success semantics (see a comment about NOTICE)

thanks!!

NikolayS:
Thanks @hobostay — merged this as-is to land the substantive fixes (Go per-message `nack`, `dlq_replay_all` exception isolation, `insert_event_raw` queue-existence check). I opened #79 with the follow-ups from review:

- `quote_ident` on `f.func_arg` was producing `vacuum "schema.table"` (one identifier with a literal dot) instead of `vacuum "schema"."table"` — fixed via `format('vacuum %I.%I', ...)`.
- `dlq_replay_all`'s per-event `raise notice` switched to `raise warning` (NOTICE is hidden under `log_min_messages = warning`); return type changed to `(replayed, failed, first_error)` so callers can detect partial success.
- Added Go tests for `Nack()` and per-message nack-on-handler-error.
- Two of your changes (`upgrade_schema` role grants, `insert_event_raw` queue check) were applied directly to `sql/pgque.sql`, which is auto-generated by `build/transform.sh`. Re-applied them as post-transform patches there so the next rebuild keeps them.

Appreciate the contribution — these were real issues worth fixing.

PR #77, opened 2026-04-19
small formatting tweak in reference.md

The-Alchemist:
the function was inside a code block so it wasn't formatted like the rest of the `pgque.*` functions in the rest of the file


NikolayS:
thanks @The-Alchemist

Issue #113, opened 2026-04-30
Umbrella: raw SQL interface audit findings from testing rounds 1–4


NikolayS:
## Summary

This is an umbrella / triage issue for the raw SQL interface audit performed against `main` at `9b3f89f`.

The audit produced issues #95–#112. This issue summarizes them, separates likely release blockers from docs/hardening/design questions, and calls out which findings appear inherited from PgQ assumptions vs locally introduced by PgQue wrappers/additions.

### Keep as real bugs / release blockers

- #96 — Security: PgQue roles are bypassed by default PUBLIC EXECUTE on functions
- #97 — Bug: concurrent `receive()` for same consumer can double-deliver same events
- #98 — Bug: `nack()` DLQ path trusts caller-supplied `pgque.message` and allows forged DLQ rows
- #101 — Security: `pgque_admin` can escalate via `queue_extra_maint` executed by SECURITY DEFINER `maint()`
- #103 — Bug: `receive()` can strand consumer on empty active batch with no `batch_id` to ack
- #107 — Bug: `batch_retry()` fails on `xid8` schema
- #109 — Bug: queue names >57 bytes fail via `pg_notify` channel length
- #110 — Bug: `pgque.maint()` VACUUM branch cannot work inside PL/pgSQL
- #111 — Client/docs: Python consumer/client contract can drop unhandled messages and misdocuments string payloads, if clients are in release scope

### Keep, but reframe / downgrade

- #95 — `receive(max_return)`
  - Real bug: `max_return <= 0` returns one row.
  - But “ack skips unreturned rows” may be intended whole-batch semantics; existing tests already codify it.
  - Suggested reframe: validate `max_return`, document that `ack()` finishes the whole underlying batch, and avoid presenting `max_return` as a safe paging/cursor API.

- #99 — Docs drift around manual scheduler, `maint_retry_events`, `force_tick`, `nack`, grants
  - Keep as docs umbrella.

- #100 — Config validation / external ticker / `force_tick()` hardening
  - Keep as hardening, probably not release blocker unless external ticker is in v0.2 scope.

- #108 — `get_batch_cursor(extra_where)` raw SQL predicate
  - Keep as API/docs hardening.
  - It is not necessarily a vulnerability if only trusted SQL writers can call it, but it must be documented as trusted-SQL-only and not safe for user input.

- #112 — Roles are global/coarse, not per-queue tenant isolation
  - Keep as explicit API contract docs.

### Merge or downgrade depending product decision

[...]

Issue #97, opened 2026-04-30
Bug: concurrent receive() for same consumer can double-deliver same events


NikolayS:
## Summary

Deep raw SQL testing found that two concurrent `pgque.receive(queue, same_consumer)` calls can both receive the same messages, with different `batch_id`s.

This appears to be a race in batch acquisition: the subscription row is read/updated without a strong enough lock or conditional update, so two sessions can both observe no active batch and allocate separate batches over the same tick range.

## Impact

For the same queue + consumer, concurrent workers can process the same event twice.

That violates the expected single-consumer cursor semantics for a registered consumer. Fan-out across different consumers is expected; duplicate delivery to the same consumer due to concurrent `receive()` is not.

## Evidence

Using a local two-session repro with an artificial sleep to widen the race, both sessions received the same `msg_id`s with different `batch_id`s:

```text
('T1', 1.03, [(1, 7, 'p1'), (2, 7, 'p2')])
('T2', 1.02, [(1, 6, 'p1'), (2, 6, 'p2')])
subscription [(7, 2, 1)]
```

Both sessions got `msg_id` 1 and 2. Final subscription state retained only one of the batch ids.

## Expected

For a given `(queue, consumer)`:

- at most one concurrent receiver should acquire the next batch
- a second concurrent receiver should either see the same active batch in a controlled/idempotent way, block, or return no new batch depending on intended semantics
- it should not allocate a distinct batch id for the same event range

## Likely fix direction

In the lower-level batch acquisition path (`next_batch_custom` / related subscription update):

- lock the subscription row (`FOR UPDATE`) before checking/updating `sub_batch`, or
- use a conditional update like `WHERE sub_batch IS NULL` and verify exactly one row updated, or
- use advisory lock keyed by subscription id

The invariant is: only one session can transition a subscription from no active batch to active batch.

## Test to add

Add a concurrency regression test with two sessions calling `receive()` for the same consumer at the same time. It should assert that the same `msg_id` is not delivered under two different batch ids.

## Environment

Tested on `main` at `9b3f89f`.

NikolayS:
## Investigation + fix summary

### (a) Intended model: single-worker-per-consumer

**Evidence (file + line + quote):**

- `sql/pgque.sql` line 178–179: `constraint subscription_pkey primary key (sub_queue, sub_consumer)` + `constraint subscription_batch_idx unique (sub_batch)` — one cursor row per (queue, consumer) pair; at most one active batch at any time.
[...]

PR #116, opened 2026-04-30
fix(pgque.nack): canonical + idempotent DLQ terminal handling

### LINKED ISSUE #98: Bug: nack() DLQ path trusts caller-supplied pgque.message and allows forged DLQ rows
## Summary

`pgque.nack(batch_id, msg pgque.message, ...)` trusts the caller-supplied `pgque.message` fields when routing to DLQ.

The retry path re-queries the canonical event, but the DLQ path calls `event_dead()` with fields from the supplied composite. A caller with an active batch can forge a `pgque.message` and insert arbitrary fake event data into `pgque.dead_letter`.

## Impact

Any writer that can call `nack()` and has an active batch can corrupt/forge DLQ records:

- fake `ev_id`
- fake `ev_type`
- fake payload
- fake retry count
- misleading reason

This can break DLQ observability, replay correctness, auditing, and operator trust.

## Repro

```sql
\i sql/pgque.sql

select pgque.create_queue('r_forge');
select pgque.set_queue_config('r_forge', 'max_retries', '0');
select pgque.register_consumer('r_forge', 'c');
select pgque.send('r_forge', 'realtype', 'realpayload');
select pgque.force_tick('r_forge');
select pgque.ticker();

-- Create active batch.
select count(*) from pgque.receive('r_forge', 'c', 1);

select sub_batch as active_batch
from pgque.subscription s
where q.queue_name = 'r_forge'
\gset

-- Forge message content. max_retries=0 sends it to DLQ.
select pgque.nack(
  :active_batch,
  row(
    999999,
    :active_batch,
    'forgedtype',
    'forgedpayload',
    999,
    now(),
    null, null, null, null
  )::pgque.message,
  interval '0 seconds',
  'forged'
);

select ev_id, ev_type, ev_data, dl_reason
from pgque.dead_letter dl
join pgque.queue q on q.queue_id = dl.dl_queue_id
where q.queue_name = 'r_forge';
```

Observed:

```text
ev_id  | ev_type    | ev_data       | dl_reason
-------+------------+---------------+----------
999999 | forgedtype | forgedpayload | forged
```

## Expected

`nack()` should only DLQ a real message from the specified active batch.

The function should not trust caller-supplied event fields for persistence.

## Suggested fix

Change `nack()` to re-query canonical event data from `pgque.get_batch_events(i_batch_id)` using `msg_id`, then call `event_dead()` with canonical fields.

Even better API shape:

```sql
pgque.nack(batch_id bigint, msg_id bigint, retry_after interval, reason text)
```

The composite `pgque.message` is convenient for clients, but it is not a trustworthy source of truth.

## Tests to add

- Forged `pgque.message` with non-existent `msg_id` must error.
- Forged fields for an existing `msg_id` must not affect DLQ contents.
- `nack()` with a message from another batch must error.

## Environment

### LINKED ISSUE #104: Bug: repeated nack() of same DLQ-bound message creates duplicate replayable DLQ rows
## Summary

Repeated `pgque.nack()` on the same active message can create duplicate DLQ rows when the message is terminal / DLQ-bound.

The retry path is deduplicated by `retry_queue` primary key. The DLQ path has no equivalent idempotency, so repeated terminal handling creates multiple replayable dead-letter entries for the same original message.

This is related to #98 (`nack()` trusts caller-supplied `pgque.message`) but distinct: even a non-forged repeated `nack()` duplicates DLQ state.

## Repro sketch

```sql
\i sql/pgque.sql

select pgque.create_queue('dup_dlq');
select pgque.set_queue_config('dup_dlq', 'max_retries', '0');
select pgque.subscribe('dup_dlq', 'c');

select pgque.send('dup_dlq', 'payload');
select pgque.force_tick('dup_dlq');
select pgque.ticker();

-- get active message/batch
select * from pgque.receive('dup_dlq', 'c', 100);

-- Call nack twice on the same active message, e.g. via repeated receive before ack.
with m as (
  select r from pgque.receive('dup_dlq', 'c', 100) as r limit 1
)
select pgque.nack((r).batch_id, r, interval '0 seconds', 'dead1') from m;

with m as (
  select r from pgque.receive('dup_dlq', 'c', 100) as r limit 1
)
select pgque.nack((r).batch_id, r, interval '0 seconds', 'dead2') from m;

select count(*), array_agg(dl_reason)
from pgque.dlq_inspect('dup_dlq', 10);
```

Observed:

```text
count=2, reasons={dead1,dead2}
```

`dlq_replay_all('dup_dlq')` then replays both rows, duplicating the payload.

## Expected

Terminal handling of a given `(queue, consumer, original ev_id)` should be idempotent while the batch is active, or repeated `nack()` should be rejected after the first terminal action.

## Suggested fix

Options:

- Add a unique constraint/index on DLQ identity, e.g. `(dl_queue_id, dl_consumer_id, ev_id)` where appropriate.
- Make `event_dead()`/`nack()` idempotent like retry insertion.
- Track message terminal state for active batches and reject repeated terminal `nack()`.

This should be paired with #98's fix to canonicalize/re-query event data instead of trusting the caller-supplied composite.

## Regression test

- max_retries=0
- receive one message
- call `nack()` twice before `ack()`
- assert exactly one DLQ row and one replayed event

## Environment

NikolayS:
## Summary

Fixes two related bugs in `pgque.nack()` / DLQ flow, tracked in the umbrella #113.

- Closes #98
- Closes #104

### Fix #98 — nack() DLQ path trusts caller-supplied pgque.message (forge)

**Root cause:** `nack()` passed caller-supplied composite fields (`i_msg.type`, `i_msg.payload`, etc.) directly to `event_dead()` without verifying the `msg_id` actually belonged to the active batch. Any writer with an active batch could forge a `pgque.message` with an arbitrary `ev_id`, payload, type, or retry count, inserting fake DLQ rows.

**Fix:** `nack()` now calls `get_batch_events(i_batch_id)` and looks up the row by `ev_id = i_msg.msg_id`. If `msg_id` is not found in the batch, it raises `msg_id % not found in batch %`. Only canonical fields from the batch data tables reach `event_dead()`.

**Red test evidence:**
```
psql:tests/test_nack_dlq_canonical.sql:84: ERROR:
  FAIL #98: nack() should have rejected forged msg_id 999999
```

**Green (after fix):**
```
NOTICE:  PASS #98: nack() rejected forged msg_id and DLQ is clean
```

### Fix #104 — repeated nack() creates duplicate replayable DLQ rows

**Root cause:** `event_dead()` used a plain `INSERT` with no duplicate guard. Calling `nack()` twice before `ack()` on the same terminal message inserted two `dead_letter` rows, causing `dlq_replay_all()` to replay the payload twice.

**Fix:** Added unique index `dl_queue_consumer_ev_idx` on `(dl_queue_id, dl_consumer_id, ev_id)` and changed the `INSERT` in `event_dead()` to `ON CONFLICT (dl_queue_id, dl_consumer_id, ev_id) DO NOTHING`. Repeated terminal `nack()` calls are now idempotent.

**Red test evidence:**
```
NOTICE:  DLQ row count: 2
ERROR:  FAIL #104: expected 1 DLQ row, got 2
```

**Green (after fix):**
```
NOTICE:  PASS #104: repeated nack() is idempotent (1 DLQ row)
```

## Changes

| File | Change |
|------|--------|
| `sql/pgque-api/receive.sql` | `nack()` re-queries canonical event via `get_batch_events`; raises on unknown `msg_id` |
| `sql/pgque-additions/dlq.sql` | New unique index `dl_queue_consumer_ev_idx`; `event_dead()` insert uses `ON CONFLICT DO NOTHING` |
| `sql/pgque.sql` | Regenerated (both changes above) |
| `tests/test_nack_dlq_canonical.sql` | New red/green regression tests for both bugs |
| `tests/test_api_dlq.sql` | Test 2 updated: no longer forges `retry_count` (was the #98 hole); uses `max_retries=0` instead |
| `tests/run_all.sql` | Added `test_nack_dlq_canonical` to suite |

## Test plan

- [x] Red tests confirmed failing on `main` before fix
- [x] Green tests pass after fix
- [x] Full suite (`tests/run_all.sql`) passes — all tests green

### DIFF
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -85,7 +85,14 @@ Grant: `pgque_writer`. Source: `sql/pgque-api/receive.sql`.
 
 #### `pgque.nack(batch_id bigint, msg pgque.message, retry_after interval default '60 seconds', reason text default null) → integer`
 
-Negative-acknowledges one message. If `msg.retry_count` is below the queue's `max_retries`, re-queues after `retry_after`; otherwise routes the event to `pgque.dead_letter` via `pgque.event_dead`. Returns `1`.
+Negative-acknowledges one message. Only `msg.msg_id` (and the `batch_id` argument) are honored from the composite — `type`, `payload`, `retry_count`, `created_at`, and the `extra*` fields are **ignored**. `nack()` re-queries the canonical event from the active batch and uses those server-side values for all decisions and writes.
+
+- If the canonical `ev_retry` is below the queue's `max_retries`, re-queues after `retry_after` (via `pgque.event_retry`).
+- If `ev_retry >= max_retries`, routes the canonical event to `pgque.dead_letter` (via `pgque.event_dead`). This is idempotent: repeated calls for the same terminal message produce exactly one DLQ row (the second call does nothing).
+- If `msg.msg_id` is not present in the active batch — including a `NULL` msg_id or a msg_id from a different batch — raises `msg_id % not found in batch %`.
+
+**v0.2 hardening note:** Prior versions trusted the caller-supplied `retry_count` to decide DLQ routing, which was a forge vector. Configure `max_retries` via `set_queue_config` to control DLQ promotion; do not forge `retry_count` in the composite.
+
 Grant: `pgque_writer`. Source: `sql/pgque-api/receive.sql`.
 
 ```sql
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -91,8 +91,6 @@ Negative-acknowledges one message. Only `msg.msg_id` (and the `batch_id` argumen
 - If `ev_retry >= max_retries`, routes the canonical event to `pgque.dead_letter` (via `pgque.event_dead`). This is idempotent: repeated calls for the same terminal message produce exactly one DLQ row (the second call does nothing).
 - If `msg.msg_id` is not present in the active batch — including a `NULL` msg_id or a msg_id from a different batch — raises `msg_id % not found in batch %`.
 
-**v0.2 hardening note:** Prior versions trusted the caller-supplied `retry_count` to decide DLQ routing, which was a forge vector. Configure `max_retries` via `set_queue_config` to control DLQ promotion; do not forge `retry_count` in the composite.
-
 Grant: `pgque_writer`. Source: `sql/pgque-api/receive.sql`.
 
 ```sql
--- a/sql/pgque-additions/dlq.sql
+++ b/sql/pgque-additions/dlq.sql
@@ -41,7 +41,14 @@ create table if not exists pgque.dead_letter (
 create index if not exists dl_queue_time_idx
     on pgque.dead_letter (dl_queue_id, dl_time);
 
+-- Unique index: one DLQ row per (queue, consumer, original ev_id).
+-- Required for idempotent insert in event_dead() (#104).
+create unique index if not exists dl_queue_consumer_ev_idx
+    on pgque.dead_letter (dl_queue_id, dl_consumer_id, ev_id);
+
 -- pgque.event_dead() -- move event to DLQ (called by nack() when max retries exceeded)
+-- The insert uses ON CONFLICT DO NOTHING so that repeated nack() calls for
+-- the same terminal message are idempotent (fix for #104).
 create or replace function pgque.event_dead(
     i_batch_id bigint,
     i_event_id bigint,
@@ -66,7 +73,8 @@ begin
         raise exception 'batch not found: %', i_batch_id;
     end if;
 
-    -- Insert into dead letter table (no re-query of batch events)
+    -- Idempotent insert: if the same (queue, consumer, ev_id) tuple already
+    -- exists (repeated nack() before ack()), silently skip the duplicate.
     insert into pgque.dead_letter (
         dl_queue_id, dl_consumer_id, dl_reason,
         ev_id, ev_time, ev_txid, ev_retry, ev_type, ev_data,
@@ -74,7 +82,8 @@ begin
     values (
         v_sub.sub_queue, v_sub.sub_consumer, i_reason,
         i_event_id, i_ev_time, i_ev_txid, i_ev_retry, i_ev_type, i_ev_data,
-        i_ev_extra1, i_ev_extra2, i_ev_extra3, i_ev_extra4);
+        i_ev_extra1, i_ev_extra2, i_ev_extra3, i_ev_extra4)
+    on conflict (dl_queue_id, dl_consumer_id, ev_id) do nothing;
 
     return 1;
 end;
--- a/sql/pgque-api/receive.sql
+++ b/sql/pgque-api/receive.sql
@@ -69,6 +69,15 @@ end;
 $$ language plpgsql security definer set search_path = pgque, pg_catalog;
 
 -- pgque.nack() -- retry or route to DLQ based on retry_count vs max_retries
+--
+-- Fix #98: re-query the canonical event row from the active batch using
+-- msg_id, instead of trusting caller-supplied pgque.message fields.
+-- A caller with an active batch could otherwise forge DLQ rows by
+-- supplying arbitrary ev_id / ev_type / ev_data in the composite.
+--
+-- Fix #104: DLQ insert is idempotent via ON CONFLICT in event_dead().
+-- Repeated nack() calls for the same terminal message produce exactly one
+-- dead_letter row.
 create or replace function pgque.nack(
     i_batch_id bigint,
     i_msg pgque.message,
@@ -77,8 +86,9 @@ create or replace function pgque.nack(
 returns integer as $$
 declare
     v_max_retries int4;
+    v_ev          record;
 begin
-    -- Single lookup: subscription -> queue config
+    -- Lookup: subscription -> queue config
     select coalesce(q.queue_max_retries, 5) into v_max_retries
     from pgque.subscription s
     join pgque.queue q on q.queue_id = s.sub_queue
@@ -88,16 +98,30 @@ begin
         raise exception 'batch not found: %', i_batch_id;
     end if;
 
-    if coalesce(i_msg.retry_count, 0) >= v_max_retries then
-        -- Move to dead letter queue (pass event fields, no re-query)
-        perform pgque.event_dead(i_batch_id, i_msg.msg_id,
+    -- Re-query the canonical event from the active batch (#98).
+    -- This ignores caller-supplied payload/type/extras and uses the real
+    -- values stored in the queue data tables.
+    select ev_id, ev_time, ev_txid, ev_retry, ev_type, ev_data,
+           ev_extra1, ev_extra2, ev_extra3, ev_extra4
+    into v_ev
+    from pgque.get_batch_events(i_batch_id)
+    where ev_id = i_msg.msg_id;
+
+    if not found then
+        raise exception 'msg_id % not found in batch %', i_msg.msg_id, i_batch_id;
+    end if;
+
+    if coalesce(v_ev.ev_retry, 0) >= v_max_retries then
+        -- Move to dead letter queue using canonical event data (#98).
+        -- event_dead() uses ON CONFLICT DO NOTHING for idempotency (#104).
+        perform pgque.event_dead(i_batch_id, v_ev.ev_id,
             coalesce(i_reason, 'max retries exceeded'),
-            i_msg.created_at, null::xid8, i_msg.retry_count,
-            i_msg.type, i_msg.payload,
-            i_msg.extra1, i_msg.extra2, i_msg.extra3, i_msg.extra4);
+            v_ev.ev_time, v_ev.ev_txid::text::xid8, v_ev.ev_retry,
+            v_ev.ev_type, v_ev.ev_data,
+            v_ev.ev_extra1, v_ev.ev_extra2, v_ev.ev_extra3, v_ev.ev_extra4);
     else
         -- Retry after delay
-        perform pgque.event_retry(i_batch_id, i_msg.msg_id,
+        perform pgque.event_retry(i_batch_id, v_ev.ev_id,
             extract(epoch from i_retry_after)::integer);
     end if;
     return 1;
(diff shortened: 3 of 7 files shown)


NikolayS:
## REV review — PR #116 (independent reviewer)

**Verdict: REQUEST CHANGES (non-blocking nits + one docs gap; core fix is sound)**

Confidence: high on the correctness of the fix; high on the docs gap; medium on test-coverage hardening suggestions.

---

### Security (HIGH priority — #98 is a forgery vector)

- **Canonical re-query is correct.** `nack()` now selects `ev_id, ev_time, ev_txid, ev_retry, ev_type, ev_data, ev_extra1..4` from `pgque.get_batch_events(i_batch_id)` filtered by `ev_id = i_msg.msg_id`, and only those canonical values are passed to `event_dead()` / `event_retry()`. Caller-supplied `type`, `payload`, `retry_count`, `created_at`, and extras are correctly ignored.
- **Cross-batch attack is blocked.** `pgque.subscription` declares `unique (sub_batch)`, so any given `batch_id` resolves to exactly one (queue, consumer) pair. `get_batch_events(batch_id)` only returns rows that belong to that batch, so passing `msg_id` from a different batch raises `msg_id % not found in batch %`. Verified.
- **SECURITY DEFINER + search_path preserved.** Both `pgque.nack` and `pgque.event_dead` keep `language plpgsql security definer set search_path = pgque, pg_catalog`.
- **Information leak in error message.** `raise exception 'msg_id % not found in batch %', i_msg.msg_id, i_batch_id` echoes the caller's input back. Not a real leak (caller already knows what they passed). Acceptable.
- **Index choice for #104 is correct.** Unique on `(dl_queue_id, dl_consumer_id, ev_id)` correctly handles fan-out: the same `ev_id` can still enter the DLQ for two different consumers (different `dl_consumer_id`), but cannot be duplicated for the same `(queue, consumer, ev_id)` triple.
- **`ON CONFLICT (col-list) DO NOTHING` does not mask other constraint violations.** Only the matching unique index on the listed columns is suppressed; FK violations on `dl_queue_id`/`dl_consumer_id` still raise. Good.

### Bug hunter

1. **NULL `msg_id` is implicitly handled but not explicitly tested.** When `i_msg.msg_id is null`, the `where ev_id = null` in the canonical re-query yields no rows, so `not found` raises `'msg_id <NULL> not found in batch %'`. Behavior is safe but the test suite never exercises it. Suggest a third test case: `nack` with a `pgque.message` whose `msg_id` is NULL — assert it raises and DLQ stays clean.

2. **Test 1 catches `when others`, which is too broad.** The forge-rejection test `exception when others then v_ok := true;` passes for any failure, including bugs unrelated to the forge check (e.g., a regression in `get_batch_events`). Tighten to match `sqlerrm like 'msg_id % not found in batch %'` or `sqlstate 'P0001'` so the test pinpoints the intended error.

3. **Forged-payload-with-valid-msg_id case is not covered.** The most interesting #98 attack is the one where the attacker passes a real `msg_id` from the active batch but forges `type`/`payload`/`extras`/`retry_count`. The PR's Test 1 only covers a non-existent `msg_id = 999999`. Suggest adding a Test 1b: receive a real message, mutate `v_msg.payload`, `v_msg.type`, `v_msg.retry_count := 999`, call `nack`, then assert the DLQ row contains the *real* `realpayload`/`realtype` and the canonical `ev_retry` (0), not the forged values. This is the assertion that proves canonical-only writes — currently nothing in the suite asserts canonical fields land in DLQ.

4. **`ev_txid::text::xid8` cast is correct but worth a one-line comment.** `get_batch_events.ev_txid` is declared `bigint` (legacy PgQ signature) while the underlying column and `dead_letter.ev_txid` are `xid8`; the text round-trip is the codebase convention (sql/pgque.sql lines 252-253 document it). Drop a 1-line comment so future readers don't think it's a typo.

5. **Concurrent double-nack race is logically safe but untested.** Two concurrent `nack` calls each pass `get_batch_events`, each call `event_dead`; the unique index + `ON CONFLICT DO NOTHING` makes the duplicate insert a no-op; both return 1. Since concurrent tests are hard in psql-only suites, this is acceptable to defer, but worth a comment in the test file noting that Test 2 covers serial-only and that the race is reasoned about above.

6. **`event_retry` (PgQ primitive) still uses a subtransaction (`exception when unique_violation`)** to suppress duplicate retry-queue inserts. Pre-existing PgQ behavior, not introduced by this PR, but flagged in CLAUDE.md design rule 4 ("no subtransactions in hot paths"). Out of scope for this PR; leave for #97/#102 follow-up.

### Test analyzer

- Red+green tests are present for both #98 and #104. Red evidence is documented in the PR body.
- `tests/test_api_dlq.sql` Test 2 was correctly updated: it used to forge `retry_count := 2` (the very vulnerability #98 fixes), now uses `max_retries=0` so the canonical `ev_retry=0` triggers DLQ on first nack. Good cleanup.
- Test isolation: each test creates and drops its own queue. No shared state. Good.
- Coverage gaps (see Bug hunter #1, #3 above): NULL msg_id, valid-msg_id-with-forged-payload.

### Guidelines / CLAUDE.md compliance

- Lowercase SQL keywords throughout. Good.
- `snake_case` identifiers. Good.
- Schema-qualification (`pgque.X`). Good.
- `SECURITY DEFINER ... SET search_path = pgque, pg_catalog`. Good.
- Conventional Commits: `fix(pgque.nack): canonical lookup + idempotent DLQ insert`. Good.
- **Both source files (`sql/pgque-api/receive.sql`, `sql/pgque-additions/dlq.sql`) AND the regenerated `sql/pgque.sql` are in the same commit.** Lesson from #114 applied. Good.

### Docs (this is the main blocking nit)

- **`docs/reference.md` was not updated.** The current entry for `pgque.nack` says: *"If `msg.retry_count` is below the queue's `max_retries`, re-queues..."* — but after this fix, `nack` IGNORES `msg.retry_count` and re-reads `ev_retry` from the canonical batch row. Users upgrading from v0.1 who relied on forged retry_count for tests/operations (as the prior `tests/test_api_dlq.sql` itself did) will be surprised. Please update the reference entry to:
  - State that `nack` re-queries the canonical event from the active batch and ignores caller-supplied `type`/`payload`/`retry_count`/`created_at`/`extras` in `msg`.
  - Document the new `msg_id % not found in batch %` error.
  - Add a one-line migration note ("v0.2 hardening: do not forge `retry_count` to drive DLQ; configure `max_retries` instead, as the updated `tests/test_api_dlq.sql` test 2 demonstrates").
- `docs/tutorial.md` line 254 ("Keep nacking. ... On the next nack it becomes retry_count=1...") still describes user-driven retry counting; the wording is fine because the canonical retry counter advances naturally, but a sentence noting that the counter is server-side would help.

### Anti-leak scan

`grep -iE "gitlab|sahmed|artifact[_-]?registry|@AR\b|wi[ -]?#?7[67]|round[ -]?[4-9]\b|R[4-9]\b"` over the diff, PR body, and commit message: clean. No external/private refs.

### CI

5/5 PG version test jobs (14, 15, 16, 17, 18) pass. `client-smoke` and `verify` pass. `claude-review` is the bot we're invoking; not a self-block.

---

### Summary

**Counts:** 0 must-fix-before-merge security/correctness issues. 1 docs gap (reference.md still describes pre-fix behavior). 3 test-coverage hardening suggestions (NULL msg_id, valid-msg_id-with-forged-payload assertion, narrower exception match). 0 anti-leak hits.

The fix is conceptually right and the implementation is clean. The docs update is the only thing I'd want before merge; the test-coverage suggestions are nice-to-have but would meaningfully harden the regression surface for a security-sensitive fix.

I am NOT approving or denying — flagging for the maintainer to address the docs gap and decide on test hardening.

NikolayS:
## REV Review — PR #116 (round 2)

**CI:** 5/5 PG matrix (14, 15, 16, 17, 18) green; `client-smoke` pass; `verify` pass; `claude-review` is this bot.
**R1 docs gap status:** addressed
**R1 hardening tests status:** Test 1 narrow / 1b forged / 1c NULL — all three present
**Migration-note scrub (`c808143`) impact:** preserved gap fix; only the historical "v0.2 hardening note" paragraph was removed (correct call — no v0.x users to migrate). The three behavioral statements survive.

### Verification table

| R1 finding | Status in r2 |
|---|---|
| `docs/reference.md` `nack()` entry described pre-fix behavior (BLOCKER) | Fixed in `38713e3`; survives scrub `c808143`. New entry explicitly says caller-supplied `type`/`payload`/`retry_count`/`created_at`/`extra*` are ignored, terminal DLQ insert is idempotent, and unknown/NULL `msg_id` raises `msg_id % not found in batch %`. |
| Test 1 `when others` is too broad | Fixed in `fc1db18`: now `exception when raise_exception` plus `assert sqlerrm like 'msg_id % not found in batch %'`. |
| Forged-payload-with-valid-msg_id case missing | Added as Test 1b in `fc1db18`: builds a `pgque.message` with the real `msg_id` but forged `type='forgetype'`, `payload='forgepayload'`, `retry_count=999`, `extras='fe1..fe4'`; asserts the DLQ row has `ev_type='realtype'`, `ev_data='realpayload'`, `ev_retry=0`, `ev_extra1 is null`. This is the assertion that proves canonical-only writes — exactly what was missing in r1. |
| NULL msg_id untested | Added as Test 1c in `fc1db18`: composite with `msg_id=null`; asserts `nack` raises `msg_id % not found in batch %` and DLQ stays empty. |
| `ev_txid::text::xid8` cast deserves a one-line comment | Added in `a2359d8`: comment in both `sql/pgque-api/receive.sql` and the regenerated `sql/pgque.sql` ("ev_txid is bigint in get_batch_events (legacy PgQ signature); text round-trip is the codebase convention to widen to xid8 without loss."). |

### Scrub impact (c808143)

Diff is exactly two deleted lines (the migration paragraph + its blank line). The three behavioral bullets remain at lines 88, 91, 92 of the post-scrub file. No regression.

### Anti-leak

`grep -iE "gitlab|sahmed|artifact[_-]?registry|@AR\b|wi[ -]?#?7[67]|round[ -]?[4-9]\b|R[4-9]\b"` over the full PR diff: clean.

### Counts

0 blocking, 0 non-blocking, 0 potential (all r1 items resolved). 0 anti-leak hits.

---

NikolayS:
Comment-scrub sweep applied per CLAUDE.md style (commit cae72ec): removed Fix #98, Fix #104, and idempotent-insert narrative from `receive.sql`, `dlq.sql`, and `pgque.sql`.

Issue #90, opened 2026-04-30
W3: PR #84 — split client-tests into per-client jobs (Go/Python/TypeScript)


NikolayS:
Parent: #85
Target PR: #84 (branch \`ci/live-pg-driver-tests\`)

## Scope

Currently \`client-tests\` runs all three driver suites in one job. The user wants per-client visibility: \"Go client tests\", \"Python client tests\", \"TypeScript client tests\" as separate jobs in the PR checks UI.

## Implementation

Edit \`.github/workflows/ci.yml\`. Replace the single \`client-tests\` job with three jobs:
- \`go-client-tests\` (name: \`Go client tests\`) — runs only the Go suite against the live PG service container with \`PGQUE_TEST_DSN\`.
- \`python-client-tests\` (name: \`Python client tests\`) — runs only the Python suite.
- \`typescript-client-tests\` (name: \`TypeScript client tests\`) — runs only the TS suite (\`npx vitest --run\`).

All three jobs:
- depend on \`verify\` (matching the current \`client-tests\` setup).
- share the same PG service-container config.
- can fail independently (parallel matrix-style or three sibling jobs — use whichever is cleaner; matrix preferred if it doesn't lose readability).

## Acceptance criteria

- After this change, opening PR #84 shows three separate check items in the GitHub UI: \"Go client tests\", \"Python client tests\", \"TypeScript client tests\".
- The Go job will FAIL on PR #84's HEAD until the Go bugfix PR (issue #PLACE_W1) lands, demonstrating the harness still catches the bug.
- The Python and TypeScript jobs should pass (assuming their PRs don't have integration bugs).

## TDD

Workflow change — no unit test, but verify by:
1. Push commit to \`ci/live-pg-driver-tests\`.
2. Watch GitHub Actions → confirm three separate jobs visible.
3. Take screenshot or paste \`gh pr checks 84\` output to PR + this issue as evidence.

## Reporting

- Comment on this issue when commit pushed.
- Comment with \`gh pr checks 84\` output once Actions runs.
- Final comment: PR ready for re-REV.

## Style + anti-leak

CLAUDE.md shell rules (\`#!/usr/bin/env bash\`, \`set -Eeuo pipefail\` if any new shell). No GitLab/AR/sahmed mentions.

## Don't merge

PR-only.

NikolayS:
Commit pushed: **29de606**

Replaced the single `client-tests` job with three sibling jobs in `.github/workflows/ci.yml`:

[...]

PR #117, opened 2026-04-30
fix(pgque.receive): finish empty batch instead of stranding consumer

### LINKED ISSUE #103: Bug: receive() can strand consumer on empty active batch with no batch_id to ack
## Summary

`pgque.receive()` can permanently strand a modern-wrapper consumer if it acquires an empty batch.

When a tick boundary contains zero visible events, `receive()` calls `next_batch()`, creates an active batch, returns zero rows, and exposes no `batch_id`. The caller cannot call `ack()` because there is no returned message containing the batch id.

Future `receive()` calls keep returning zero rows from the same active empty batch, even after new events arrive.

## Minimal repro

```sql
\i sql/pgque.sql

select pgque.create_queue('empty_batch');
select pgque.subscribe('empty_batch', 'c');

-- Create an empty tick interval.
select pgque.force_tick('empty_batch');
select pgque.ticker();

-- receive() acquires an active batch but returns no rows.
select count(*) from pgque.receive('empty_batch', 'c', 100);
-- 0

select sub_batch, sub_last_tick, sub_next_tick
from pgque.subscription s
join pgque.consumer c on c.co_id = s.sub_consumer
where q.queue_name = 'empty_batch'
  and c.co_name = 'c';
-- sub_batch=1, sub_last_tick=1, sub_next_tick=2

-- New event arrives later.
select pgque.send('empty_batch', 'later');
select pgque.force_tick('empty_batch');
select pgque.ticker();

-- Still stuck on the empty active batch.
select count(*) from pgque.receive('empty_batch', 'c', 100);
-- 0
```

At this point the modern wrapper user has no `batch_id` to acknowledge. A low-level operator could inspect `pgque.subscription.sub_batch` and call `ack()` manually, but wrapper users are stuck.

## Impact

A consumer using only the documented modern raw SQL API can stop receiving forever after an empty batch.

This is especially plausible because ticks can happen without visible events depending on force ticks, idle ticks, rollback windows, or snapshot boundaries.

## Suggested fix

`receive()` should handle empty batches internally, e.g.:

- if `get_batch_events(batch_id)` returns zero rows, call `finish_batch(batch_id)` automatically and continue to the next available batch, or
- expose batch metadata even for empty batches so callers can `ack()`, or
- make `next_batch()` skip empty batches for the wrapper path

The wrapper API should not expose an unacknowledgeable active batch state.

## Regression test

Empty tick → `receive()` returns zero → later send/tick → next `receive()` should return the later event without manual low-level `ack(sub_batch)`.

## Environment

Existing `tests/run_all.sql` passes.

NikolayS:
## Summary

Closes #103

### Root cause confirmed

`pgque.receive()` calls `next_batch()`, which opens a batch (sets `sub_batch`) for
every available tick window — even tick windows that contain zero visible events.
The old code then iterated `get_batch_events()`, returned `[]` with no rows, and
returned without calling `finish_batch()`.

Result: `sub_batch` stayed set. Every subsequent call to `receive()` hit the same
empty batch (because `next_batch()` sees `sub_batch IS NOT NULL` and returns the
existing batch_id immediately). The consumer was permanently stranded: zero rows
forever, even after new events arrived and a fresh tick fired. The caller had no
`batch_id` to `ack()` because no rows were returned.

This is the root cause of the empty-receive pattern in the Python (19 failed) and
TypeScript (6 failed) test suites reported in issue #90 / PR #84.

### Fix

In `pgque.receive()`, after iterating `get_batch_events()`, if `cnt == 0` (no
rows yielded), immediately call `finish_batch(v_batch_id)`. This advances
`sub_last_tick` and clears `sub_batch`, so the next `receive()` call can pick up
the following tick window normally.

The fix composes existing PgQ primitives only — no new state or logic —
satisfying design rule #3 ("modern API functions must reduce cleanly to PgQ
primitives").

Files changed:
- `sql/pgque-api/receive.sql` — source: added empty-batch auto-finish with comment
- `sql/pgque.sql` — assembled output: same fix reflected
- `tests/test_receive_empty_batch.sql` — new regression test (red → green)
- `tests/run_all.sql` — new test added to suite

### Red test output (before fix)

```
ERROR:  BUG #103: receive() left sub_batch set after returning 0 rows —
        consumer is stranded on empty batch id 10
CONTEXT:  PL/pgSQL function inline_code_block line 27 at ASSERT
```

### Green test output (after fix)

```
NOTICE:  PASS: receive() on empty tick window returns 0 rows and does not strand consumer
NOTICE:  PASS: second receive() sees new event after empty-batch window
NOTICE:  PASS: receive() empty-batch trap regression test (#103)
=== ALL TESTS PASSED ===
```

Full `tests/run_all.sql` passes locally (PG 16, macOS).

### Python / TypeScript test suites (PR #84 / issue #90)

The empty-receive pattern in both driver suites is caused by this same bug:
`receive()` opens an empty batch on the first call (initial tick before events
are visible), returns `[]`, and every subsequent call returns the same empty
batch. After this fix, the empty batch is finished internally and the next call
correctly sees newly arrived events.

The fix unblocks the failing tests listed in PR #84's CI run
(https://github.com/NikolayS/pgque/actions/runs/25158408976):

- Python: all `test_receive.py`, `test_nack.py`, `test_consumer.py`,
  `test_send.py::test_jsonb_payload_round_trip` failures are downstream of
  empty-receive
- TypeScript: all `client.test.ts` and `consumer.test.ts` receive-after-send
  failures are downstream of empty-receive; `nack.test.ts` failures are
  downstream (no batch_id to nack when receive returns [])

## Test plan

- [x] Red test `tests/test_receive_empty_batch.sql` fails on `main` with exact
  bug message
- [x] Green: same test passes after fix
- [x] Full `tests/run_all.sql` passes with fix
- [ ] CI: Go/Python/TypeScript driver tests via PR #84 (expect pass after merge)

### DIFF
--- a/sql/pgque-api/receive.sql
+++ b/sql/pgque-api/receive.sql
@@ -24,6 +24,17 @@ exception when duplicate_object then null;
 end $$;
 
 -- pgque.receive() -- wraps next_batch + get_batch_events
+--
+-- Fix for issue #103 — empty-batch trap:
+-- next_batch() opens a batch even when the tick window contains zero visible
+-- events. Before this fix, receive() would return [] while leaving sub_batch
+-- set, permanently stranding the consumer: subsequent calls kept returning
+-- the same empty batch and could never see later events.
+--
+-- Fix: after iterating get_batch_events(), if no rows were yielded, call
+-- finish_batch() immediately to advance the consumer's cursor. The caller
+-- still receives [], but the consumer is no longer stranded; the next
+-- receive() call will pick up the following tick window normally.
 create or replace function pgque.receive(
     i_queue text, i_consumer text, i_max_return int default 100)
 returns setof pgque.message as $$
@@ -36,7 +47,7 @@ begin
         raise exception 'pgque.receive: max_return must be >= 1, got %', i_max_return;
     end if;
 
-    -- Get next batch (may return NULL if no events)
+    -- Get next batch (may return NULL if no tick window is ready)
     v_batch_id := pgque.next_batch(i_queue, i_consumer);
     if v_batch_id is null then
         return;
@@ -56,6 +67,14 @@ begin
         cnt := cnt + 1;
         exit when cnt >= i_max_return;
     end loop;
+
+    -- Issue #103: if the batch had zero visible events, finish it now so the
+    -- consumer is not stranded. Callers receive [] either way; the difference
+    -- is that sub_batch is cleared here rather than being left open forever.
+    if cnt = 0 then
+        perform pgque.finish_batch(v_batch_id);
+    end if;
+
     return;
 end;
 $$ language plpgsql security definer set search_path = pgque, pg_catalog;
--- a/sql/pgque-api/receive.sql
+++ b/sql/pgque-api/receive.sql
@@ -24,17 +24,6 @@ exception when duplicate_object then null;
 end $$;
 
 -- pgque.receive() -- wraps next_batch + get_batch_events
---
--- Fix for issue #103 — empty-batch trap:
--- next_batch() opens a batch even when the tick window contains zero visible
--- events. Before this fix, receive() would return [] while leaving sub_batch
--- set, permanently stranding the consumer: subsequent calls kept returning
--- the same empty batch and could never see later events.
---
--- Fix: after iterating get_batch_events(), if no rows were yielded, call
--- finish_batch() immediately to advance the consumer's cursor. The caller
--- still receives [], but the consumer is no longer stranded; the next
--- receive() call will pick up the following tick window normally.
 create or replace function pgque.receive(
     i_queue text, i_consumer text, i_max_return int default 100)
 returns setof pgque.message as $$
@@ -68,9 +57,7 @@ begin
         exit when cnt >= i_max_return;
     end loop;
 
-    -- Issue #103: if the batch had zero visible events, finish it now so the
-    -- consumer is not stranded. Callers receive [] either way; the difference
-    -- is that sub_batch is cleared here rather than being left open forever.
+    -- Empty batch: finish immediately to advance the consumer cursor.
     if cnt = 0 then
         perform pgque.finish_batch(v_batch_id);
     end if;
--- a/sql/pgque.sql
+++ b/sql/pgque.sql
@@ -4581,17 +4581,6 @@ exception when duplicate_object then null;
 end $$;
 
 -- pgque.receive() -- wraps next_batch + get_batch_events
---
--- Fix for issue #103 — empty-batch trap:
--- next_batch() opens a batch even when the tick window contains zero visible
--- events. Before this fix, receive() would return [] while leaving sub_batch
--- set, permanently stranding the consumer: subsequent calls kept returning
--- the same empty batch and could never see later events.
---
--- Fix: after iterating get_batch_events(), if no rows were yielded, call
--- finish_batch() immediately to advance the consumer's cursor. The caller
--- still receives [], but the consumer is no longer stranded; the next
--- receive() call will pick up the following tick window normally.
 create or replace function pgque.receive(
     i_queue text, i_consumer text, i_max_return int default 100)
 returns setof pgque.message as $$
@@ -4625,9 +4614,7 @@ begin
         exit when cnt >= i_max_return;
     end loop;
 
-    -- Issue #103: if the batch had zero visible events, finish it now so the
-    -- consumer is not stranded. Callers receive [] either way; the difference
-    -- is that sub_batch is cleared here rather than being left open forever.
+    -- Empty batch: finish immediately to advance the consumer cursor.
     if cnt = 0 then
         perform pgque.finish_batch(v_batch_id);
     end if;
--- a/sql/pgque.sql
+++ b/sql/pgque.sql
@@ -4581,6 +4581,17 @@ exception when duplicate_object then null;
 end $$;
 
 -- pgque.receive() -- wraps next_batch + get_batch_events
+--
+-- Fix for issue #103 — empty-batch trap:
+-- next_batch() opens a batch even when the tick window contains zero visible
+-- events. Before this fix, receive() would return [] while leaving sub_batch
+-- set, permanently stranding the consumer: subsequent calls kept returning
+-- the same empty batch and could never see later events.
+--
+-- Fix: after iterating get_batch_events(), if no rows were yielded, call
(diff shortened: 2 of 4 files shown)


NikolayS:
## REV review — PR #117

Independent reviewer. Not approving / denying / merging.

### Summary

The fix is a textbook minimum-surface composition fix:

```sql
if cnt = 0 then
    perform pgque.finish_batch(v_batch_id);
end if;
```

placed after the `get_batch_events` loop in `pgque.receive`. The reasoning in the PR body matches the actual `pgque.next_batch_custom` mechanics in `sql/pgque.sql` (lines ~2115–2163): when `sub_batch is not null` the function returns the existing `batch_id` immediately, so a never-finished empty batch traps the consumer indefinitely. Calling `finish_batch` clears `sub_batch`, sets `sub_last_tick = sub_next_tick`, and lets the next `receive()` advance.

CI is green on PG 14–18.

### Security
- `pgque.receive` retains `security definer set search_path = pgque, pg_catalog`. PASS.
- `pgque.finish_batch` is itself security definer with the same pinned search_path, so the inner `perform` is safe. PASS.
- No new attack surface; no new GRANTs. PASS.

### Bug hunter — edge cases (the headline)
- **Partial-batch case** (`max_return < cnt`): condition is strictly `cnt = 0`. The loop only `exit`s when `cnt >= i_max_return`, so any partial yield leaves `cnt > 0` and `finish_batch` is not called. The pre-existing whole-batch-ownership semantics documented in `docs/reference.md:79` are preserved. PASS.
- **`next_batch` returned NULL**: the function returns at line 30 (`if v_batch_id is null then return; end if;`) before the new code runs. No `finish_batch(NULL)` risk. PASS.
- **`retry_queue` path**: `pgque.receive` does not query `retry_queue` directly. Retry events surface via `pgque.maint_retry_events` -> `pgque.queue_*` and are picked up through the normal batch/tick flow, so they are subject to the same empty-batch trap and benefit from the same fix. No separate code path needs handling here. PASS.
- **Concurrency**: PgQ assumes one worker per `(queue, consumer)`. Two simultaneous `receive()` calls on the same subscription is already an unsupported pattern at the engine level (the `next_batch`/`finish_batch` UPDATEs on `pgque.subscription` race regardless). The fix does not make this worse: a duplicated `finish_batch` against an already-cleared `sub_batch` simply returns 0 with a warning (`finish_batch: batch % not found`) — no corruption. NEUTRAL / PASS.
- **Design rule #3** (modern API reduces to PgQ primitives): satisfied — only `pgque.next_batch`, `pgque.get_batch_events`, `pgque.finish_batch` are used. PASS.
- **Design rule #4** (no subtransactions in hot paths): no `begin ... exception` introduced. PASS.

### Tests
- Red/green TDD ordering visible in commit history (`ad5c1a0` red test, then `a3ceece` fix). PASS.
- `tests/test_receive_empty_batch.sql` reproduces the bug crisply via `force_tick + ticker` with no events queued, then asserts directly on `pgque.subscription.sub_batch is null` — i.e. it tests the actual stranding state, not just the `[]` return value. The "second `receive()` sees the new event after the empty window" assertion is the strand-release proof the review brief asks for. PASS.
- The `assert v_msg.type = 'hello'` post-second-receive sanity-checks correct ordering. PASS.
- `tests/run_all.sql` is updated in the same commit as the fix (no PR-#114-style drift between source and registered suite). PASS.

### Guidelines
- Lowercase SQL keywords, snake_case, schema-qualified `pgque.*` throughout the diff. PASS.
- `sql/pgque-api/receive.sql` (source) and `sql/pgque.sql` (assembled) updated in the same commit `a3ceece`. The textual changes match line-for-line. No drift. PASS.
- Conventional Commit `fix(pgque.receive): finish empty batch, don't strand consumer`, subject 51 chars (one over the 50-char guideline; cosmetic only — flag, do not block). MINOR.
- Apache-2.0/PgQ-derivation header preserved in `tests/test_receive_empty_batch.sql`. PASS.

### Docs
The caller-visible API contract in `docs/reference.md:70-79` is still accurate (`receive` still returns `setof pgque.message`, still empty when nothing is available). The internal "empty batches are auto-finished" behavior is an implementation detail that does not need to be surfaced to API users — they could not observe `sub_batch` from the public API anyway. **Optional improvement**: a one-line note that empty tick windows are safe (no need to manually finish) would help advanced users who read the function source. Not blocking.

### Python / TypeScript unblock hypothesis — confidence: HIGH

The failure pattern reported in https://github.com/NikolayS/pgque/issues/90 (and PR #84's CI run) is exactly: send → tick → receive returns `[]` → all subsequent receives also return `[]`. That is the issue #103 stranding signature, not a driver-level bug. Reasoning:

1. Test fixtures typically `create_queue` + `subscribe` + (optionally) tick before sending. The very first `receive()` opens a batch on whatever the initial tick window is, which is almost always empty given fixture ordering.
2. With the bug, every subsequent `receive()` returns the same empty batch forever — perfectly matching `expected [] to have a length of 1` across both Python and TypeScript suites.
3. nack failures are a direct downstream symptom (no `batch_id` to nack when `receive` returns `[]`).
4. `test_smoke.py::test_smoke_send_receive_ack` failing on `register_consumer_at` ("Event queue not created yet") is a different issue — fixture ordering, not the empty-batch trap. That specific failure will likely persist after this PR merges.

I expect the fix to clear ~all `assert 0 == 1` / "expected length 1" failures in both driver suites, and the nack downstream cascade. Recommend confirming by re-running PR #84's CI after #117 merges; the smoke-test queue-lifecycle failure will need its own fix.

### Anti-leak

Scanned PR diff, body, title, and commit messages for restricted markers. CLEAN.

### Verdict

| Perspective    | Result |
|----------------|--------|
| Security       | PASS   |
| Bug hunter     | PASS (with concurrency note) |
| Tests          | PASS   |
| Guidelines     | PASS (1 minor: subject 51 chars) |
| Docs           | PASS (optional reference.md note suggested) |

- Blockers: 0
- Major: 0
- Minor: 1 (commit subject 51 chars)
- Nits / optional: 1 (one-line `reference.md` clarification on empty windows)

Confidence in the fix: **high**. Confidence the fix unblocks the Python/TS receive-after-send failures: **high**. Confidence it also clears the `register_consumer_at`/"Event queue not created yet" smoke-test failure: **low** — that looks like a separate fixture-ordering issue.

Reviewer recommendation (advisory only): merge after the maintainer re-runs PR #84 CI to confirm the unblock, and treat the smoke-test queue-lifecycle issue as a separate follow-up.

— REV reviewer (not approving / merging)

PR #118, opened 2026-04-30
fix: revoke PUBLIC EXECUTE and harden SECURITY DEFINER queue_extra_maint

### LINKED ISSUE #96: Security: pgque roles are bypassed by default PUBLIC EXECUTE on functions
## Summary

Round-2 raw SQL security testing found that PgQue's role model is porous because most functions keep PostgreSQL's default `PUBLIC EXECUTE` privilege.

Although docs describe `pgque_reader`, `pgque_writer`, and `pgque_admin` as distinct access levels, a read-only / ungranted role can execute writer/admin-ish `SECURITY DEFINER` functions such as `create_queue`, `drop_queue`, and `send`.

## Impact

A role intended to be read-only can mutate PgQue state:

- create queues
- drop queues
- send events
- possibly manage consumers / config / lifecycle functions depending on schema usage and function grants

This weakens the intended `pgque_reader` / `pgque_writer` / `pgque_admin` separation.

## Repro

```sql
\i sql/pgque.sql

create role r_reader login;
grant pgque_reader to r_reader;

set role r_reader;

select
  has_schema_privilege(current_user, 'pgque', 'USAGE') as schema_usage,
  has_function_privilege(current_user, 'pgque.create_queue(text)', 'EXECUTE') as can_create,
  has_function_privilege(current_user, 'pgque.drop_queue(text)', 'EXECUTE') as can_drop,
  has_function_privilege(current_user, 'pgque.send(text,text,text)', 'EXECUTE') as can_send;

select pgque.create_queue('r_reader_created');
reset role;
```

Observed:

```text
schema_usage | can_create | can_drop | can_send
-------------+------------+----------+---------
t            | t          | t        | t

create_queue
------------
1
```

Also verified with an ungranted PUBLIC role in a local probe:

```text
PASS: ungranted PUBLIC role can create queue
PASS: ungranted PUBLIC role can drop queue it did not own
```

## Expected

- `pgque_reader` should not be able to execute writer/admin mutating APIs.
- Ungranted PUBLIC roles should not be able to create/drop queues or mutate PgQue state.

## Suggested fix

Use deny-by-default function privileges:

```sql
revoke execute on all functions in schema pgque from public;
```

Then explicitly grant the intended functions to each role:

- reader: introspection/read-only functions only
- writer: send/receive/ack/nack/subscribe APIs and necessary PgQ primitives
- admin: queue lifecycle/config/maintenance functions
- uninstall: keep superuser/schema-owner only

Need to account for functions created after blanket revokes too, especially colocated API/DLQ functions.

## Tests to add

Add a role matrix test using actual `SET ROLE`, not only `has_function_privilege`:

- reader cannot execute: `create_queue`, `drop_queue`, `send`, `receive`, `ack`, `nack`, `dlq_replay`, `set_queue_config`
- writer can execute intended producer/consumer APIs but cannot purge/drop/admin config if policy says so
- admin can execute admin functions
- PUBLIC cannot mutate PgQue

## Evidence

Existing `tests/run_all.sql` passed before this probe.

### LINKED ISSUE #101: Security: pgque_admin can escalate via queue_extra_maint executed by SECURITY DEFINER maint()
## Summary

Round-3 adversarial security review found a privilege escalation path: `pgque_admin` can cause attacker-controlled code to run as the `SECURITY DEFINER` owner of `pgque.maint()` via `queue_extra_maint`.

This is not just the broad PUBLIC EXECUTE problem from #96. It is a separate SECURITY DEFINER + dynamic execution issue.

## Why it happens

`pgque_admin` receives broad privileges:

- `ALL ON SCHEMA pgque`
- `ALL ON ALL TABLES IN SCHEMA pgque`
- `EXECUTE ON ALL FUNCTIONS IN SCHEMA pgque`

`pgque.maint_operations()` reads function names from `pgque.queue.queue_extra_maint`.

`pgque.maint()` is `SECURITY DEFINER` and dynamically executes those names:

```sql
execute 'select ' || f.func_name || '(' || quote_literal(f.func_arg) || ')' into r;
-- or
execute 'select ' || f.func_name || '()' into r;
```

A `pgque_admin` can therefore:

1. create a function in `pgque`, e.g. `pgque.evil_maint(text)`
2. update `pgque.queue.queue_extra_maint`
3. call `pgque.maint()`
4. have that function execute as the installer / SECURITY DEFINER owner

If PgQue was installed by a superuser or high-privileged owner, this becomes escalation to that owner.

## Evidence

Local repro showed `pgque_admin` had the needed privileges and the injected function ran as the owner:

```text
can_create_schema=t can_update_queue=t can_maint=t
pgque.maint() -> evil function inserted:
who=tars sess=tars is_super=t
```

## Impact

A role intended to be PgQue admin/operator can become the PgQue install owner. If the install owner is superuser, this is superuser escalation.

## Suggested fix

- Do not execute arbitrary text from `queue_extra_maint` under SECURITY DEFINER.
- Replace `queue_extra_maint text[]` with a controlled/validated registry, e.g. `regprocedure` entries owned/approved by install owner.
- Consider removing `CREATE` on schema from `pgque_admin` unless explicitly required.
- Consider not granting raw `UPDATE` on internal tables to `pgque_admin`; use validated admin functions instead.
- Quote/validate dynamic function references at minimum, but validation/ownership is the main issue.

## Regression test

As `pgque_admin`:

- create `pgque.evil_maint(text)`
- try to attach it through `queue_extra_maint`
- call `pgque.maint()`
- assert it does **not** run as install owner / is rejected

## Environment

NikolayS:
## Summary

Fixes two HIGH-severity security issues found in the round 2-3 raw SQL audit.

Closes #96
Closes #101

---

## Issue #96 — PUBLIC EXECUTE revoked from all pgque functions

**Root cause:** PostgreSQL grants `EXECUTE` to `PUBLIC` for every new function by default. PgQue functions were created without any revoke, so any role (even ungranted) could call `create_queue`, `send`, `receive`, `maint`, etc.

**Fix:** Deny-by-default posture:
- `sql/pgque-additions/roles.sql`: blanket `revoke execute on all functions in schema pgque from public;` (covers all core + additions functions created before roles.sql runs in assembly order)
- `sql/pgque-additions/dlq.sql`: per-function `revoke execute ... from public` for each DLQ function (created after roles.sql)
- `sql/pgque-api/maint.sql`, `receive.sql`, `send.sql`: per-function revokes colocated with function definitions (API layer runs after roles.sql)
- The blanket `revoke execute on function pgque.uninstall() from public` that previously existed is now redundant with the blanket revoke and folded in

**Red test output** (`tests/test_security_public_execute.sql`):
```
ERROR:  PUBLIC EXECUTE not revoked from 29 function(s): pgque.create_queue(text),
pgque.drop_queue(text), pgque.set_queue_config(text, text, text), pgque.insert_event(text, text, text),
pgque.send(text, text), pgque.receive(text, text, integer), pgque.ack(bigint),
pgque.maint(), pgque.ticker(), pgque.force_tick(text), ... [29 total]
```

**Green:** `PASS: security_public_execute - PUBLIC EXECUTE revoked from all mutating functions`

---

## Issue #101 — queue_extra_maint ownership check in SECURITY DEFINER maint()

**Root cause:** `pgque.maint()` is `SECURITY DEFINER` and dynamically executes function names stored in `pgque.queue.queue_extra_maint`. A `pgque_admin` with `CREATE` on the pgque schema and `UPDATE` on pgque.queue can inject an attacker-owned function name, then call `maint()` to run that function as the install owner.

**Option chosen: Option A — ownership check validation.**
Before executing any `queue_extra_maint` entry, `maint()` now:
1. Validates the entry resolves to a `regprocedure` (catches invalid/nonexistent names)
2. Compares the function's owner to the owner of `maint()` itself (the install owner)
3. If the owners differ, skips the function and emits a `WARNING` — never executes it

This preserves `queue_extra_maint` functionality for legitimate same-owner extension hooks while closing the escalation path. Rationale vs alternatives:
- Option B (drop SECURITY DEFINER from maint()): would break the pg_cron + unprivileged-caller pattern
- Option C (superuser-only maint()): too restrictive for ops teams using pgque_admin

**Red test output** (`tests/test_security_extra_maint.sql`):
```
ERROR:  SECURITY DEFINER escalation: evil_maint_probe executed 1 time(s) under maint()
— queue_extra_maint ownership check missing
```

**Green:**
```
WARNING:  pgque.maint: skipping queue_extra_maint entry pgque.evil_maint_probe
— owner (pgque_attacker) does not match maint() owner (nik), refusing to execute under SECURITY DEFINER
PASS: security_extra_maint - attacker-owned function in queue_extra_maint was not executed by maint()
```

---

## Test plan

- [ ] `psql -d pgque_test -f sql/pgque.sql` (fresh install)
- [ ] `psql -d pgque_test -f tests/test_security_public_execute.sql` — PASS
- [ ] `psql -d pgque_test -f tests/test_security_extra_maint.sql` — PASS
- [ ] `psql -d pgque_test -f tests/run_all.sql` — all existing tests pass (rotation timing issue is pre-existing on main, unrelated to this PR)
- [ ] Verify `SET search_path` preserved on all SECURITY DEFINER functions: `test_security_definer` passes


NikolayS:
## REV review — PR #118

**Verdict: REQUEST CHANGES (recommend)** — the fixes land the right approach for both #96 and #101, but there are two concrete defects that should be addressed before merge (one is a likely runtime regression on non-superuser installs; one is a docs/contract gap that will mislead users about the new posture). I am NOT approving or merging — flagging for the maintainer.

CI: green on PG 14/15/16/17/18, client-smoke, verify. claude-review pending.

---

### Security (HEADLINE)

**#96 — PUBLIC EXECUTE revoke**

Coverage check (audited via line numbers in `sql/pgque.sql` on the head ref):

- Blanket `revoke execute on all functions in schema pgque from public` lands at line 4273 (end of `roles.sql`).
- All 56 functions defined before line 4273 (PgQ core, additions through `lifecycle.sql`/`status`) are covered by the blanket.
- The 18 functions defined AFTER line 4273 (5 in `dlq.sql`, `maint`, `receive`/`ack`/`nack`, 8 send variants, `subscribe`/`unsubscribe`) all have colocated explicit `revoke execute … from public` statements. I cross-checked each name. No gaps in this PR.
- Inheritance chain `admin > writer > reader` is preserved in `roles.sql`. Reader/writer/admin grants still resolve.
- `uninstall()` is correctly revoked from both PUBLIC (via the blanket) and `pgque_admin` (explicit revoke).

**Concrete fragility (S2):** the safety of the "per-file colocation" pattern depends on every future SQL file that adds a function remembering to colocate its own `revoke execute … from public`. There is no defense-in-depth net. I'd recommend appending one tail-end statement to `transform.sh` (after the pgque-api section) of the form:

```sql
-- Defense-in-depth: re-revoke any function added in pgque-api/* that may
-- have shipped without an explicit revoke.
revoke execute on all functions in schema pgque from public;
-- Then re-grant the API-layer functions explicitly.
```

This costs one line in the assembled output and turns the per-file rule from a tribal-knowledge invariant into a guaranteed posture. Without it, a future PR adding an api file is a footgun. Comment in `roles.sql` says "Note: functions created AFTER this point … carry their own colocated revoke" — that's a maintainer convention, not a build-time check.

**#101 — `queue_extra_maint` ownership check**

The Option A approach (resolve to `regprocedure`, compare owner to `maint()` owner) is sound and the chosen `IS DISTINCT FROM` form correctly handles the "both same role" case.

**Critical (S1) — non-superuser install will break the check:** the maint-owner and extra-maint-owner lookups both query `pg_authid`:

```sql
select rolname into v_maint_owner
from pg_proc p
join pg_authid a on a.oid = p.proowner   -- pg_authid requires superuser
…
```

`pg_authid` is restricted (only `pg_read_all_data` / superuser can read it). When pgque is installed by a non-superuser owner (the realistic managed-DB case — RDS, Aurora, Cloud SQL, AlloyDB, Supabase, Neon, Crunchy Bridge — per `CLAUDE.md`'s "managed database environments" framing), `maint()` runs as that owner under SECURITY DEFINER and the `pg_authid` SELECT will raise `permission denied for table pg_authid`. The whole `maint()` call fails, breaking rotation/retry/vacuum until reverted.

**Fix:** replace `pg_authid` with `pg_roles` (the public-readable view). Both queries should be:

```sql
select rolname into v_maint_owner
from pg_proc p
join pg_roles r on r.oid = p.proowner
join pg_namespace n on n.oid = p.pronamespace
where n.nspname = 'pgque' and p.proname = 'maint' and …;
```

This is a 2-character change (`pg_authid a` → `pg_roles r`, plus join target). Please add a regression test that runs `\i pgque.sql` as a non-superuser install owner and verifies `maint()` succeeds with a benign queue_extra_maint entry. Without that, this defect can re-land.

**Edge case (S2) — same-superuser collapse is not documented.** If `pgque_admin` happens to also be the install owner (single-superuser setup, common in dev/test), the ownership check passes by definition and the escalation path remains. This is intentional but should be called out in `docs/reference.md` and/or release notes. Otherwise dev environments will look "secure" while production with separated roles is the only place the check actually fires.

**Edge case (S3) — null/empty `queue_extra_maint`.** Verified by reading `pgque.maint_operations()`: it only emits entries from the `queue_extra_maint` array branch when `queue_extra_maint is not null`. The `f.func_arg is not null` branch in `maint()` correctly gates the new validation. No-op behavior on null/empty is preserved.

**Edge case (S3) — regprocedure resolution determinism.** The cast is `format('select %L::regprocedure', f.func_name || '(text)')`. The `%L` quoting prevents injection; the `(text)` suffix matches the only signature `maint()` actually invokes (`select fname(quote_literal(arg))`). Resolution happens with `search_path = pgque, pg_catalog`, which matches the search_path used by the actual `execute` line, so the validated OID is consistent with what gets executed. Good. Worth a one-line comment in the source explaining this invariant for future readers.

**Edge case (S3) — WARNING log spam DoS.** A `pgque_admin` could attach attacker-owned hooks to many queues and call `maint()` repeatedly to fill logs. Acceptable trade-off vs raising (which would break legitimate parallel hooks). The WARNING text quotes `f.func_name` and the two role names — no privileged information leaked.

---

### Bug hunter

- Per-function revokes in dlq.sql / maint.sql / receive.sql / send.sql cross-check: all 18 post-roles functions accounted for above.
- (S2) **Breaking-change for existing users with non-install-owner queue_extra_maint hooks:** any external integration that has registered an admin-owned (rather than install-owner-owned) function in `queue_extra_maint` will silently start being skipped after this fix, with only a WARNING in the log. The PR does not include a release-note / migration callout. Recommend: add a NEWS entry or `docs/reference.md` migration paragraph under "Upgrading" so operators check for and re-own existing hooks. (Realistic prevalence: low — `SPECx.md` already says queue_extra_maint should be CHECK-constrained to NULL in v1, but no constraint exists in the SQL.)
- (S3) `f.func_name` is interpolated raw into the dynamic `execute 'select ' || f.func_name || …'` line. The regprocedure validation upstream rejects ill-formed names, so injection is closed via that funnel. Worth a comment.

---

### Test analyzer

- Red+green for #96: present (`test_security_public_execute.sql`). Coverage list of 29 functions is broad and representative; uses `has_function_privilege('pgque_none_role', …, 'EXECUTE')` which is a reliable signal. Good.
- Red+green for #101: red coverage is good (attacker-owned function, asserts canary table is empty). Benign proof (insert into sentinel table) — not destructive. Good.
- (S2) **Missing positive test for legitimate same-owner queue_extra_maint hook.** The current test only asserts that an attacker-owned function is blocked. A second test that creates an install-owner-owned function in `pgque.*`, attaches it to `queue_extra_maint`, calls `maint()`, and asserts the function ran (canary row inserted) would lock in the green path. Right now a future regression that breaks the legitimate path (e.g., the `pg_authid` issue above) wouldn't be caught.
- (S2) **No test for non-superuser install owner.** This is the realistic deployment shape on managed databases — see Security S1 above. Without this test, the `pg_authid` defect sails through CI.
- Tests are added to `tests/run_all.sql` ✓.

---

### Guidelines

- Lowercase SQL keywords throughout. ✓
- snake_case identifiers. ✓
- Schema-qualified references (`pgque.queue`, `pgque.maint`). ✓
- All `SECURITY DEFINER` functions retain `set search_path = pgque, pg_catalog`. The modified `maint()` keeps it on line 92. ✓ (Key design rule #1 honored.)
- Source files (`pgque-additions/*`, `pgque-api/*`) and the regenerated single-file `sql/pgque.sql` are all updated in the same commit. ✓
- Conventional Commits subject (`fix: …`). ✓ Subject is 47 chars (within the <50 guideline).
- Two-commit branch (test commit `e6927e3` first, fix commit `ed3e6a1` second) — clean red→green TDD pattern matching the project's TDD rule. ✓

---

### Docs

- (S2) **`docs/reference.md` is stale.** The current text still says (lines 133, 433):
  - "Most functions in this section are left on PUBLIC by default — tighten with `revoke execute … from public` if your policy demands it"
  - "any role that can connect can still call [`uninstall()`] and drop the schema. Tighten with `revoke execute on function pgque.uninstall() from public;` if stricter control is required"
  - "All other functions … (notably `start`, `stop`, `status`, `maint`, `ticker`, `force_tick`, `create_queue`, `drop_queue`, `set_queue_config`) — revoke and re-grant explicitly if your policy demands it."

  All three statements are now incorrect after this PR. Users reading the docs will think they need to add manual revokes; some will trip over already-revoked functions. `reference.md` needs to be flipped to a deny-by-default narrative, and the per-role grant table needs to reflect that `pgque_reader/_writer/_admin` are now strictly required to call any function.
- (S2) No migration / upgrade note for existing pgque installs that may hold `queue_extra_maint` entries owned by `pgque_admin` rather than the install owner. Add a short paragraph to `docs/reference.md` (or a new `docs/upgrading.md`) calling this out.

---

### Summary

| Severity | Count |
|---|---|
| S1 (blocking) | **1** — `pg_authid` lookup breaks `maint()` on non-superuser installs |
| S2 (should-fix before merge) | **5** — defense-in-depth tail revoke, missing green test, missing non-superuser test, stale `reference.md`, missing migration note |
| S3 (nit / follow-up) | **3** — same-superuser-collapse callout, regprocedure resolution comment, raw-name interpolation comment |

**Confidence:** high. The diff is small (387/-12) and audit was line-by-line.

**Breaking-change risk for #101: low–medium.** Low because:
- The `queue_extra_maint` column was preserved-but-discouraged per `SPECx.md`.
- Most pgque installs are unlikely to have non-install-owner functions registered.

Bumped to medium by the `pg_authid` defect: ANY pgque install on a non-superuser owner will see `maint()` start failing after this PR, regardless of whether they use `queue_extra_maint`. That is a higher-impact regression than the `queue_extra_maint`-specific concern. Fixing the `pg_authid` issue drops this back to low.

This is a code review only — I have not approved, denied, or merged the PR.

NikolayS:
## REV Review — PR #118 (round 2)

**CI:** all green (test 14/15/16/17/18, client-smoke, verify, claude-review).
**R1 BLOCKING resolution status:** addressed (with one caveat noted below).
**R1 should-fix status:**
- S2-1 (defense-in-depth tail revoke): addressed. `revoke execute on all functions in schema pgque from public;` is the very last statement in `sql/pgque.sql` (line 4921 of 4922), after every `create or replace function pgque.X` and after every per-function `grant execute ... to pgque_*`. Verified no further function declarations follow. Comment correctly notes role-specific grants are untouched.
- S2-2 (green test for legit same-owner hook): addressed (Test B in `tests/test_security_extra_maint.sql`).
- S2-3 (non-superuser install test): **partially addressed** — see Potential below.
- S2-4 (`docs/reference.md` for `maint()`): addressed for `maint()` specifically (grant flipped to `pgque_admin`, ownership-check + `pg_get_userbyid` notes added, migration note present). The broader `reference.md` "PUBLIC by default" staleness for the rest of the API (lines 110, 115, 120, 125, 138, 143, 148, 173, 178, 185, 367, 372, 392, 396, plus the trailing paragraph at line 439 — "All other functions ... are on PUBLIC `execute` by default") is **still stale**: the tail blanket revoke at line 4921 of `pgque.sql` now revokes PUBLIC EXECUTE from every function in the schema, so those `Grant: PUBLIC (default)` lines and the line-439 paragraph are no longer accurate. Not blocking, but should be cleaned up in a follow-up doc PR.
- S2-5 (migration note): addressed — present in the `maint()` entry of `docs/reference.md`.

---

### Blocking
None.

The R1 BLOCKING `pg_authid` finding is addressed at both call sites in `sql/pgque-api/maint.sql` (and the assembled `sql/pgque.sql`), via `pg_catalog.pg_get_userbyid(p.proowner)`. `pg_get_userbyid` is a built-in stable function in `pg_catalog`, callable by every role without `pg_authid` privilege. Name-based vs OID-based semantics: both the pre-fix code (`select rolname from pg_authid`) and the post-fix code (`pg_get_userbyid(...)`) compare role *names*, so the ownership check semantics are unchanged. `is distinct from` correctly handles the (unlikely) NULL case where the owner OID has been dropped.

### Non-blocking (should-fix or follow-up)

1. **Test C does not fully exercise the managed-DB scenario.** The test does `set role pgque_nonsuperuser` and then calls `pgque.maint()`. Because `maint()` is `SECURITY DEFINER`, the function body runs with the *owner's* privileges, not the caller's. In CI the install owner is `postgres` (a superuser), so even without the `pg_get_userbyid` fix, the `pg_authid` SELECT inside `maint()` would still succeed — the SECURITY DEFINER context restores superuser. The pre-flight check (caller cannot read `pg_authid`) does correctly assert that the *caller* is restricted, but the realistic managed-DB shape (install owner is a non-superuser) is not actually reproduced. To make Test C a true regression catcher, the install owner itself would need to be a non-superuser (e.g., `\\c -` reconnect as a non-superuser that owns the schema, or `alter function pgque.maint() owner to <nonsuper>` before the test). This is the same defect class that R1 flagged. Filing it as non-blocking S2 since the actual security fix is sound and CI does prove `pg_get_userbyid` doesn't itself raise; but the test would not have caught the original `pg_authid` bug if it had been written this way upfront.

2. **`docs/reference.md` `Grant: PUBLIC (default)` lines are now broadly stale.** As above — the tail blanket revoke means roughly 14 entries that say `Grant: PUBLIC (default)` (e.g., `start`, `stop`, `status`, `ticker`, `force_tick`, `create_queue`, `drop_queue`, `set_queue_config`, sequence helpers, queue-config setters, trigger functions) are no longer reachable by PUBLIC. The trailing paragraph at line 439 reinforces the same stale claim. Recommend a follow-up doc PR (or amend) to flip the narrative to deny-by-default and update each `Grant:` line to the actual role(s) that retain access.

### Potential
- **Same-superuser collapse (R1 S3 not addressed in R2 docs).** When `pgque_admin` is granted to the same role that owns `maint()` (single-superuser dev/test setups), the ownership check tautologically passes and the escalation path remains. R1 recommended documenting this; the R2 doc edits don't mention it. Not blocking — it's a deployment-specific caveat — but worth a sentence in the `maint()` entry.
- **`f.func_name` raw interpolation.** The `regprocedure` validation upstream guards injection, but a one-line comment explaining this invariant in `sql/pgque-api/maint.sql` would help future readers. Same point R1 raised, not addressed in R2. Non-blocking.
- **WARNING log spam DoS** from R1 stands as accepted trade-off; no action needed.

### Summary table

| Severity | Count |
|---|---|
| S1 (blocking) | 0 |
| S2 (should-fix / follow-up) | 2 (Test C scope, broader `reference.md` staleness) |
| S3 (nit / potential) | 2 (same-superuser-collapse callout, raw-name comment) |

### Anti-leak
No private references, internal URLs, or sensitive data quoted. All identifiers in the review are present in the public PR diff or PG catalog names.

---
*REV-style review (security, bugs, tests, guidelines, docs). SOC2 items skipped per project policy. Not approving, denying, or merging — flagging for the maintainer.*

NikolayS:
Comment-scrub sweep applied per CLAUDE.md style (commit 2634042): condensed Security note (issue #101) / Managed-DB compatibility (issue #96) blocks into short invariant notes in `maint.sql` and `pgque.sql`.

NikolayS:
## Managed-DB compatibility verification (head `2634042`)

Verdict: **RDS-COMPAT** with two minor test-runner caveats (not blockers).

### Setup (PostgreSQL 16, simulating RDS / Aurora / Cloud SQL / AlloyDB / Supabase / Neon / Crunchy Bridge)

```bash
# Non-superuser admin: CREATEDB + CREATEROLE, NOT SUPERUSER, NOT BYPASSRLS
sudo -u postgres psql <<'SQL'
create role rdsadmin_like login createdb createrole password 'rdsadmin_like';
create database pgque_rds_test owner rdsadmin_like;
SQL

# Install pgque.sql AS the non-superuser, into a DB it owns
PGPASSWORD=rdsadmin_like psql -h localhost -U rdsadmin_like -d pgque_rds_test \
  -v ON_ERROR_STOP=1 -f sql/pgque.sql
# exit: 0, no errors / no warnings
```

Confirmed `rdsadmin_like` is non-superuser and `select count(*) from pg_catalog.pg_authid` errors with `permission denied for table pg_authid` — faithful managed-DB simulation.

### Constraints checked

| # | Constraint | Result |
|---|---|---|
| 1 | `\i pgque.sql` succeeds as non-superuser owning the DB (no SUPERUSER, no BYPASSRLS) | PASS |
| 2 | All three roles `pgque_reader`/`pgque_writer`/`pgque_admin` created via plain `create role` (no SUPERUSER) | PASS |
| 3 | `revoke execute on all functions in schema pgque from public` succeeds when run by the function owner (non-superuser) | PASS — `proacl` shows no `=X/...` PUBLIC entry on send/receive/ack/maint/create_queue/dlq_purge/uninstall |
| 4 | `has_function_privilege('pgque_none_role', 'pgque.<fn>', 'EXECUTE')` returns `f` for an ungranted role across send/maint/create_queue/receive/ack/dlq_purge/uninstall | PASS |
| 5 | `pg_get_userbyid(p.proowner)` resolves without `SELECT` on `pg_authid` (the explicit managed-DB compat goal) | PASS — verified from a non-superuser session that gets `permission denied for pg_authid` but reads `pg_get_userbyid(...)` cleanly |
| 6 | SECURITY DEFINER `pgque.maint()` runs successfully under a non-owner managed-admin (`pgque_ops` granted `pgque_admin`) without touching `pg_authid` | PASS |
| 7 | SECURITY DEFINER ownership-check skips an attacker-owned `queue_extra_maint` entry with the documented `WARNING` | PASS — `pgque.evil_maint_probe` not executed; canary table empty |
| 8 | All `SECURITY DEFINER` functions in `pgque` have `search_path = pgque, pg_catalog` pinned | PASS — `pg_proc.proconfig` audit returns 0 rows missing the pin |
| 9 | `pgque.uninstall()` revoked from `pgque_admin` and from PUBLIC | PASS |
| 10 | `tests/test_security_public_execute.sql` as the non-superuser install owner | PASS |

### Test-runner caveats (NOT blockers, NOT bugs in the install path)

1. `tests/test_security_extra_maint.sql` Test C (the explicit "managed-DB simulation") uses `set role pgque_nonsuperuser`. On PostgreSQL 16 the historical CREATEROLE-implies-membership behavior is gone, so a non-superuser test runner with `CREATEROLE` cannot `set role` to a freshly-created role without an explicit `grant ... to current_user`. Symptom when run as `rdsadmin_like`: `ERROR: permission denied to set role "pgque_attacker"` at line 45. The test is effectively superuser-only as written. Suggested fix in a follow-up: add `grant pgque_attacker to current_user with admin option;` after creating the role (and likewise for `pgque_nonsuperuser`).

2. `tests/test_security_extra_maint.sql` Test B assumes the test session_user equals the install owner, so the "legit hook" is created with the same owner as `maint()`. When the test runner is `postgres` and the install owner is `rdsadmin_like`, the legit hook is owned by `postgres`, the ownership check correctly skips it (good), and Test B's assertion fails (`run_count = 0`). The security behavior is correct; the test scaffolding is brittle. Suggested follow-up: create the legit hook with `set role <maint owner>` derived from `pg_get_userbyid((select proowner from pg_proc where proname='maint' ...))`.

### Reproduction

Exact one-liner that demonstrates the install-as-non-superuser path works end-to-end:

```bash
PGPASSWORD=rdsadmin_like psql -h localhost -U rdsadmin_like -d pgque_rds_test \
  -v ON_ERROR_STOP=1 -f sql/pgque.sql && \
PGPASSWORD=rdsadmin_like psql -h localhost -U rdsadmin_like -d pgque_rds_test \
  -v ON_ERROR_STOP=1 -f tests/test_security_public_execute.sql
# both exit 0; "PASS: security_public_execute" printed
```

### Recommendation

Merge as-is for the security fix. File a follow-up to harden the two test scripts so they run on PG16+ without superuser (Caveats 1 and 2). The install path itself is fully managed-DB compatible.

---

NikolayS:
Scrubbed to ~+272/-32 (was +483/-14). No behavior change — all security guarantees preserved.

**What changed (commit `0f63687`):**
- Dropped 5 per-function `revoke` lines from `dlq.sql`, 3 from `receive.sql`, 8 from `send.sql`, and 1 from `maint.sql` — the blanket `revoke execute on all functions in schema pgque from public` at the end of `roles.sql` and at the tail of `pgque.sql` covers them all
- Trimmed the 11-line "Deny by default" comment block in `roles.sql`/`pgque.sql` to 1 line; trimmed the 4-line "Defense-in-depth" tail comment to 1 line
- Dropped the 4-line `queue_extra_maint` header comment in `maint.sql` (variable names and the regprocedure cast make intent clear); trimmed two 3-line `raise warning` concatenations to 1 line each
- Replaced the 72-line hardcoded 29-function list in `test_security_public_execute.sql` with a 15-line `pg_proc` introspection block that auto-covers future additions
- Dropped Test C from `test_security_extra_maint.sql` (~63 lines) — maint() is SECURITY DEFINER so it runs as install owner regardless of caller; Test B already covers the green path
- Dropped two implementation-detail paragraphs from `docs/reference.md`; the `Grant: pgque_admin` line change is the substantive update

---

NikolayS:
Rebased onto `origin/main` (`714007e`).

**Base SHA:** `d96817b` → `714007e`
**New head:** `4956718`

**Conflicts resolved** (all in comment blocks only — no logic changes):
- `sql/pgque-api/maint.sql` (×4 commits): Conflict was introductory comment block. PRs #119 dropped VACUUM from `maint()`, so "rotation step1, retry, and vacuum" text from this branch's commits was inaccurate. Resolved by keeping HEAD's accurate description ("rotation step1 and retry") while preserving all security code (ownership check, `pg_get_userbyid`, revoke/grant).
- `sql/pgque.sql` (×4 commits): Same inline comment conflict, same resolution.
- `docs/reference.md` (×1 commit): Kept HEAD description (no vacuum) + this PR's `pgque_admin` grant line + ownership check paragraphs. The scrub commit later removes the verbose paragraphs (auto-merged cleanly).

**One commit auto-dropped:** `72f193a` (docs: drop pre-release migration note) — already upstream.

**File line counts post-rebase:**
- `sql/pgque.sql`: 4884 lines (was ~4843 pre-rebase)

---

NikolayS:
## REV — Final round (post-rebase, post-scrub) on 4956718

**Verdict: PASS (1 non-blocking finding)**

### 1. Security — solid
- **PUBLIC EXECUTE coverage:** the trailing `revoke execute on all functions in schema pgque from public` in `sql/pgque-api/send.sql` is the LAST statement in the assembled `pgque.sql`. It is a proper catch-all that covers every function created in any source file (core, lifecycle, roles, dlq, maint, receive, send), including functions touched by intermediate merges (#119, #117, #116, #120, #122 — all of which modify existing functions, no new ones added). The earlier blanket revoke in `roles.sql` is a redundant defense-in-depth that runs against pre-API functions only.
- **`uninstall()`:** revoked from PUBLIC by the trailing blanket revoke and from `pgque_admin` explicitly at `roles.sql`. Effectively superuser-only. Correct.
- **`queue_extra_maint` ownership check:** uses `pg_get_userbyid(proowner)` (managed-DB safe — no `pg_authid` access required), resolves the target via `regprocedure` cast inside an EXCEPTION block, compares to `maint()`'s own owner, skips with WARNING on mismatch. The dynamic resolution of `maint()`'s owner per call is robust against rename/re-own.
- **SECURITY DEFINER + search_path:** `maint()` retains `set search_path = pgque, pg_catalog`. Confirmed.
- **Test coverage post-rebase:** `test_security_public_execute.sql` uses `pg_proc` introspection (not a hardcoded list), so it auto-extends to any function added by future PRs. Catches all 74 current functions in the schema.

### 2. Bugs — none introduced by rebase
- `maint()` regression-tested in CI green on PG 14/15/16/17/18 after rebasing onto #119 (VACUUM drop). The two blocks (#119's vacuum-skip and this PR's owner check) compose cleanly: vacuum branch `continue`s before owner check.
- All test files referenced in `tests/run_all.sql` exist on disk.
- The 4956718 fix addresses a real regression introduced by the scrub (the trailing revoke had only existed in committed `pgque.sql`, not in any source file, so `transform.sh` rebuilds dropped it). Now anchored in `send.sql`.

### 3. Tests — red→green correctness
- `test_security_public_execute`: queries `pg_proc` for any function executable by an ungranted role. Cannot pass for the wrong reason — the only way to satisfy it is to actually revoke PUBLIC.
- `test_security_extra_maint` Test A: attacker-owned `evil_maint_probe` registered in `queue_extra_maint`, asserts canary table is empty after `maint()`. Test B: legit same-owner hook MUST execute (run_count >= 1) — proves the owner check isn't trivially "skip everything." Both halves needed; both present.

### 4. Guidelines — compliant
- Lowercase SQL, snake_case, schema-qualified, search_path pinned.
- Conventional Commits (`fix(security):`, `chore:`, `fix(pgque.maint):`).
- No person names, no GitLab/internal/Round/WI references in commit messages or code. `Co-Authored-By: Claude` lines are acceptable per project convention.

### 5. Docs — one stale claim (NON-BLOCKING)
**`docs/reference.md` line 438** still states:
> `pgque.uninstall()` is revoked from `pgque_admin`, but **not from PUBLIC** — any role that can connect can still call it and drop the schema. Tighten with `revoke execute on function pgque.uninstall() from public;` if stricter control is required.

This contradicts the deny-by-default posture established by the trailing blanket revoke. After this PR, `uninstall()` IS revoked from PUBLIC (via the catch-all). The "tighten with…" sentence is now describing the actual default, not an opt-in. Same paragraph then correctly says all other functions are deny-by-default — internally inconsistent.

**Suggested edit:**
```
`pgque.uninstall()` is revoked from both `pgque_admin` and PUBLIC — only the
schema owner / superuser can run it. All other functions in the schema follow
the same deny-by-default posture (`revoke execute … from public` is applied
globally); use explicit `grant execute` to open them to additional roles.
```

The `Grant: pgque_admin` line for `maint()` (line 167) is correct.

### 6. Architecture — clean
- No composite/non-1:1 abstraction introduced.
- Deny-by-default realized through one mechanism (blanket revoke) anchored at the tail of the assembled file. The duplicate revoke in `roles.sql` is harmless redundancy with intent-documenting value; could be removed but not required.
- Security logic stays at the security boundary: ownership check is inside the SECURITY DEFINER function it protects.

### Summary
- **Blockers: 0**
- **Non-blocking: 1** — `docs/reference.md` line 438 contradicts the new posture; please fix in this PR or a follow-up.
- CI: PG 14/15/16/17/18 all green; `claude-review` and `client-smoke` still in progress at time of review.

Rebase preserved correctness; the scrub introduced a regression that 4956718 correctly fixed by anchoring the trailing revoke in source rather than in the generated file.

---

NikolayS:
Nit addressed (commit `384694c`). Updated `docs/reference.md` line ~438 to reflect the deny-by-default posture introduced by the schema-wide blanket `revoke execute on all functions in schema pgque from public`. Old wording implied PUBLIC still had execute on `uninstall()` and suggested a manual tighten step. New wording:

> `pgque.uninstall()` is revoked from both `pgque_admin` (explicitly) and PUBLIC (via the schema-wide blanket revoke). Only the schema/install owner (typically a superuser) can run it.

---

PR #120, opened 2026-04-30
fix(pgque.batch_retry): cast NULL to xid8 not int8

### LINKED ISSUE #107: Bug: batch_retry() fails on xid8 schema (NULL::int8 inserted into ev_txid xid8)
## Summary

`pgque.batch_retry(batch_id, retry_seconds)` is broken on the current PG14+/`xid8` schema.

Calling it for a valid active batch fails with a type mismatch:

```text
ERROR: column "ev_txid" is of type xid8 but expression is of type bigint
LINE 6: b.ev_id, b.ev_time, NULL::int8, _s.sub_id, ...
```

## Repro

```sql
\i sql/pgque.sql

select pgque.create_queue('batch_retry_dup');
select pgque.subscribe('batch_retry_dup', 'c');
select pgque.send_batch('batch_retry_dup', 'evt', array['x','y']::text[]);
select pgque.force_tick('batch_retry_dup');
select pgque.ticker();

select pgque.next_batch('batch_retry_dup', 'c') as batch_id \gset
select pgque.batch_retry(:batch_id, 0);
```

Observed:

```text
ERROR: column "ev_txid" is of type xid8 but expression is of type bigint
```

## Cause

`batch_retry()` inserts:

```sql
NULL::int8
```

into `retry_queue.ev_txid`, which is `xid8`.

Should be `NULL::xid8` or plain `NULL`.

## Impact

The exposed low-level primitive is unusable on the current schema. It also hides/blocks testing of cross-consumer `batch_retry()` interference paths similar to #102/#106.

## Suggested test

Add regression:

- create queue, subscribe, send two events, tick
- `next_batch()`
- `batch_retry(batch_id, 0)` returns 2
- second call is idempotent / returns 0
- retry rows redeliver after `maint_retry_events()` + tick

## Environment

Tested on `main` at `9b3f89f` with PostgreSQL 18.3.

NikolayS:
Closes #107

## Summary

`pgque.batch_retry()` was inserting `NULL::int8` into `retry_queue.ev_txid`, which is an `xid8` column (modernized from `bigint` during the PgQ→PgQue transform). PostgreSQL 14+ has no implicit cast from `int8` to `xid8`, so every call to `batch_retry()` failed with:

```
ERROR: column "ev_txid" is of type xid8 but expression is of type bigint
```

## Fix

Changed `NULL::int8` → `NULL::xid8` in `pgque.batch_retry()`:

- **`sql/pgque.sql`** line 2437: the actual runtime fix
- **`build/transform.sh`**: added a post-transform `sedi` patch so future regenerations from the PgQ submodule apply the same correction automatically

No other `NULL::int8` casts into `xid8` columns were found in the codebase (audited via `grep -rn "NULL::int8\|NULL::bigint"`).

## Test

Added `tests/test_core_batch_retry.sql` (also wired into `tests/run_all.sql`) covering:

1. `batch_retry(batch_id, 0)` returns event count without type mismatch (the core regression)
2. Second call is idempotent (returns 0)
3. `maint_retry_events()` + tick re-delivers events with `ev_retry >= 1`

**Red commit** `7c4b286` — test fails with exact error from #107  
**Green commit** `1774de5` — all tests pass

```
Running: test_core_batch_retry
NOTICE:  PASS: batch_retry() returned 2 rows (no xid8 type mismatch)
NOTICE:  PASS: batch_retry() idempotent on second call
NOTICE:  PASS: retried events redelivered (count=2, min ev_retry=1)
NOTICE:  PASS: core_batch_retry
...
=== ALL TESTS PASSED ===
```

### DIFF
--- a/build/transform.sh
+++ b/build/transform.sh
@@ -533,6 +533,16 @@ awk '
 
 echo "PASS: insert_event_raw raises clear error on unknown queue"
 
+# Fix batch_retry(): NULL::int8 cast into retry_queue.ev_txid which is xid8.
+# The ev_txid column was modernized to xid8 by the transform, but this explicit
+# NULL cast in the INSERT was not updated. Fix: NULL::xid8.
+# See: https://github.com/NikolayS/pgque/issues/107
+BATCH_RETRY_FILE="${OUTPUT_DIR}/functions/pgque.batch_retry.sql"
+sedi 's/b\.ev_id, b\.ev_time, NULL::int8, _s\.sub_id,/b.ev_id, b.ev_time, NULL::xid8, _s.sub_id,/' \
+  "${BATCH_RETRY_FILE}"
+
+echo "PASS: batch_retry NULL::int8 -> NULL::xid8 for ev_txid (xid8) column"
+
 # -- Assembly: build sql/pgque.sql ------------------------------------
 
 echo ""
--- a/build/transform.sh
+++ b/build/transform.sh
@@ -533,10 +533,6 @@ awk '
 
 echo "PASS: insert_event_raw raises clear error on unknown queue"
 
-# Fix batch_retry(): NULL::int8 cast into retry_queue.ev_txid which is xid8.
-# The ev_txid column was modernized to xid8 by the transform, but this explicit
-# NULL cast in the INSERT was not updated. Fix: NULL::xid8.
-# See: https://github.com/NikolayS/pgque/issues/107
 BATCH_RETRY_FILE="${OUTPUT_DIR}/functions/pgque.batch_retry.sql"
 sedi 's/b\.ev_id, b\.ev_time, NULL::int8, _s\.sub_id,/b.ev_id, b.ev_time, NULL::xid8, _s.sub_id,/' \
   "${BATCH_RETRY_FILE}"
--- a/sql/pgque.sql
+++ b/sql/pgque.sql
@@ -2434,7 +2434,7 @@ begin
         ev_type, ev_data, ev_extra1, ev_extra2,
         ev_extra3, ev_extra4)
     select distinct _retry, _s.sub_queue,
-           b.ev_id, b.ev_time, NULL::int8, _s.sub_id, coalesce(b.ev_retry, 0) + 1,
+           b.ev_id, b.ev_time, NULL::xid8, _s.sub_id, coalesce(b.ev_retry, 0) + 1,
            b.ev_type, b.ev_data, b.ev_extra1, b.ev_extra2,
            b.ev_extra3, b.ev_extra4
       from pgque.get_batch_events(i_batch_id) b
--- a/tests/run_all.sql
+++ b/tests/run_all.sql
@@ -37,6 +37,9 @@
 \echo 'Running: test_core_retry'
 \i tests/test_core_retry.sql
 
+\echo 'Running: test_core_batch_retry'
+\i tests/test_core_batch_retry.sql
+
 \echo 'Running: test_core_rotation'
 \i tests/test_core_rotation.sql
 
--- a/tests/test_core_batch_retry.sql
+++ b/tests/test_core_batch_retry.sql
@@ -1,0 +1,110 @@
+-- test_core_batch_retry.sql -- Regression test for batch_retry() xid8 cast bug
+-- Copyright 2026 Nikolay Samokhvalov. Apache-2.0 license.
+--
+-- Regression: batch_retry() was inserting NULL::int8 into retry_queue.ev_txid
+-- which is xid8. Fix: use NULL::xid8.
+-- See: https://github.com/NikolayS/pgque/issues/107
+
+-- Step 1: setup
+do $$
+begin
+  perform pgque.create_queue('test_batch_retry');
+  perform pgque.register_consumer('test_batch_retry', 'c1');
+end $$;
+
+-- Step 2: insert two events (separate transaction for snapshot visibility)
+do $$
+begin
+  perform pgque.insert_event('test_batch_retry', 'batch.retry.test', 'evt1');
+  perform pgque.insert_event('test_batch_retry', 'batch.retry.test', 'evt2');
+end $$;
+
+-- Step 3: tick so events are visible to consumers
+do $$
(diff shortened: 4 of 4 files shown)


NikolayS:
## REV review: PR #120 -- fix(pgque.batch_retry): cast NULL to xid8 not int8

**Verdict:** Looks good (advisory-only; not approving). Confidence: **HIGH**. Classification: **MINOR_NITS**.

### Counts

| Bucket | Count |
| --- | --- |
| Critical | 0 |
| Major | 0 |
| Minor / nit | 2 |
| Anti-leak hits | 0 |

### Step 1 -- context verified

- Diff is 140+ / 1- across 4 files: `build/transform.sh` (sedi patch), `sql/pgque.sql` (line 2437 fix), `tests/run_all.sql` (wire-in), `tests/test_core_batch_retry.sql` (new red+green test). Closes #107.
- Two commits: `7c4b286` red test for #107, `1774de5` green fix. Order matches the project's red/green TDD rule.
- CI: `verify`, `test (14|15|16|17|18)` all green at the time of review.
- Issue #107 repro and root-cause description match exactly what this PR fixes.

### Step 2 -- Bug hunter (HEADLINE)

Independently re-ran the audit on the PR branch:

```
grep -nE "NULL::int8|NULL::bigint" sql/pgque.sql sql/pgque-additions/*.sql sql/pgque-api/*.sql sql/experimental/*.sql
```

Result: **zero hits** post-fix. The original PgQ source has exactly one such cast (`pgq/functions/pgq.batch_retry.sql:38`) -- it lands in `retry_queue.ev_txid`, which the transform modernizes to `xid8`. So the engineer's audit claim is independently confirmed: this was the only mismatched `NULL::int8 -> xid8` in the codebase.

**`sedi` patch robustness** (`build/transform.sh:540-544`):

- `sedi` is defined at line 442 with portable BSD-vs-GNU `sed -i` branching by `uname -s`. macOS + Linux both work. Good.
- Script uses `set -Eeuo pipefail` at line 2, so any sed failure aborts the build. Good.
- Pattern is well-anchored on `b\.ev_id, b\.ev_time, NULL::int8, _s\.sub_id,` -- specific enough that it cannot accidentally rewrite a legitimately-`int8` `NULL` cast elsewhere. Good.
- **Minor (nit) #1 -- silent-no-op risk on upstream drift.** The sedi has no post-check that the substitution actually fired. If a future PgQ release renames `_s` -> `s` or reflows the SELECT list, the pattern silently no-ops and the build still succeeds with a broken `sql/pgque.sql`. Recommend adding an assertion immediately after the sedi call, e.g.:

  ```bash
  grep -q 'NULL::xid8, _s\.sub_id,' "${BATCH_RETRY_FILE}" \
    || { echo "ERROR: batch_retry NULL::xid8 patch did not apply"; exit 1; }
  ```

  This is consistent with the style already used elsewhere (the `PASS:` echos are advisory-only -- they don't actually verify the substitution). Not a blocker because the new `tests/test_core_batch_retry.sql` would catch a regression in CI on PG14-18, but a build-time assertion fails earlier and with a clearer message.

- **Edge case -- `NULL::xid8` semantics.** `xid8` is a distinct type but `NULL` is type-only metadata; PG resolves to `xid8 NULL` at the column level either way. No semantic change for downstream consumers (`pg_visible_in_snapshot`, `<` / `<=` comparisons): `ev_txid IS NULL` rows are never compared, only filtered. Verified upstream consumers (`batch_event_sql`) handle `NULL ev_txid` rows via the standard `<= tx_end` window which a `NULL` row simply won't satisfy -- correct.

### Step 2 -- Test analyzer

- Red commit `7c4b286` and green commit `1774de5` are properly ordered. Per `CLAUDE.md` rule 6, this satisfies red/green TDD for a bug fix.
- The test exercises the exact failure mode from #107 (full repro: queue + consumer + 2 events + tick + `next_batch` + `batch_retry(batch_id, 0)`).
- Idempotency check (`batch_retry` second call returns 0) is included.
- Re-delivery path is exercised via `maint_retry_events` + tick + assert `ev_retry >= 1`.
- **Minor (nit) #2 -- assertion on `xid8` storage type is implicit.** The test asserts row counts and `ev_retry` values, but doesn't directly assert that `retry_queue.ev_txid IS NULL` and is typed `xid8`. The cast bug is currently caught by "function does not error", which is sufficient -- but a one-line `assert pg_typeof((select ev_txid from pgque.retry_queue where ev_id = ...)) = 'xid8'::regtype` would document the invariant explicitly. Optional, file-and-forget.

### Step 2 -- Guidelines

- `CLAUDE.md` SQL style: lowercase keywords -- yes. `snake_case` -- yes. Schema-qualified -- yes.
- Conventional Commits: `fix(batch_retry): ...` and `test: red test for ...` -- yes, both <50 chars.
- Source (`build/transform.sh`) and regenerated (`sql/pgque.sql`) are in the same commit -- yes. This satisfies the project rule that the transform is the source of truth.
- Shell style: 2-space indent in the new sedi block -- yes. Quoted vars `"${BATCH_RETRY_FILE}"` -- yes.

### Step 2 -- Docs

- `docs/reference.md` line 383 entry for `pgque.batch_retry()` only states return type and behavior at a high level (`returns the number of events enqueued for retry`). No claim about cast types. No doc change needed.
- Migration note for stuck rows: not needed. The function was 100% broken on the modernized schema -- any prior caller hit the type mismatch and rolled back, so no partial/stuck rows exist in the wild.

### Step 4 -- Anti-leak

`grep -iE "gitlab|sahmed|artifact[_-]?registry|@AR\b|round[ -]?[4-9]\b|R[4-9]\b"` over diff + commit messages + PR body: 0 hits.

### Summary

Tight, well-scoped fix. The transform-level patch is the right place (any fix only in `sql/pgque.sql` would be lost on the next regeneration). Two purely-optional nits noted above; neither should block merge.

(Posted as advisory review -- not approving / merging per REV protocol.)

NikolayS:
Comment-scrub sweep applied per CLAUDE.md style (commit 3c31fa3): removed narrative comment block (Fix batch_retry / issue #107 / See URL) before the sedi patch in `build/transform.sh`.

PR #122, opened 2026-04-30
fix(pgque.create_queue): reject queue names > 57 bytes

### LINKED ISSUE #109: Bug: queue names longer than 57 bytes fail via pg_notify channel name in ticker/create_queue
## Summary

Queue names longer than 57 bytes fail because PgQue uses the raw queue name in the NOTIFY channel:

```sql
pg_notify('pgque_' || i_queue_name, tick_id::text)
```

PostgreSQL channel names are limited to 63 bytes. The `pgque_` prefix consumes 6 bytes, leaving 57 bytes for `queue_name`.

This also affects `create_queue()`, because `create_queue()` immediately calls `pgque.ticker(i_queue_name)`, which emits the notification.

## Repro

```sql
\i sql/pgque.sql

-- 57 chars works: length('pgque_' || queue_name) = 63
select pgque.create_queue(repeat('q', 57));

-- 58 chars fails: length('pgque_' || queue_name) = 64
select pgque.create_queue(repeat('q', 58));
```

Observed for longer names:

```text
ERROR: channel name too long
CONTEXT: SQL statement "SELECT pg_notify('pgque_' || i_queue_name, currval(q.queue_tick_seq)::text)"
PL/pgSQL function ticker(text) line 99 at PERFORM
SQL statement "SELECT pgque.ticker(i_queue_name)"
PL/pgSQL function create_queue(text) line 73 at PERFORM
```

## Expected

Either:

- validate queue names up front with a clear error and documented max length, or
- derive a safe fixed-length notification channel, e.g. using queue id or hash, not raw queue name.

Using queue id might be best:

```text
pgque_<queue_id>
```

with helper function to compute/listen channel for clients.

## Tests

- queue name at max supported length succeeds
- queue name over max fails with clear validation error, or succeeds if hashed/id channel is used
- docs mention LISTEN/NOTIFY channel naming

## Environment

NikolayS:
## Summary

Closes #109

Queue names longer than 57 bytes fail at `ticker()` with PostgreSQL's own
opaque `channel name too long` error because `pg_notify('pgque_' || name, ...)`
produces a channel name > 63 bytes.

**Option A chosen (fail-fast validation):** Validate the queue name length at
the top of `create_queue()` before any sequences, tables, or rows are created.
This gives callers a clear, actionable error and matches PostgreSQL's own
identifier-length convention (e.g. `NAMEDATALEN - 1 = 63`).

The limit is 57 bytes: `63 (PG channel max) - 6 (len('pgque_'))`.

## Changes

- `sql/pgque.sql`: `octet_length(i_queue_name) > 57` check in `create_queue()`,
  raises `queue name too long: N bytes (max 57)` before any state is written.
- `build/transform.sh`: corresponding post-transform patch so future `./build/transform.sh`
  runs keep the fix.

## Test output

```
PASS: 57-byte queue name accepted
PASS: 58-byte queue name rejected with: queue name too long: 58 bytes (max 57). pg_notify channel 'pgque_<queue_name>' must fit in 63 bytes.
PASS: 63-byte queue name rejected
=== ALL TESTS PASSED ===
```

### DIFF
--- a/build/transform.sh
+++ b/build/transform.sh
@@ -476,6 +476,7 @@ awk '
 /^    if i_queue_name is null then$/ {
   print
   in_null_check = 1
+  next
 }
 in_null_check && /^    end if;$/ {
   q = "\047"
--- a/build/transform.sh
+++ b/build/transform.sh
@@ -478,15 +478,16 @@ awk '
   in_null_check = 1
 }
 in_null_check && /^    end if;$/ {
+  q = "\047"
   print
   print ""
   print "    -- pg_notify channel names are limited to 63 bytes by PostgreSQL."
-  print "    -- PgQue prefixes them with '\''pgque_'\'' (6 bytes), leaving 57 bytes for"
+  print "    -- PgQue prefixes them with " q "pgque_" q " (6 bytes), leaving 57 bytes for"
   print "    -- the queue name.  Reject names that would overflow the channel name"
   print "    -- before any state is written, so callers get a clear error."
   print "    if octet_length(i_queue_name) > 57 then"
-  print "        raise exception '\''queue name too long: % bytes (max 57). '\"'\"'"
-  print "            '\''pg_notify channel '\''pgque_<queue_name>'\'' must fit in 63 bytes.'\'','\"'\"'"
+  print "        raise exception " q "queue name too long: % bytes (max 57). " q
+  print "            " q "pg_notify channel " q q "pgque_<queue_name>" q q " must fit in 63 bytes." q ","
   print "            octet_length(i_queue_name);"
   print "    end if;"
   in_null_check = 0
--- a/build/transform.sh
+++ b/build/transform.sh
@@ -466,6 +466,38 @@ awk '
 
 echo "PASS: pg_notify injected into ticker function"
 
+# Patch create_queue: validate queue name length before any state is written.
+# pg_notify channel names are limited to 63 bytes by PostgreSQL.  PgQue
+# prefixes them with 'pgque_' (6 bytes), so queue names > 57 bytes would
+# overflow the channel name and produce a cryptic PG error in ticker().
+# Inject the length check right after the existing NULL guard.
+CREATE_QUEUE_FILE="${OUTPUT_DIR}/functions/pgque.create_queue.sql"
+awk '
+/^    if i_queue_name is null then$/ {
+  print
+  in_null_check = 1
+}
+in_null_check && /^    end if;$/ {
+  print
+  print ""
+  print "    -- pg_notify channel names are limited to 63 bytes by PostgreSQL."
+  print "    -- PgQue prefixes them with '\''pgque_'\'' (6 bytes), leaving 57 bytes for"
+  print "    -- the queue name.  Reject names that would overflow the channel name"
+  print "    -- before any state is written, so callers get a clear error."
+  print "    if octet_length(i_queue_name) > 57 then"
+  print "        raise exception '\''queue name too long: % bytes (max 57). '\"'\"'"
+  print "            '\''pg_notify channel '\''pgque_<queue_name>'\'' must fit in 63 bytes.'\'','\"'\"'"
+  print "            octet_length(i_queue_name);"
+  print "    end if;"
+  in_null_check = 0
+  next
+}
+in_null_check { print; next }
+{ print }
+' "${CREATE_QUEUE_FILE}" > "${CREATE_QUEUE_FILE}.tmp" && mv "${CREATE_QUEUE_FILE}.tmp" "${CREATE_QUEUE_FILE}"
+
+echo "PASS: create_queue queue name length check added"
+
 # Fix inherited PgQ copy-paste bug: sqltriga comment says logutriga
 sedi 's/Function: pgque.logutriga()/Function: pgque.sqltriga()/' "${OUTPUT_DIR}/lowlevel_pl/sqltriga.sql"
 
@@ -626,6 +658,7 @@ echo "--   4. pg_current_xact_id() cast to ::text::bigint (xid8→bigint)" >> "$
 echo "--   5. SECURITY DEFINER functions get SET search_path = pgque, pg_catalog" >> "${INSTALL_FILE}"
 echo "--   6. pgq_node/Londiste hooks removed from maint_operations" >> "${INSTALL_FILE}"
 echo "--   7. pg_notify() injected into ticker for LISTEN/NOTIFY wakeup" >> "${INSTALL_FILE}"
+echo "--   8. create_queue() rejects queue names > 57 bytes (pg_notify limit)" >> "${INSTALL_FILE}"
 echo "-- ======================================================================" >> "${INSTALL_FILE}"
 echo "" >> "${INSTALL_FILE}"
 
--- a/build/transform.sh
+++ b/build/transform.sh
@@ -466,11 +466,6 @@ awk '
 
 echo "PASS: pg_notify injected into ticker function"
 
-# Patch create_queue: validate queue name length before any state is written.
-# pg_notify channel names are limited to 63 bytes by PostgreSQL.  PgQue
-# prefixes them with 'pgque_' (6 bytes), so queue names > 57 bytes would
-# overflow the channel name and produce a cryptic PG error in ticker().
-# Inject the length check right after the existing NULL guard.
 CREATE_QUEUE_FILE="${OUTPUT_DIR}/functions/pgque.create_queue.sql"
 awk '
 /^    if i_queue_name is null then$/ {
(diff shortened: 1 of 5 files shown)


NikolayS:
## REV review — PR #122 fix(pgque.create_queue): reject queue names > 57 bytes

**Verdict:** LOOP — release-blocking issue in `build/transform.sh`.

The SQL fix and tests look right. The build script change does not run.

### Confidence
High on the build-script bug (reproduced locally), medium on the smaller items.

### Counts
- BLOCKERS: 1
- MAJOR: 1
- MINOR: 3
- NITS: 2

---

### BLOCKER

**B1. `build/transform.sh` has a bash syntax error and aborts on every CI run.**
All `test (14|15|16|17|18)`, `verify`, and `client-smoke` jobs are red on this PR with:

```
build/transform.sh: line 490: syntax error near unexpected token `('
build/transform.sh: line 490: `  print "            octet_length(i_queue_name);"'
```

Reproducible locally with `bash -n build/transform.sh`. Root cause: the `'\''queue name too long: % bytes (max 57). '\"'\"'` / `'\''pg_notify channel ...'\'','\"'\"'` quoting inside the awk-in-shell heredoc-style block leaves bash in an unquoted state by line 490, so `octet_length(i_queue_name)` is parsed as shell. `sql/pgque.sql` itself is fine because it is the hand-edited regenerated artifact; the patch breaks future regenerations.

Suggested fix: rewrite this awk block as a separate `awk -f patch.awk` file, or use a `printf '%s\n' '...'` block, or simplest of all — drop the embedded inner single-quote escapes by using `chr(39)` in the awk script:

```awk
print "    if octet_length(i_queue_name) > 57 then"
q = "\047"  # awk for single quote
print "        raise exception " q "queue name too long: % bytes (max 57). " q
...
```

Or wrap the awk program in a heredoc passed to `awk -f /dev/stdin <<'AWK' ... AWK` so the shell never has to escape the inner SQL quotes.

Whatever the chosen form, please confirm with `bash -n build/transform.sh && ./build/transform.sh` before re-pushing.

---

### MAJOR

**M1. The fix needs to land in `pgq` source (`sql/pgque-core/...` / functions/) plus regen, not in `sql/pgque.sql` directly.**
Looking at the diff, `sql/pgque.sql` is regenerated by `build/transform.sh` from `pgq/`. Currently the SQL change only lives in:
1. `sql/pgque.sql` (regenerated artifact, hand-edited)
2. `build/transform.sh` (post-transform patch — broken, see B1)

If somebody runs `./build/transform.sh` after the broken patch is fixed, fine. But there is no upstream-hostile patch that survives a future `pgq` submodule bump cleanly — this is consistent with how other PgQue patches work (`pg_notify` injection, xid8 fix), so it is not strictly wrong, just worth confirming the maintainer is happy with the all-in-`transform.sh` approach. CLAUDE.md "Source + regen `sql/pgque.sql` in same commit" is satisfied iff B1 is fixed.

---

### MINOR

**m1. `docs/reference.md` does not document the 57-byte limit.**
`pgque.create_queue(queue text)` entry at `docs/reference.md:107` says only "Creates a queue with default settings...". With this PR it can now raise on long names — please add a one-liner like "Queue names are limited to 57 bytes (UTF-8); the `pgque_<name>` LISTEN/NOTIFY channel must fit in 63 bytes." Same applies to the 2-arg overload at line 475.

**m2. Tests miss multi-byte UTF-8 boundary.**
`tests/test_queue_name_length.sql` uses `repeat('q', N)` (1 byte/char). Because the fix correctly uses `octet_length` (good!), please add at least one positive and one negative case at the byte boundary with non-ASCII, e.g.:
- 19 × 3-byte char (57 bytes, accept) — `repeat('日', 19)`
- 20 × 3-byte char (60 bytes, reject)

This locks in `octet_length` vs `length` so a future "simplification" to `length()` is caught.

**m3. Tests do not assert no partial state remains after a rejected create.**
The fix is correctly placed before any DDL/DML, so this is fine in practice — but a regression that moves the check below the `insert into pgque.queue` row would not be caught. Add after the 58-byte rejection:
```sql
assert not exists (select 1 from pgque.queue where queue_name = repeat('q', 58)),
  'rejected queue must not leave partial state';
```

---

### NITS

**n1. Edge case worth a comment, not a fix.** Empty queue name `''` has `octet_length = 0`, so the fix does not reject it; the existing PgQ code path also accepts empty names (it goes on to create `pgque.event_<id>` tables since identifiers are derived from queue_id, not the name). Out of scope for this PR but maybe worth a follow-up issue: should `create_queue('')` be rejected?

**n2. Error message wording.** Current message:
```
queue name too long: 58 bytes (max 57). pg_notify channel 'pgque_<queue_name>' must fit in 63 bytes.
```
Reads clearly. Consider tightening "must fit in 63 bytes" → "is limited to 63 bytes by PostgreSQL" so the user does not have to remember why 63. Cosmetic only.

---

### Other perspectives (clean)

- **Security:** `SECURITY DEFINER set search_path = pgque, pg_catalog` preserved on `create_queue`. Fix runs before any state mutation, so no partial-state risk. No new attack surface.
- **Bug hunter, other call sites:** Verified `pgque_` is the only prefix used by `pg_notify` (sql/pgque.sql:687, 792), so 57 is the right limit. Other identifier construction in `create_queue` (`pgque.event_<id>`, `_tick_seq`, `_id_seq`) uses `queue_id` (integer), not `queue_name`, so they are unaffected by name length. No other call sites construct PG identifiers from the raw queue name.
- **`octet_length` vs `length`:** correct choice. Bytes, not characters, are the PG identifier limit.
- **Race:** two concurrent `create_queue` calls with too-long names both raise before touching state — no race.
- **Conventional Commits / commit history:** `fix(pgque.create_queue):` and `test:` — clean.
- **Anti-leak scan:** clean.

---

### Suggested next loop
1. Rewrite the `build/transform.sh` patch so `bash -n build/transform.sh` passes, then run `./build/transform.sh` and confirm `sql/pgque.sql` matches what is checked in.
2. Add docs sentence in `docs/reference.md` for both `create_queue` overloads.
3. Add the multi-byte UTF-8 test case and the no-partial-state assert.

Not approving / not merging — independent reviewer, leaving the call to the maintainer after the next push.

NikolayS:
## REV Review — PR #122 (round 2)

**CI:** 7/7 green (PG 14-18 + verify + client-smoke + claude-review)
**R1 BLOCKING resolution:** addressed
**R1 minors status:** m1 addressed / m2 addressed / m3 addressed

### R1-resolution verification

| R1 finding | Commit | Status | Notes |
|---|---|---|---|
| BLOCKING (bash quoting) | `4ee5f42` | addressed | `bash -n build/transform.sh` clean. Awk uses `q = "\047"` and concatenation; `awk 'BEGIN { print "\047" }'` returns `'`. Ran the awk block on a synthetic `create_queue.sql` — emits valid PL/pgSQL identical to what is checked into `sql/pgque.sql:1452-1463`. |
| Missing `next` | `d8932e5` | addressed | Both action blocks (`/^    if i_queue_name is null then$/` and `in_null_check && /^    end if;$/`) end with `next`, plus an `in_null_check { print; next }` passthrough — no double-emission of `end if;`. |
| m1 (docs) | in PR | addressed | `docs/reference.md:109` (1-arg overload) and `:477` (2-arg overload) both mention the 57-byte limit and the `pgque_<name>` channel reason. |
| m2 (UTF-8 boundary) | in PR | addressed | `tests/test_queue_name_length.sql:64` accepts `repeat('日', 19)` (57 bytes) and `:80` rejects `repeat('日', 20)` (60 bytes). Locks in `octet_length` semantics. |
| m3 (no partial state) | in PR | addressed | `tests/test_queue_name_length.sql:102` asserts `not exists (... from pgque.queue ...)` after a rejected 58-byte create. |

### Blocking
None.

### Non-blocking
None.

### Potential
None new in r2. Earlier nits (n1 empty-name, n2 wording) remain out of scope and non-blocking.

### Anti-leak
Clean (diff + body + commits).

### Quick sweep

- **Bug hunter:** Awk regex anchors are tight (`/^    if i_queue_name is null then$/` requires exact 4-space indent matching the regenerated source). The `q q` doubled-quote pattern correctly produces SQL-escaped single quotes inside the literal (`''pgque_<queue_name>''`). No regression vectors spotted.
- **Test analyzer:** Boundary cases test the right behavior (octet_length, not length); 57/58 ASCII + 57/60 UTF-8 + no-partial-state + 63 ASCII — comprehensive.
- **Docs/guidelines:** Both overloads documented; `octet_length` matches CLAUDE.md PG14+ idiom; SECURITY DEFINER `set search_path = pgque, pg_catalog` preserved upstream.

---

NikolayS:
Comment-scrub sweep applied per CLAUDE.md style (commit 6ac8a43): removed narrative comment block ('Patch create_queue: validate queue name length...') before the awk patch in `build/transform.sh`.

PR #130, opened 2026-04-30
fix(clients/typescript): per-queue ticker overload, fix nack test schema, serialise vitest

NikolayS:
## Summary

- **Drop `Client.flush()`**: composite `force_tick + ticker` is test-helper convenience, not core driver API. Drivers are 1:1 wrappers of SQL primitives.
- **`Client.ticker(queue?: string)`**: updated to accept an optional queue argument. With a queue it calls `pgque.ticker(queue text)`; without one it calls the no-arg global overload. Existing callers with no argument are unaffected.
- **`Client.forceTick(queue)`**: unchanged, exact wrapper for `pgque.force_tick($1)`.
- **`advanceQueue(client, queue)` test helper**: added to `test/helpers.ts`, exported for test-only use. Composes `forceTick` + `ticker` — the composition lives in the test layer, not the public API.
- **All test call sites** updated from `client.flush(queue)` to `advanceQueue(client, queue)`.
- **`nack.test.ts`**: column reference corrected from `ev_retry_after_ts` to `ev_retry_after` to match actual schema.
- **`vitest.config.ts`**: `fileParallelism: false` — prevents cross-suite races on the shared Postgres database.

## Test plan

- [ ] `cd clients/typescript && npm test` passes with `PGQUE_TEST_DSN` set
- [ ] No public `Client` method named `flush` exists
- [ ] `client.ticker()` (no arg) and `client.ticker(queue)` (with arg) both compile and call correct SQL
- [ ] `advanceQueue` is not exported from `src/index.ts` (test-only)

### DIFF
--- a/clients/typescript/src/client.ts
+++ b/clients/typescript/src/client.ts
@@ -234,23 +234,19 @@ export class Client {
     }
   }
 
-  /**
-   * Force a tick on `queue` and immediately rotate the ticker so any
-   * pending events become visible in a new batch. Mostly useful for tests
-   * that need deterministic rotation without waiting for pg_cron.
-   *
-   * Calls `pgque.force_tick(queue)` (bumps the event sequence) and then
-   * `pgque.ticker()` (performs the rotation). Each runs on a separate pool
-   * connection, ensuring the producing transaction's events are visible
-   * to the ticker's snapshot.
-   */
+  /** Exact wrapper for pgque.force_tick(queue). Bumps the event-seq threshold so the next ticker run produces a tick. */
   async forceTick(queue: string): Promise<void> {
     try {
       await this.pool.query('select pgque.force_tick($1)', [queue]);
-      await this.pool.query('select pgque.ticker()');
     } catch (err) {
       throw mapPgError('force_tick', err, { queue });
     }
+  }
+
+  /** Test/demo helper: force the tick threshold for `queue`, then run the per-queue ticker. */
+  async flush(queue: string): Promise<void> {
+    await this.pool.query('select pgque.force_tick($1)', [queue]);
+    await this.pool.query('select pgque.ticker($1)', [queue]);
   }
 }
 
--- a/clients/typescript/src/client.ts
+++ b/clients/typescript/src/client.ts
@@ -234,10 +234,20 @@ export class Client {
     }
   }
 
-  /** Run `pgque.force_tick(queue)`. Mostly useful for tests. */
+  /**
+   * Force a tick on `queue` and immediately rotate the ticker so any
+   * pending events become visible in a new batch. Mostly useful for tests
+   * that need deterministic rotation without waiting for pg_cron.
+   *
+   * Calls `pgque.force_tick(queue)` (bumps the event sequence) and then
+   * `pgque.ticker()` (performs the rotation). Each runs on a separate pool
+   * connection, ensuring the producing transaction's events are visible
+   * to the ticker's snapshot.
+   */
   async forceTick(queue: string): Promise<void> {
     try {
       await this.pool.query('select pgque.force_tick($1)', [queue]);
+      await this.pool.query('select pgque.ticker()');
     } catch (err) {
       throw mapPgError('force_tick', err, { queue });
     }
--- a/clients/typescript/src/client.ts
+++ b/clients/typescript/src/client.ts
@@ -225,12 +225,20 @@ export class Client {
     return new Consumer(this, queue, name, opts);
   }
 
-  /** Run `pgque.ticker()` once. Mostly useful for tests + manual rotation. */
-  async ticker(): Promise<void> {
+  /**
+   * Wrapper for pgque.ticker(): if `queue` is given, runs the per-queue
+   * overload (`pgque.ticker(queue text)`); otherwise runs the no-arg global
+   * overload. Returns the underlying tick result count.
+   */
+  async ticker(queue?: string): Promise<void> {
     try {
-      await this.pool.query('select pgque.ticker()');
+      if (queue !== undefined) {
+        await this.pool.query('select pgque.ticker($1)', [queue]);
+      } else {
+        await this.pool.query('select pgque.ticker()');
+      }
     } catch (err) {
-      throw mapPgError('ticker', err);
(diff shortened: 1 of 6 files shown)


NikolayS:
## REV review — PR #130 (head `1a70916`)

**Verdict: PASS (REQUEST_CHANGES_NON_BLOCKING)** — the diagnosis is correct, the patch is minimal and right-sized for v0.1, and there are no blockers. A few non-blocking notes below for follow-up.

---

### 1. Security
- `forceTick()` continues to use parametrized `$1` for the queue name; `pgque.ticker()` takes no args. No client-side string interpolation introduced. DSN handling is unchanged (`connect()` still passes through `pg.Pool({connectionString: dsn, ...})`). No new surface.
- `nack.test.ts` continues to use `$1` binding for `queue_name`. Clean.
- No secrets / DSNs leaked in the diff or PR body.

### 2. Bugs
- **Two-call sequence is correct.** `pgque.force_tick(queue)` only bumps `queue_event_seq` so the next `ticker()` cycle sees a non-empty tick; `pgque.ticker()` is what closes the current tick and rotates. The previous helper genuinely produced empty batches whenever `queue_ticker_max_count` / `queue_ticker_max_lag` were not crossed. Fix matches PgQ semantics in `blueprints/SPEC.md` (rotation mechanics).
- **Separate-connection rationale is partly off.** The commit text claims separate pool connections are needed "so the ticker's snapshot correctly sees the producing transaction's events." The actual reason events are visible is that `send()` has already returned, i.e. its COMMIT is durable before `forceTick()` is called; visibility is independent of which pool slot runs `force_tick` vs `ticker`. Two `pool.query()` calls on a default `pg.Pool` already use separate logical sessions per `query()`, but they aren't a guarantee — they happen to be acquired/released around each await. Non-blocking nit, but please relax the doc comment to avoid implying a snapshot/connection invariant the code doesn't actually enforce.
- **Race window between the two calls.** Yes, in production a concurrent `pgque.ticker()` (e.g. pg_cron) could rotate between `force_tick` and our `ticker()`. For test usage that's harmless — both calls converge on "tick is closed, new batch exists." Not worth advisory-locking in v0.1; tests don't run pg_cron, and adding a queue-scoped advisory lock around a test helper would couple it to internal lock IDs. Recommend documenting that `forceTick()` is **test-only** (the JSDoc already says "mostly useful for tests" — consider promoting to "intended for tests; do not call from production code paths").
- **`fileParallelism: false` is correct, not masking a driver bug.** The race is in test fixtures (`setupTestQueue` / `teardownTestQueue` in `helpers.ts` create+drop randomly-named queues, but each test file opens its own `Client` against the same DB). The driver itself doesn't share mutable state across connections beyond pg-types parser mutation, which is set-once. Cross-file serialization is a property of the test harness, not the SUT. Within-file `concurrent: false` is still the vitest default, so the existing intra-file ordering is preserved. Each file already uses `randomSuffix()` for queue names, so in principle parallelism could be reinstated later if `pgque.drop_queue` and `pgque.create_queue` are confirmed safe under contention; but for v0.1 this is fine.

### 3. Test analysis
- `nack honors custom retryAfter and reason` previously asserted on a **non-existent** column; the test was passing only because `expect(retry.rows.length).toBe(1)` is satisfied by any row, and the broken `select` would have actually thrown… meaning the test was failing, not green-for-the-wrong-reason. Post-fix it asserts what its name claims (a row landed in `retry_queue`). Note it does **not** actually assert that the value of `ev_retry_after` is ~5s in the future, even though the test name says "honors custom retryAfter". Consider tightening to `expect(retry.rows[0]!.ev_retry_after.getTime() - Date.now()).toBeGreaterThan(3000)` in a follow-up — non-blocking.
- `first nack lands in retry_queue` and `nack at retry limit routes to dead_letter` assert what their names claim, and they will now actually run end-to-end (previously they short-circuited at the empty `receive()` from the broken `forceTick`).
- The fix doesn't break prior signal: every test that depended on the helper was effectively a no-op against the SUT before; now they exercise it.

### 4. Guidelines (CLAUDE.md)
- Conventional Commit subject `fix(clients/typescript): make forceTick rotate, fix nack test schema, serialise` — well within 50–72 chars, present tense, scoped. Good.
- No mention of internal hosts, GitLab, sahmed, AR, round-N, WI artifacts. Clean.
- TypeScript-only change, no SQL touched, so SECURITY DEFINER / search_path rules are N/A.
- TDD ordering rule (red/green) is satisfied by the existing tests now actually exercising the driver.

### 5. Docs
- JSDoc on `forceTick()` is updated and accurately describes both calls; `ticker()`'s own JSDoc is unchanged and consistent.
- `clients/typescript/README.md` does not mention `forceTick` at all (the API table omits it as a test-only helper), so no doc drift to fix.
- Repo-wide search for `ev_retry_after_ts` returns only the pre-fix blob — no remaining stale references in source, tests, or docs.

---

### Non-blocking suggestions
1. Soften the JSDoc claim about "separate pool connection ensures snapshot visibility" — visibility is established by `send()`'s commit, not by connection separation.
2. Tighten the "honors custom retryAfter" assertion to actually verify the timestamp is in the future by ~the requested interval.
3. Consider marking `forceTick()` `@internal` or `@experimental` in the next minor; production code should rely on pg_cron or `ticker()` directly.

**Blocker count: 0.**

— REV reviewer (multi-perspective)

---

NikolayS:
Design change in this revision:

`forceTick(queue)` is now a 1:1 wrapper for `pgque.force_tick(queue)` — threshold bump only, single query, no rotation side effect.

A new `flush(queue)` method handles the test pattern of "send → tick → receive": it calls `pgque.force_tick($1)` then `pgque.ticker($1)` (the per-queue overload, not the global one). All test call sites that need a batch to be produced before `receive()` have been updated to `flush()`.

The `ev_retry_after_ts` → `ev_retry_after` schema fix and `fileParallelism: false` are unchanged.

---

NikolayS:
Applied the design feedback: `Client.flush()` is removed. `ticker()` now accepts an optional `queue` argument (no-arg path unchanged, so existing callers are unaffected). `forceTick()` remains a plain wrapper. The `advanceQueue(client, queue)` composition is now in `test/helpers.ts` — test layer only, not public API. All test call sites updated. Commit: `9ad7fb7`.

---

PR #132, opened 2026-04-30
fix(clients/go): green up live-pg test suite

NikolayS:
## Summary

The Go client test suite added in #81 fails against a live Postgres because three test-side assumptions don't hold against the actual SQL engine. None of the failures are driver bugs; the driver code is unchanged.

Three commits, three independent root causes:

- `0c3edca` — `tick` helper called only `pgque.ticker()`. Freshly-created queues use the default thresholds (`queue_ticker_max_count=500`, `queue_ticker_max_lag=3s`), so ticker doesn't fire on small unit-test event counts. Helper now calls `pgque.force_tick(queue)` first; signature gains a `queue` parameter and call sites cascade.
- `2b9270e` — `TestEvent_PayloadIsArray` asserted `"[1,2,3]"`. JSONB canonicalises array text with a space after each comma, so the round-trip is `"[1, 2, 3]"`.
- `3171a08` — `TestRace_ConcurrentSend` sent 200 events in one tick window but called Receive with `i_max_return=100`. `pgque.receive` yields up to the cap and Ack finishes the whole batch, dropping the remaining 100. Sized the cap above the total.

After these fixes the full live-PG suite (`go test -race -v ./...` against PG 16) is green, including the consumer/concurrency/edge tests.

This PR targets `main` directly, not `ci/live-pg-driver-tests` (#84) — same model used for #129 (Python) and #130 (TypeScript). Once merged, #84 should rebase and the Go driver job should turn green.

## Test plan

- [x] `go test -race -v ./...` against a live PG 16 with `pgque.sql` installed: all PASS or SKIP (no FAIL)
- [ ] CI `Go client tests` job on PR #84 turns green after rebase

---

### DIFF
--- a/clients/go/bench_test.go
+++ b/clients/go/bench_test.go
@@ -39,7 +39,7 @@ func BenchmarkReceive_Empty(b *testing.B) {
 	queue, consumer := benchSetup(b, client)
 	ctx := context.Background()
 
-	if _, err := client.Pool().Exec(ctx, "select pgque.ticker()"); err != nil {
+	if _, err := client.Pool().Exec(ctx, "select pgque.ticker($1)", queue); err != nil {
 		b.Fatal(err)
 	}
 
@@ -66,7 +66,7 @@ func BenchmarkSendReceiveAck(b *testing.B) {
 		}); err != nil {
 			b.Fatal(err)
 		}
-		if _, err := client.Pool().Exec(ctx, "select pgque.ticker()"); err != nil {
+		if _, err := client.Pool().Exec(ctx, "select pgque.ticker($1)", queue); err != nil {
 			b.Fatal(err)
 		}
 		msgs, err := client.Receive(ctx, queue, consumer, 100)
@@ -97,7 +97,7 @@ func BenchmarkConsumer_DispatchThroughput(b *testing.B) {
 			b.Fatal(err)
 		}
 	}
-	if _, err := client.Pool().Exec(ctx, "select pgque.ticker()"); err != nil {
+	if _, err := client.Pool().Exec(ctx, "select pgque.ticker($1)", queue); err != nil {
 		b.Fatal(err)
 	}
 
--- a/clients/go/concurrency_test.go
+++ b/clients/go/concurrency_test.go
@@ -42,7 +42,7 @@ func TestRace_ConcurrentSend(t *testing.T) {
 		}(g)
 	}
 	wg.Wait()
-	tick(t, client)
+	tick(t, client, queue)
 
 	total := 0
 	for {
@@ -161,7 +161,7 @@ func TestRace_HandlerNackUnderLoad(t *testing.T) {
 			t.Fatal(err)
 		}
 	}
-	tick(t, client)
+	tick(t, client, queue)
 
 	rng := rand.New(rand.NewSource(1))
 	var rngMu sync.Mutex
@@ -217,7 +217,7 @@ func TestConcurrent_TwoConsumersDistinctNames(t *testing.T) {
 			t.Fatal(err)
 		}
 	}
-	tick(t, client)
+	tick(t, client, queue)
 
 	var countA, countB int64
 
--- a/clients/go/concurrency_test.go
+++ b/clients/go/concurrency_test.go
@@ -117,7 +117,7 @@ func TestRace_SendReceiveLoop(t *testing.T) {
 				return
 			default:
 			}
-			if _, err := client.Pool().Exec(ctx, "select pgque.ticker()"); err != nil {
+			if _, err := client.Pool().Exec(ctx, "select pgque.ticker($1)", queue); err != nil {
 				if ctx.Err() != nil {
 					return
 				}
--- a/clients/go/concurrency_test.go
+++ b/clients/go/concurrency_test.go
@@ -44,9 +44,14 @@ func TestRace_ConcurrentSend(t *testing.T) {
 	wg.Wait()
 	tick(t, client, queue)
 
+	expected := goroutines * perGoroutine
+	// pgque.receive truncates yielded rows at i_max_return, but Ack
+	// finishes the entire batch — events past the cap are not yielded
+	// again without a fresh tick. Size the cap above the total so the
+	// whole tick window flows out in one call.
 	total := 0
 	for {
-		msgs, err := client.Receive(ctx, queue, consumer, 100)
+		msgs, err := client.Receive(ctx, queue, consumer, 2*expected)
 		if err != nil {
 			t.Fatal(err)
 		}
@@ -58,7 +63,6 @@ func TestRace_ConcurrentSend(t *testing.T) {
 			t.Fatal(err)
 		}
 	}
-	expected := goroutines * perGoroutine
 	if total != expected {
 		t.Fatalf("expected %d messages received, got %d", expected, total)
 	}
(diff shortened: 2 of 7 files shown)


NikolayS:
## REV review — PR #132

**Verdict: PASS (non-blocking follow-up noted)**

Test-only change, three independent root causes all correctly diagnosed and fixed without touching driver code. Reviewed against the 6 REV perspectives (SOC2 skipped per scope).

### 1. Security
N/A — test code only, no SECURITY DEFINER surface, no credentials, no SQL injection vectors (queue name flows through a parameterized `$1`).

### 2. Bugs / root cause
All three fixes target the actual root cause:

- **`force_tick` + `ticker`** — Correct. Default `queue_ticker_max_count=500` and `queue_ticker_max_lag=3s` mean the global `pgque.ticker()` is a no-op for unit-test event volumes. `force_tick(queue)` flips the per-queue flag so the next ticker pass actually rotates. Calling both (force then global) is the right sequence; reordering would not fire.
- **`"[1, 2, 3]"`** — Not a hidden round-trip bug. JSONB stores as a parsed tree and emits canonical text with a space after each comma when cast back via `::text`. The driver is a passive byte conduit here; asserting against PG's canonical form is correct. (If the test wanted to assert "exactly the bytes I sent," it should `json_build_array(...)` or compare structurally — that would be a separate test, not a fix to this one.)
- **Receive cap `2*expected`** — Correct read of `pgque.receive` semantics: `i_max_return` truncates the yielded rowset but the batch covers the full tick window, and Ack closes the batch — leftover events are not re-yielded without another rotation. The fix is sound. Minor nit: a comment-only sentence about "another tick would also drain it" might help future readers, but the inline comment already added is sufficient.

No fix masks a real driver issue.

### 3. Test analysis
Tests still assert what their names claim. `TestRace_ConcurrentSend` still validates concurrent Send correctness (200 rows in, 200 rows out, in-order batch); raising the cap doesn't weaken the assertion since the post-loop `total != expected` check is preserved. `TestEvent_PayloadIsArray` still validates array round-trip — it now asserts the canonical text form, which is the actual contract. No test passes for the wrong reason.

### 4. Guidelines
- Conventional commits: `fix(clients/go): ...` — compliant.
- No person names, no GitLab refs, no internal infra/Round/WI references in diff or PR body.
- Helper comment in `helpers_test.go` is clear and pins the rationale (default thresholds) — good documentation hygiene.

### 5. Docs
N/A.

### 6. Architecture (the critical one)
1:1 SQL-API discipline preserved:
- No changes to `clients/go/pgque.go`, `consumer.go`, or any driver source — confirmed by file list (5 files, all `*_test.go`).
- `tick()` remains a package-private test helper (lowercase, in `helpers_test.go`); it does not leak into the public `Client` surface as a composite "send-and-tick" method. This is the right boundary — production code should not be calling a force_tick path.
- Signature change `tick(t, client)` → `tick(t, client, queue)` is purely internal to the test binary; cascading updates across 4 files are mechanical and complete.

No architectural drift.

### Follow-up (non-blocking, separate issue)
`TestNack_ToDLQAtRetryLimit` self-skips with `t.Skipf("set_queue_max_retries not available: %v", err)` because it calls `pgque.set_queue_max_retries($1, 2)` — that function name does not appear to exist in v0.1; the actual API is `pgque.set_queue_config(...)`. The test currently green-skips silently against the real install, meaning DLQ-at-retry-limit is **not actually exercised** by the live-PG suite. Worth filing a separate issue to (a) switch the test to `pgque.set_queue_config` (or the correct setter), and (b) downgrade the `t.Skipf` to `t.Fatalf` so a missing setter fails loudly rather than silently. Out of scope for this PR — recommend merging as-is.

### Summary
Three test-only fixes, all correctly root-caused, no driver behavior change, no public-API leak, no guideline violations. Ship it.


---

NikolayS:
**Fix for `TestNack_ToDLQAtRetryLimit`** — commit f4fe706 on this branch.

**Changes in `clients/go/integration_test.go`:**

1. Replaced `pgque.set_queue_max_retries($1, 2)` with `pgque.set_queue_config($1, 'max_retries', '2')` — the correct v0.1 API. Note: `set_queue_config` prepends `queue_` internally, so the param is `'max_retries'` not `'queue_max_retries'`.

2. Replaced `t.Skipf` on the old error path with `t.Fatalf` so the test never silently skips on a bad SQL call.

**Two real driver issues exposed by running the test end-to-end:**

- `nack()` routes individual events to `retry_queue`/`dead_letter` but does not call `finish_batch`. The batch must be explicitly closed with `Ack` after all `Nack` calls in a cycle, or PgQ cannot advance the consumer cursor. Fixed in the test loop.
- The default 60-second retry delay in `nack()` causes `maint_retry_events()` to find nothing until the delay expires. The test now expires `retry_after` times inline before each `maint_retry_events()` call so cycles run immediately without sleeping.

**Local result:** `PASS` (`go test -race -v -run TestNack_ToDLQAtRetryLimit ./clients/go/`)


---

NikolayS:
## REV re-review @ `f4fe706` (delta-only)

**Verdict: PASS (no blockers, 1 non-blocking note)**

Re-reviewed only the new commit `f4fe706` ("fix(test): use set_queue_config for max_retries"). Prior 3-commit set was already PASS; this round just confirms the new commit doesn't regress that.

### Bugs / test correctness — clean

1. **`set_queue_config` parameter name is correct.** Verified against `sql/pgque.sql` lines 3967–4006: the function prepends `queue_` to the param name (`v_param_name := 'queue_' || x_param_name`) and whitelists `queue_max_retries`. So `set_queue_config($1, 'max_retries', '2')` resolves to `update pgque.queue set queue_max_retries = '2'…`. Correct.
2. **DLQ assertion path is sound.** `dlqCount` (helpers_test.go) joins `pgque.dead_letter dl` on `q.queue_id = dl.dl_queue_id` and asserts `> 0`. The only code path that writes to `pgque.dead_letter` is `pgque.event_dead`, called from the nack handler when `ev_retry >= queue_max_retries`. Test cannot pass for the wrong reason — there is no alternate path to populate that table.
3. **Ack-after-Nack matches PgQ batch model.** `pgque.nack` routes individual events to retry_queue/dead_letter but does not finish the batch; cursor advance still requires `finish_batch` (called by `Ack`). The added `client.Ack(ctx, batchID)` after the per-event nack loop is the canonical pattern documented in `docs/reference.md` for `pgque.nack`.
4. **`Skipf → Fatalf`** correctly closes the silent-skip hole. With the bogus function name fixed, there is no remaining reason to skip-on-error here.

### Architecture — clean

- Diff confirmed test-only: 5 files (`concurrency_test.go`, `consumer_test.go`, `edge_test.go`, `helpers_test.go`, `integration_test.go`). **No `client.go` change** — the 60-second hardcoded `retry_after` mentioned in the commit body did NOT leak into a public API change. Correctly deferred to a follow-up Go-vs-Python alignment PR.

### Style / hygiene — clean

- Conventional commit subject: `fix(test): use set_queue_config for max_retries` ✓
- No emojis, no anti-leak issues, no SECURITY DEFINER touched, no SQL style regressions.

### Non-blocking note

The inline `update pgque.retry_queue set ev_retry_after = now() - interval '1 second'` has **no queue filter** — it expires retry delays for *every* queue in the database. Fine for the dedicated `pgque_test` DB, but if this test ever runs against a shared DB it would interfere with concurrent tests. Cheap fix for a future polish: add `where ev_queue = (select queue_id from pgque.queue where queue_name = $1)`. Not blocking — the test DB convention is established.

LGTM, ship it.

— REV

---

NikolayS:
Switched the `tick()` helper in `clients/go/helpers_test.go` to the per-queue `pgque.ticker($1)` overload (matches the fix from #130). Avoids cross-test side effects when test queues run in parallel — only the queue under test is ticked. Comment tightened to reflect the new behavior. Pushed as `edc7243`.

---

Issue #134, opened 2026-04-30
Clients: high-level consumers can skip unreturned batch rows when max_messages < batch size


NikolayS:
## Problem

All three high-level consumers can silently skip/drop messages for a consumer when the underlying PgQ batch is larger than the client-side `max_messages`/`maxMessages` limit.

`pgque.receive(queue, consumer, max_return)` limits only the rows returned to the client. `pgque.ack(batch_id)` finishes the **entire underlying PgQ batch**, including rows that were not returned because of `max_return`.

Current drivers use unsafe defaults/patterns:

- Python: `Consumer(..., max_messages=100)` then `ack(batch_id)` after processing returned rows (`clients/python/pgque/consumer.py`).
- TypeScript: `ConsumerOptions.maxMessages` defaults to `100`, then `ack(batchId)` (`clients/typescript/src/consumer.ts`).
- Go: `Consumer.Start()` hardcodes `Receive(..., 100)` and then `Ack(batchID)`; there is no `WithMaxMessages` option (`clients/go/consumer.go`).

Default PgQue `queue_ticker_max_count` is 500, so a normal default batch can exceed 100.

## Repro evidence

Fresh DB, `sql/pgque.sql`, direct SQL:

```sql
select pgque.create_queue('q');
select pgque.register_consumer('q','c');
select pgque.send('q', 'default', jsonb_build_object('i', g))
from generate_series(1, 105) g;
select pgque.force_tick('q');
select pgque.ticker('q');

create temp table got as select * from pgque.receive('q','c',100);
select count(*) as returned, min(msg_id), max(msg_id), count(distinct batch_id) from got;
-- returned=100, min=1, max=100, batches=1

select pgque.ack((select batch_id from got limit 1));
select count(*) as returned_after_ack from pgque.receive('q','c',100);
-- returned_after_ack=0
```

The 5 unreturned rows are no longer visible for that consumer after ack.

## Why this matters

This violates the mental model of `max_messages`: users expect it to cap one poll cycle, not to make unreturned rows disappear after the driver auto-acks.

Existing tests encode the dangerous pattern in places, e.g. “receive at most N” tests ack a partial batch without asserting the remainder is preserved.

## Suggested fix options

Pick one policy and align all drivers/docs/tests:

1. Safe default: high-level consumers request at least the queue's `ticker_max_count` by default (500), or expose a `full_batch`/safe receive helper.
[...]

Issue #142, opened 2026-04-30
Go client: Consumer docs describe the opposite handler/error semantics


NikolayS:
## Problem

Go Consumer public comments currently describe behavior that is not what the code does.

Examples in `clients/go/consumer.go`:

- `HandlerFunc` comment says returning an error causes the entire batch to be **not acked** and redelivered.
- `Handle` comment says messages with no registered handler are silently skipped.
- `Start` comment says a batch is acked only if every handled message returned nil.

Actual behavior:

- Handler errors are caught, the individual message is `Nack()`ed, and the batch is still `Ack()`ed later.
- Missing handlers are logged and `Nack()`ed.
- Batch `Ack()` is attempted even when handler errors occurred; see separate issue about acking after failed nack.

## Why this matters

This is a dangerous contract mismatch. Users may design handlers expecting whole-batch redelivery on error, while the driver actually performs per-message retry/DLQ plus final batch ack.

## Suggested fix

After deciding the cross-driver unknown-handler and nack-failure policies, update Go comments and README to match exactly.

Also add tests that encode the intended Consumer contract:

- handler success -> ack
- handler error -> nack + ack only if nack succeeded
- missing handler -> chosen policy
- panic -> same as handler error

## Acceptance criteria

- Go doc comments match runtime behavior.
- `go test` includes contract tests for the documented behavior.

NikolayS:
Closing as completed. Doc comments on `HandlerFunc`, `Consumer.Handle`, `Consumer.Start` rewritten to match runtime behavior in #153 / #156 (merged).

---
_Generated by [Claude Code](https://claude.ai/code)_

Issue #138, opened 2026-04-30
Go client: add missing thin wrappers/options for parity and ease of use


NikolayS:
## Problem

The Go client lags Python/TypeScript and SQL API ergonomics in several places, making examples/tests fall back to raw SQL and making the three drivers feel like different products.

## Gaps

Go currently lacks wrappers/options for:

- `send_batch` (Python has it; SQL has it)
- `subscribe` / `unsubscribe` (TypeScript has these; SQL has wrappers)
- `ticker(queue?)` / `force_tick(queue)` (TypeScript has these; tests use raw SQL)
- `Nack` options: retry delay and DLQ reason (Python and TypeScript expose these; Go hardcodes `'60 seconds'::interval, null`)
- `Consumer` max messages option (`Consumer.Start()` hardcodes 100)

## Why this matters

- Ease of use: users have to escape to `Pool().Exec(...)` for common PgQue operations.
- Alignment: Python/Go/TS examples diverge unnecessarily.
- “Don't reinvent the wheel”: drivers should be thin, consistent wrappers over SQL primitives, not partial APIs with ad-hoc raw-SQL holes.

## Suggested fix

Add thin wrappers that map directly to SQL primitives:

- `SendBatch(ctx, queue, type string, payloads []any) ([]int64, error)` or an explicit JSON/text split.
- `Subscribe(ctx, queue, consumer string) (int, error)` / `Unsubscribe(...)`.
- `Ticker(ctx, queue ...string) error` and `ForceTick(ctx, queue string) error`.
- `Nack(ctx, batchID, msg, opts...)` with `WithRetryAfter`, `WithReason`.
- `WithMaxMessages(n int)` for high-level Consumer, after resolving the safe default in the batch-size issue.

## Acceptance criteria

- Go README can show a normal send/tick/receive/ack flow without raw SQL.
- Go has parity with TypeScript/Python for nack retry delay/reason.
- Tests cover each wrapper as a direct SQL primitive wrapper, not a semantic reinvention.

Issue #136, opened 2026-04-30
Clients: align unknown event type policy across Python, Go, and TypeScript consumers


NikolayS:
## Problem

The three high-level Consumers disagree on what happens when a message has no registered handler.

Current behavior:

- Python: logs warning and **acks** the message/batch by default.
- Go: logs and **nacks** the message, then acks the batch.
- TypeScript: logs and **nacks** the message, then acks the batch.

This is a material API contract difference. The same application ported between drivers can either drop unknown event types or send them to retry/DLQ.

## Why this matters

Unknown event type handling is a data-safety policy, not cosmetic behavior.

Ack-by-default is convenient for handler-as-filter semantics, but it can silently discard messages after deploy/config mistakes. Nack-by-default is safer for data retention, but can poison retry/DLQ if event types are intentionally filtered.

## Suggested decision

Pick one cross-driver policy and document it prominently.

My recommendation: default to **nack unknown event types** unless the user explicitly registers a catch-all/default handler that chooses to ack/ignore. If handler-as-filter semantics are desired, make that an explicit option such as:

- Python: `Consumer(..., unknown_handler="ack"|"nack")`
- Go: `WithUnknownHandlerPolicy(AckUnknown|NackUnknown)`
- TypeScript: `unknownHandlerPolicy?: 'ack' | 'nack'`

## Acceptance criteria

- Python, Go, and TypeScript have the same default policy.
- All three READMEs and docstrings describe the policy.
- Tests cover unknown event type behavior in all three drivers.

PR #153, opened 2026-04-30
fix(clients/go): nack-fail handling, unknown-type policy, options, SendBatch

NikolayS:
## Summary

Tightens the Go client's at-least-once contract and rounds out the API surface to match the cross-driver decisions for v0.2.0.

- **Nack-fail data-loss fix.** Previously, a failing per-message `Nack` was logged and the batch was Ack'd anyway — the failure was silently dropped, the consumer cursor advanced past it, and PgQue had no way to redeliver. The Consumer now tracks per-poll Nack failures and skips the batch `Ack` if any required `Nack` errored, letting PgQue redeliver on the next `Receive`. A red/green test using a stub backend that always errors on `Nack` proves `Ack` is no longer called in that path.
- **`WithUnknownHandlerPolicy(p)`** with `NackUnknown` (default, current behavior) and `AckUnknown` (silently skip messages whose Type has no registered handler). The default keeps the cross-driver contract — a producer/consumer mismatch never silently drops messages.
- **`WithMaxMessages(n)`** ConsumerOption. Default `WithMaxMessages(500)` matches `ticker_max_count` so common batches drain in one Receive. This is a mitigation, not a fix — see #134 (still open) for the SQL-side root cause: `pgque.ack(batch_id)` finishes the entire underlying PgQ batch, so any rows beyond `maxMessages` (e.g. `ticker_max_lag` bursts or operator changes to `ticker_max_count`) are silently skipped after ack.
- **`Client.SendBatch(ctx, queue, type, payloads)`** — thin 1:1 wrapper for `pgque.send_batch(text, text, jsonb[])`. Each payload is JSON-marshalled.
- **`Client.Nack` is now variadic**: `WithRetryAfter(time.Duration)` and `WithReason(string)`. Calls with no options preserve the existing `60s`, NULL-reason defaults — fully backward compatible.
- **Doc comments** on `HandlerFunc`, `Consumer.Handle`, and `Consumer.Start` rewritten to match runtime behavior: per-message Nack on handler error or unknown type, batch Ack only when every required Nack succeeded.

Closes the Go portion of #135, #136, #142, and the in-scope items of #138 (SendBatch + Nack options). Mitigates #134 (Go portion) — the underlying ack-drops-unreturned-rows behavior is SQL-side and remains open. Subscribe / Unsubscribe / Ticker wrappers remain deferred to v0.2.1.

## Test plan

- [x] `go vet ./...` clean
- [x] `gofmt -l` clean
- [x] `go test -race -v ./...` against PG 16 with `\i sql/pgque.sql`: full suite passes (49 tests, 0 failures, 1 pre-existing skip)
- [x] New tests:
  - `TestConsumer_NackFailure_DoesNotAck` — red/green for the data-loss fix
  - `TestConsumer_NackSuccess_StillAcks` — green-path control
  - `TestConsumer_AckUnknownPolicy_SkipsNack` — policy override
  - `TestConsumer_WithMaxMessages_FetchesEntireBatch` — 105-msg batch via `WithMaxMessages(500)`
  - `TestSendBatch_RoundTrip`, `TestSendBatch_EmptySlice`
  - `TestNack_WithRetryAfter` — verifies the option threads to `pgque.nack`
- [ ] CI green on PG 14/15/16/17/18

## Backward compatibility

- `Option` is preserved as a type alias for the new `ConsumerOption`.
- `Client.Nack(ctx, batchID, msg)` keeps the same default behavior; new options are variadic and additive.
- `Client.NewConsumer` still takes the same `...Option` pattern.

## Out of scope (deferred to v0.2.1)

- `Subscribe` / `Unsubscribe` wrappers
- `Ticker` / `ForceTick` wrappers (driver-level — tests use these via `Pool().Exec`)

---

### DIFF
--- a/clients/go/README.md
+++ b/clients/go/README.md
@@ -38,7 +38,7 @@ func main() {
     //   select pgque.create_queue('orders');
     //   select pgque.register_consumer('orders', 'order_worker');
 
-    // Producer side
+    // Producer side -- single event
     _, err = client.Send(ctx, "orders", pgque.Event{
         Type:    "order.created",
         Payload: map[string]any{"order_id": 42},
@@ -47,6 +47,7 @@ func main() {
         log.Fatal(err)
     }
 
+    // Producer side -- batch (one type, many payloads)
     ids, err := client.SendBatch(ctx, "orders", "order.created", []any{
         map[string]any{"order_id": 43},
         map[string]any{"order_id": 44},
@@ -57,7 +58,10 @@ func main() {
     log.Printf("published batch event IDs: %v", ids)
 
     // Consumer side
-    consumer := client.NewConsumer("orders", "order_worker")
+    consumer := client.NewConsumer("orders", "order_worker",
+        pgque.WithMaxMessages(500),                            // ticker_max_count default
+        pgque.WithUnknownHandlerPolicy(pgque.NackUnknown),     // also the default
+    )
     consumer.Handle("order.created", func(ctx context.Context, msg pgque.Message) error {
         log.Printf("got %s: %s", msg.Type, msg.Payload)
         return nil
@@ -67,6 +71,35 @@ func main() {
     }
 }
 ```
+
+## Consumer options
+
+| Option                                  | Default        | Notes                                                                 |
+| --------------------------------------- | -------------- | --------------------------------------------------------------------- |
+| `WithPollInterval(d time.Duration)`     | `30s`          | Idle backoff between polls when the queue is empty.                   |
+| `WithMaxMessages(n int)`                | `500`          | Per-Receive batch cap. Match `ticker_max_count` for full throughput.  |
+| `WithUnknownHandlerPolicy(p)`           | `NackUnknown`  | `AckUnknown` silently skips messages with no registered handler.      |
+
+## Nack options
+
+`Client.Nack` takes optional, variadic `NackOption`s:
+
+```go
+err := client.Nack(ctx, batchID, msg,
+    pgque.WithRetryAfter(5*time.Minute), // override 60s default
+    pgque.WithReason("payment-declined"), // recorded on the dead_letter row
+)
+```
+
+Calls without options preserve the historical defaults: 60-second retry
+delay, NULL reason.
+
+## At-least-once contract
+
+If a per-message Nack call fails, the Consumer leaves the batch unacked
+so PgQue redelivers it on the next Receive. Acking a batch whose Nack
+failed would silently drop the failure information — the Go consumer
+prefers redelivery and lets the at-least-once retry path do its job.
 
 ## Tests
 
--- a/clients/go/README.md
+++ b/clients/go/README.md
@@ -77,7 +77,7 @@ func main() {
 | Option                                  | Default        | Notes                                                                 |
 | --------------------------------------- | -------------- | --------------------------------------------------------------------- |
 | `WithPollInterval(d time.Duration)`     | `30s`          | Idle backoff between polls when the queue is empty.                   |
-| `WithMaxMessages(n int)`                | `500`          | Per-Receive batch cap. Match `ticker_max_count` for full throughput.  |
+| `WithMaxMessages(n int)`                | `500`          | Per-Receive limit; matches default `ticker_max_count`. If a batch exceeds this, `Ack` still finishes it and unreturned rows are skipped — size `>=` the queue's `ticker_max_count`. |
 | `WithUnknownHandlerPolicy(p)`           | `NackUnknown`  | `AckUnknown` silently skips messages with no registered handler.      |
 
 ## Nack options
--- a/clients/go/README.md
+++ b/clients/go/README.md
@@ -59,8 +59,7 @@ func main() {
 
     // Consumer side
     consumer := client.NewConsumer("orders", "order_worker",
-        pgque.WithMaxMessages(500),                            // ticker_max_count default
-        pgque.WithUnknownHandlerPolicy(pgque.NackUnknown),     // also the default
+        pgque.WithUnknownHandlerPolicy(pgque.NackUnknown), // also the default
     )
     consumer.Handle("order.created", func(ctx context.Context, msg pgque.Message) error {
         log.Printf("got %s: %s", msg.Type, msg.Payload)
@@ -77,8 +76,8 @@ func main() {
 | Option                                  | Default        | Notes                                                                 |
 | --------------------------------------- | -------------- | --------------------------------------------------------------------- |
 | `WithPollInterval(d time.Duration)`     | `30s`          | Idle backoff between polls when the queue is empty.                   |
-| `WithMaxMessages(n int)`                | `500`          | Per-Receive limit; matches default `ticker_max_count`. If a batch exceeds this, `Ack` still finishes it and unreturned rows are skipped — size `>=` the queue's `ticker_max_count`. |
-| `WithUnknownHandlerPolicy(p)`           | `NackUnknown`  | `AckUnknown` silently skips messages with no registered handler.      |
+| `WithMaxMessages(n int)`                | `math.MaxInt32` | Per-Receive limit. The default requests the whole PgQ batch before `Ack`. If you lower it below the real batch size, `Ack` still finishes the batch and unreturned rows are skipped. |
+| `WithUnknownHandlerPolicy(p)`           | `NackUnknown`  | `AckUnknown` logs and skips messages with no registered handler.      |
 
 ## Nack options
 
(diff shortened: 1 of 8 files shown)


NikolayS:
## REV review — PR #153 (`fix/clients-go-v0.2.0` @ `57f9573`)

**Verdict: REQUEST_CHANGES_NON_BLOCKING** — one weak test, one minor doc/clarity nit. Code is correct and ships safely; the two items below would make the test suite carry its own weight.

---

### 1. Security
- `WithReason` flows through `$13` as a bound parameter (`reason` is an untyped `any`, pgx binds as `text` or `NULL`); `WithRetryAfter` flows through `$12::interval`. No string concatenation, no SQL injection vector. The `pgque.nack` SQL function additionally re-queries the canonical event from `get_batch_events()` and ignores caller-supplied composite fields (#98 hardening), which is a strong defense in depth for the `Nack(ctx, batchID, msg, ...)` path. **PASS.**

### 2. Bugs
- `nackFailed` is a per-poll local in `Consumer.Start`, set to `true` on any `Nack` error in the loop, checked after the loop, never reset. Skips `Ack` for **any** nack failure in the batch — not just the most recent. Correct.
- `WithMaxMessages(0)` / negative is silently ignored (`if n > 0 { c.maxMessages = n }`), keeping the 500 default. Defensive, fine — though arguably should `panic` or `log` to surface caller mistakes; not blocking.
- `SendBatch(ctx, q, t, nil)`: `make([]string, 0)` → empty `text[]` → `$3::jsonb[]` cast → SQL `foreach` iterates zero times → `bigint[] '{}'`. Round-trips clean. Test `TestSendBatch_EmptySlice` covers this. ✓
- **Overload note (non-blocking):** `SendBatch` binds `encoded []string`, then casts `$3::jsonb[]`. PG resolves to the `jsonb[]` overload because of the explicit cast. If the cast were ever dropped, the call would silently bind to `pgque.send_batch(text, text, text[])` — the fast path. Not a current bug, but worth a one-line code comment so future refactors don't drop the cast.
- `consumerBackend` is satisfied by `*Client` (variadic `...NackOption` matches). Race-free under `-race` (per author's local run).

### 3. Test analysis
- ✅ `TestConsumer_NackFailure_DoesNotAck` — stub records `ackCount`; asserts `ackCount == 0`. Genuinely proves Ack is **not** called when Nack errors. Strong red/green guard.
- ✅ `TestConsumer_NackSuccess_StillAcks` — green-path control. Good.
- ✅ `TestConsumer_AckUnknownPolicy_SkipsNack` — asserts `nackCount == 0` and `ackCount > 0`. Good.
- ⚠️ **`TestConsumer_WithMaxMessages_FetchesEntireBatch` passes for partially the wrong reason.** It sends 105 msgs and waits up to 5s with a 50ms poll interval → ~100 poll cycles. If `WithMaxMessages(500)` were broken and the Consumer reverted to e.g. 100 per Receive, the test would **still pass** because subsequent polls would mop up the remaining 5 in the same tick window. The test proves "no silent drop," not "fetched in a single Receive." Suggested fix: instrument the stub-backend path to count `Receive` calls (mirror the `consumer_nackfail_test.go` pattern), assert `receiveCalls == 1` AND `seen == 105`. Or replace the live-DB version with a stub that records the `maxMessages` argument it was called with. **Non-blocking** — the integration coverage is still useful, just rename to `_DeliversAllMessages` to set honest expectations.

### 4. Guidelines
- Conventional commit title `fix(clients/go): ...` ✓
- Copyright headers correct on new files ✓
- No PR/issue numbers leak into runtime source files; test files reference issues (#98, #104, #135) which is acceptable as test rationale.
- `gofmt`/`go vet` clean per author's local run.

### 5. Docs
- README example typechecks: `WithMaxMessages`, `WithUnknownHandlerPolicy`, `NackUnknown`, `SendBatch(ctx, queue, type, []any{...})`, `Nack(ctx, batchID, msg, WithRetryAfter(...), WithReason(...))` all match exported signatures. ✓
- `HandlerFunc`, `Consumer.Handle`, `Consumer.Start` doc comments accurately describe per-message Nack + batch-Ack-only-if-all-Nacks-succeeded behavior post-#142. The README's "At-least-once contract" section is consistent with the runtime code. ✓
- Minor: README's "Consumer options" table lists `WithPollInterval` default as `30s` ✓ but doesn't document that `WithMaxMessages(n)` with `n <= 0` is silently ignored. Worth a one-line footnote.

### 6. Architecture (CLAUDE.md rule 7 — 1:1 SQL wrappers)
| Client method | SQL function | 1:1? |
|---|---|---|
| `Send` | `pgque.send(text, text, jsonb)` | ✓ |
| `SendBatch` | `pgque.send_batch(text, text, jsonb[])` | ✓ |
| `Receive` | `pgque.receive(text, text, int)` | ✓ |
| `Ack` | `pgque.ack(bigint)` | ✓ |
| `Nack(...opts)` | `pgque.nack(bigint, message, interval, text)` | ✓ — variadic options configure existing 1:1 args |

No new composite `Client` method introduced. `Consumer` is the API-layer abstraction above and is permitted to call multiple SQL functions.

**`consumerBackend` test hook:** unexported interface in `consumer.go`; the `ConsumerBackend` type alias and `SetConsumerBackend` setter live in `export_test.go`, which the Go toolchain only compiles in test builds. Production binaries never see these symbols. **Does not leak public API surface.** ✓

---

### Blockers
None.

### Non-blocking change requests
1. **Tighten `TestConsumer_WithMaxMessages_FetchesEntireBatch`**: assert exactly one `Receive` call with `maxMessages == 500` (use the stub-backend pattern). As written it cannot distinguish broken plumbing from correct plumbing.
2. Add a one-line comment on the `$3::jsonb[]` cast in `Client.SendBatch` explaining it pins overload resolution to the JSON path.
3. README: footnote `WithMaxMessages(n)` ignoring `n <= 0`.

Nice work on the data-loss fix and the at-least-once doc clarification. The stub-backend pattern in `consumer_nackfail_test.go` is exactly the right shape — extending it to the max-messages test would close the last hole.

---

NikolayS:

So this should either:

NikolayS:
Wording correction pushed in 57c1b84.

- `clients/go/options.go`: `WithMaxMessages` doc-comment reframed — default 500 mitigates the common case but is not a hard ceiling. Added a WARNING explaining that `pgque.ack(batch_id)` finishes the entire underlying PgQ batch, so rows beyond `maxMessages` (via `ticker_max_lag` bursts or operator changes to `ticker_max_count`) are skipped after ack. Links to #134.
- `clients/go/README.md`: Consumer-options table row rewritten — no longer calls 500 a "cap"; explains the ack-drops-unreturned-rows behavior and recommends sizing `>= ticker_max_count`.
- PR body updated: `Closes #134` removed; replaced with "Mitigates #134 (Go portion)" and a paragraph noting #134 remains open as the SQL-side root cause.

Tests and runtime code untouched — wording-only change.

---

NikolayS:
## REV review — PR #153 (`fix/clients-go-v0.2.0` @ `169efa4`)

**Verdict: PASS after follow-up fix** — the blocking REV finding was fixed; no known blocking issues remain.

### What was reviewed

REV lanes were run on the Go client PR after the rebase/safety follow-ups:

- Security
- Bug hunter
- Test analyzer
- Guidelines
- Docs

### Blocking finding found and fixed

REV found one real blocker at prior head `870e6df`:

- `Client.Nack` passed `time.Duration` directly as `$12::interval`.
- pgx v5 expects a supported interval value, so `Nack` could fail at encode time.

Fixed in `169efa4`:

- `Client.Nack` now passes `pgtype.Interval{Microseconds: retryAfter.Microseconds(), Valid: true}`.

### Current findings

#### BLOCKING

None known after `169efa4`.

- Add stronger `SendBatch` round-trip assertions for payload order/content and ID-to-message correspondence.
- Add stronger `WithReason(...)` DLQ assertion when retries are exhausted.
- Add negative/default `SendBatch` tests for empty type, marshal failure, and atomic no-partial insert.
- Tighten one test assertion from “Ack > 0” to “Ack == 1”.
- Docs polish:
  - repeat direct `Receive`/`Ack` partial-batch caveat in `Receive` GoDoc;
  - clarify README `WithReason` comment: reason is recorded on DLQ only if retry budget is exhausted.

### Validation evidence

- GitHub CI green on head `169efa4`:
  - PG14–18 SQL tests
  - Python client tests
  - Go client tests
  - TypeScript client tests
  - `verify`
- Local Go validation via `golang:1.23` Docker image:
  - `gofmt`
  - `go test ./...` ✅
  - `go test -race ./...` ✅

### Notes

The high-level consumer now defaults `maxMessages` to `math.MaxInt32` to avoid ack-skipping unreturned rows. REV noted the expected operational tradeoff: extremely large PgQ batches can produce large in-memory result sets. That is intentional and documented.

Issue #149, opened 2026-04-30
TypeScript client: top-level undefined payload is not JSON null as documented


NikolayS:
## Problem

TypeScript `send()` docs say top-level `undefined` becomes JSON `null`, but actual JavaScript behavior is different.

In `clients/typescript/src/client.ts`:

```ts
const payload = JSON.stringify(event.payload);
```

`JSON.stringify(undefined)` returns JavaScript `undefined`, not the string `'null'`. Passing that through `pg` can send SQL NULL, not JSON text `null`.

Cross-driver comparison:

- Python maps `None` to string `'null'` before `::jsonb`.
- Go `json.Marshal(nil)` returns bytes for `null`.
- TypeScript top-level `undefined` risks SQL NULL / unexpected behavior.

## Why this matters

This is a subtle payload corruption / API surprise. SQL NULL payload and JSON null payload are not the same thing in PgQue/PgQ internals and can interact badly with typed clients that assume strings.

## Suggested fix

- Explicitly reject top-level `undefined`, or coerce it to JSON null deliberately:
  ```ts
  const payload = event.payload === undefined ? 'null' : JSON.stringify(event.payload);
  ```
- Keep behavior aligned with Python/Go.
- Add tests for `undefined`, `null`, omitted payload, and object properties containing `undefined`.

## Acceptance criteria

- TypeScript docs match runtime behavior.
- Top-level `undefined` either throws a clear `PgqueSqlError`/input error or stores JSON null consistently.

NikolayS:
Closing as completed. Top-level `undefined` payload now coerces to JSON `'null'` via shared `serializePayload()` in `clients/typescript/src/client.ts`; vitest covers `undefined` / `null` / omitted / nested-`undefined` in `clients/typescript/test/client.test.ts`. Shipped in #154 / #156 (merged).

---
_Generated by [Claude Code](https://claude.ai/code)_

Issue #151, opened 2026-04-30
TypeScript client: ticker()/forceTick() discard SQL return values and docs misstate behavior


NikolayS:
## Problem

TypeScript exposes `ticker()` / `forceTick()` public methods but discards useful SQL return values, and comments/docs are misleading.

SQL returns:

- `pgque.ticker(queue)` → tick id or NULL if no tick was created.
- `pgque.ticker()` → number of queues processed.
- `pgque.force_tick(queue)` → last tick id / status value from SQL.

Current TS methods return `Promise<void>`. The comment above `ticker()` says it returns the underlying tick result count, but the method discards the row.

## Why this matters

These are supposed to be thin wrappers over SQL primitives. Hiding return values makes tests and manual operation less informative, and the comment is currently wrong.

## Suggested fix

- Make `ticker(queue)` return `bigint | null` and `ticker()` return count, or split into `ticker(queue)` and `tickerAll()` to avoid union confusion.
- Make `forceTick(queue)` return the SQL result.
- Update README/API table.
- Keep any combined force+ticker behavior test-helper-only, not public API.

## Acceptance criteria

- TS method signatures match SQL primitive return values or docs explicitly justify discarding them.
- Tests assert return values for tick-created and no-tick-needed cases.

Issue #148, opened 2026-04-30
Clients: do not hide ack() no-op result in TypeScript/Go


NikolayS:
## Problem

Some drivers hide the SQL result of `pgque.ack(batch_id)`.

SQL `pgque.ack` / `finish_batch` returns an integer status: `1` on success, `0` when no active batch was finished (e.g. stale/double ack / batch not found).

Current state:

- Python `ack()` returns the integer.
- TypeScript `ack(batchId): Promise<void>` discards the result.
- Go `Ack(...) error` discards the result.

This means stale/double ack can look successful in TS/Go.

## Why this matters

A no-op ack is often a correctness signal: wrong batch ID, already-finished batch, wrong consumer lifecycle, or race. Hiding it makes debugging harder and weakens tests.

## Suggested fix

Align all drivers:

- Return the integer (`0`/`1`) from `ack`, or
- Raise/return a typed `BatchNotFound`/`NoBatchFinished` error when the SQL result is `0`.

Do the same review for `ticker()` / `force_tick()` wrappers where SQL returns useful values.

## Acceptance criteria

- TS/Go tests prove stale/double ack is not silently treated as success.
- Driver docs state what `ack()` returns or throws.

Issue #147, opened 2026-04-30
TypeScript consumer: use PgQue LISTEN/NOTIFY wakeups instead of poll-only 30s latency


NikolayS:
## Problem

The TypeScript high-level Consumer ignores PgQue's existing LISTEN/NOTIFY wakeup path and only sleeps/polls.

SQL `pgque.ticker(queue)` emits:

```sql
pg_notify('pgque_' || queue_name, tick_id::text)
```

Python Consumer uses `LISTEN pgque_<queue>` and wakes on notify. TypeScript Consumer defaults to `pollInterval = 30_000`, so a message can sit for up to ~30s after a tick even though the database emitted a notification.

Go also currently polls only, so this may be a broader parity decision, but TS is especially visible because the Node `pg` driver supports notifications.

## Why this matters

PgQue docs/positioning imply low-latency delivery with ticker cadence. Poll-only 30s default makes the driver feel sluggish and inconsistent with Python.

## Suggested fix

- Add LISTEN/NOTIFY support to TS Consumer using a dedicated connection/client.
- Keep polling as fallback / safety net.
- Consider doing the same in Go or document why Go intentionally remains poll-only.
- Tests should simulate/verify wake-up without waiting for full poll interval.

## Acceptance criteria

- TS Consumer wakes promptly on `pgque_<queue>` notify.
- README documents latency behavior.
- Cross-driver docs explain whether each Consumer uses LISTEN, polling, or both.

PR #154, opened 2026-04-30
fix(clients/typescript): nack-fail handling, unknown-type policy, undefined payload, safe max_messages default

### LINKED ISSUE #134: Clients: high-level consumers can skip unreturned batch rows when max_messages < batch size

### LINKED ISSUE #135: Clients: Go/TypeScript consumers ack batch even when nack fails
## Problem

Go and TypeScript high-level consumers log `nack()` errors but still `ack()` the batch afterwards.

That means a message whose handler failed or whose event type is unhandled can be lost if `nack()` fails transiently/permanently: the failed message is not moved to retry/DLQ, then the final `ack()` advances the consumer past the whole batch.

## Affected code

Go:

- `clients/go/consumer.go`: missing handler / handler error calls `Nack()`, logs `nackErr`, continues, then `Ack(batchID)`.

TypeScript:

- `clients/typescript/src/consumer.ts`: `tryNack()` catches/logs `client.nack()` errors, caller continues, then `client.ack(batchId)`.

Python is safer here because `_poll_once()` wraps receive/nack/ack in one transaction and does not swallow `nack()` exceptions before `ack()`.

## Expected behavior

If the driver decides a message must be nacked, failure to nack must prevent acking that batch, or at least prevent acking the affected message's batch.

Better to redeliver the whole batch than to mark a failed/unhandled message as consumed without retry/DLQ.

## Suggested fix

- Track whether every required `nack()` succeeded.
- If any `nack()` fails, do **not** `ack()` the batch; return/log a terminal poll-cycle error and let PgQue redeliver.
- Add tests with a fake/mock client or permission-induced `nack()` failure proving `ack()` is not called after failed `nack()`.
- Align Go/TS/Python Consumer docs around this rule.

## Acceptance criteria

- Go and TS tests fail red/green if `ack()` happens after failed `nack()`.
- Runtime behavior preserves at-least-once semantics under `nack()` failure.

NikolayS:
## Summary

Cross-driver v0.2.0 must-fix work for the TypeScript client. Sibling PRs:
- #153 — Go
- #155 — Python

Closes #135 (TS), #136 (TS), #149.
**Mitigates #134** (TS portion). Root-cause fix needs SQL-side partial-ack support; #134 stays open as the follow-up.

### Nack-fail handling (#135)
`Consumer` now tracks per-poll nack failures. If any required `nack()` fails, the batch `ack()` is skipped so PgQ redelivers the affected messages instead of silently advancing past them.

### Unknown-type policy (#136)
`ConsumerOptions.unknownHandlerPolicy` (`'ack' | 'nack'`, default `'nack'`). The new cross-driver default is **nack** (data safety): unknown event types route to the retry queue / DLQ via PgQ. Set `'ack'` for handler-as-filter semantics (logs WARNING, lets the batch ack absorb the message).

### Safer `maxMessages` default (mitigates #134)
`ConsumerOptions.maxMessages` default bumped from `100` → `500` to match PgQue's default `ticker_max_count`. Note: `ticker_max_count` is the *threshold* at which the ticker fires, **not a hard ceiling on batch size** — `ticker_max_lag` can fire on bursts after more than `ticker_max_count` events accumulate, and operators can raise `ticker_max_count`. `pgque.ack(batch_id)` finishes the entire underlying batch, so any rows past `maxMessages` are skipped on ack. The new default removes the data-loss window for the **common case** (default config, batch ≤ 500), but it does **not** close #134. README and JSDoc now spell this out and recommend sizing `maxMessages >= ticker_max_count` for the workload. The structural fix needs SQL-side partial-ack support and is tracked in #134.

### Undefined payload coercion (#149)
`client.send()` (and an omitted payload field) now coerces top-level `undefined` to the JSON literal `null`. `JSON.stringify(undefined)` returns `undefined` which would otherwise bind as SQL NULL and reject the JSONB column.

## Tests

- Red/green vitest mock proving `ack()` is not called after a failed `nack()`.
- Opt-in test for `unknownHandlerPolicy='ack'`.
- Live-pg round-trips for `undefined` / `null` / omitted payload / object property containing `undefined`.

## README

Documents the new defaults, the undefined coercion rule, and (per the wording correction commit) the precise scope of the `maxMessages=500` mitigation vs. the still-open #134.

## Test plan

- [ ] `PGQUE_TEST_DSN=… npx vitest --run` (live PG)
- [ ] `tsc --noEmit`
- [ ] CI green

## Coordination

Part of the v0.2.0 data-safety + Consumer alignment wave. Reverses portions of the warn+ack default that #121 introduced for Python; this PR brings TS into the new cross-driver default (nack).

---


NikolayS:
## REV review — PR #154

**Verdict: REQUEST_CHANGES_BLOCKING**

Implementation looks correct, but the test deliverables and README update advertised in the PR body are not in the diff. `changed_files: 3` is exactly `src/client.ts`, `src/consumer.ts`, `src/types.ts` — no test file or README touched by this commit.

### 1. Security
N/A. No new dynamic SQL; nack still uses parameterized `ROW(...)::pgque.message`. Pass.

### 2. Bugs
- **Nack-fail tracking covers both paths**: handler-error and unknown-type-nack both go through `tryNack` and OR into `anyNackFailed`. Good.
- **`unknownHandlerPolicy='ack'` does NOT early-ack**: it `continue`s and lets the end-of-loop batch ack absorb the message. Correct.
- **Top-level `undefined` coercion**: `event.payload === undefined ? 'null' : JSON.stringify(...)` is the right fix; addresses the SQL-NULL bind hazard. Note: the TS client has no `send_batch` API, so the "list with undefined element" question is moot here.
- **Per-poll failure flag**: `anyNackFailed` is loop-local with sequential `await`s — no race.
- **Subtle but acceptable**: when `unknownHandlerPolicy='ack'` is set and a *different* message in the same batch suffers a failed nack, the acked-by-policy message will also be redelivered (whole batch is skipped). That is the right at-least-once tradeoff and matches issue #135's "better to redeliver the whole batch" guidance, but it is worth a sentence in the JSDoc for `unknownHandlerPolicy: 'ack'` so users don't assume per-message ack.

### 3. Test analysis — BLOCKING
PR body promises three test additions; none are present:
- **Missing**: red/green mock-based test asserting `client.ack` is NOT called after `client.nack` rejects. The closest existing test (`unhandled message types are nacked, not silently consumed`) uses live-pg and asserts retry-queue rows, but does **not** lock in the new behavior — it would have passed before this PR (nack succeeded, ack happened). Issue #135 acceptance criteria explicitly require this test ("Go and TS tests fail red/green if `ack()` happens after failed `nack()`").
- **Missing**: opt-in test for `unknownHandlerPolicy='ack'` (covering both the no-handler warn-and-ack path and the no-early-ack invariant).
- **Missing**: live-pg round-trips for `undefined` / omitted `payload` / `{a: undefined}`. Existing `client.test.ts` covers `null` and nested objects, but nothing exercises the new top-level `undefined → 'null'` branch. Without a test, regressing this back to `JSON.stringify(undefined)` would not be caught.
- Recommend asserting that an `undefined` payload reads back as JSON `null` (not SQL NULL) via `JSON.parse(msg.payload) === null`, and noting in test or doc that `JSON.stringify({a: undefined})` silently drops `a` (already documented in client.ts JSDoc — good).

### 4. Guidelines
Conventional commit subject is fine. Comments describe intent, not narrative. No anti-leak issues. Pass.

### 5. Docs — BLOCKING (minor)
- README is unchanged in this commit despite the PR body saying "Documents the new defaults and the undefined coercion rule." The README's API table and prose still imply the old behavior; new options (`unknownHandlerPolicy`, `maxMessages: 500`) and the `undefined → null` coercion are not surfaced anywhere user-visible.
- `tsc --noEmit` was listed as an unchecked test-plan item; please confirm it runs clean before merge. Types look fine on inspection (`'ack' | 'nack'` literal union, optional property).
- I do not see references to the v0.2.1-deferred items (#147 LISTEN, #148 ack return, #151 ticker return, #145 bigint parser) being added to README in this PR — that is correct, they should not be documented here. The bigint global-mutation caveat is already in README from a prior PR.

### 6. Architecture
CLAUDE.md rule 7 (drivers mirror SQL API 1:1): respected. `unknownHandlerPolicy` is configuration of the existing `Consumer`, not a new public composite method. No `flush()`-style helper. `maxMessages: 500` aligns with `ticker_max_count` and is a tuning default, not a new surface. Pass.

### Summary of blockers
1. Add the red/green mock test for #135 (assert `ack` not called when `nack` rejects). Without this, the bug is unguarded.
2. Add the live-pg `undefined` / omitted-payload round-trip test (#149).
3. Add the opt-in `unknownHandlerPolicy='ack'` test (#136).
4. Update README to document `unknownHandlerPolicy`, `maxMessages: 500` default, and the `undefined → null` coercion rule.

Non-blocking suggestion: tighten the JSDoc on `unknownHandlerPolicy: 'ack'` to note that an unrelated nack failure in the same batch will still cause the absorbed unknown-type message to be redelivered.

Once the test files and README land matching the PR description, this is a straightforward approve.

---

NikolayS:
Added the missing tests and README docs on top of `73bb3b7`:

- `6e825a7` test(clients/typescript): add nack-fail mock + opt-in/undefined coverage
  - `Consumer (in-memory mocks) > does not call ack when nack fails for a handler error` — vitest mock asserts `ack.mock.calls.length === 0` when `nack` rejects.
  - `Consumer.unknownHandlerPolicy=ack (env-gated) > acks unknown event types via the batch when policy is "ack"` — live-pg, asserts `pgque.retry_queue` count is 0, no DLQ row, and the unknown `msgId` is not redelivered.
  - `Client > top-level undefined payload becomes JSON null (explicit)` — live-pg, asserts `m.payload === 'null'` and `JSON.parse(...) === null`.
  - `Client > omitted payload field becomes JSON null` — same expectation when `payload` is absent.
- `cfe07eb` docs(clients/typescript): document v0.2.0 defaults and undefined coercion
  - README now documents `ConsumerOptions.maxMessages` default `500` (with the `ticker_max_count` rationale), `unknownHandlerPolicy: 'ack' | 'nack'` (default `'nack'`, cross-driver consistent), and the top-level-`undefined` → JSON `null` coercion (plus the spec-mandated drop of `undefined` object properties).

Local results: `tsc --noEmit` clean on both project + test configs; `PGQUE_TEST_DSN=… npx vitest --run` reports `Tests 26 passed (26)` (was 22 before, +4 new).

---

NikolayS:
## REV re-review (delta) — head `cfe07eb`

**Verdict: PASS** (non-blocking nits below).

Re-scoped to the two follow-up commits `6e825a7` (tests) and `cfe07eb` (README). The previous round was BLOCKING because tests + README claimed in the original commit message were absent; both have now landed and the assertions are strong enough to lift the block.

### What was verified

**1. Bug-fix mock test (#135) — strong zero-invocation assertion**

`clients/typescript/test/consumer.test.ts` (Consumer in-memory mocks suite):

```ts
expect(fakeClient.nack).toHaveBeenCalledTimes(1);
// Strong assertion: ack must NEVER be called for the batch when its
// nack failed — the batch should be redelivered on the next poll.
expect(fakeClient.ack).toHaveBeenCalledTimes(0);
expect(fakeClient.ack.mock.calls.length).toBe(0);
```

This is exactly what the previous round asked for: the test does not merely prove `nack` threw, it proves `ack` was invoked **zero** times for the batch. The redundant `mock.calls.length` check is belt-and-suspenders and that's fine. The fake client is a structural duck-typed `Client` (`as unknown as Client`) — no public API surface invented.

**2. Live-pg undefined coercion (#149) — JSON literal, not SQL NULL, not JS undefined**

`clients/typescript/test/client.test.ts`:

```ts
// top-level undefined
expect(m!.payload).toBe('null');
expect(JSON.parse(m!.payload)).toBeNull();

// omitted payload
expect(m!.payload).toBe('null');
expect(JSON.parse(m!.payload)).toBeNull();
```

Both flavors (`payload: undefined` and omitted-field-via-cast) round-trip through real Postgres and assert the **string literal `'null'`** — which discriminates against (a) JS `undefined` (`.toBe('null')` would fail), (b) SQL `NULL` from a dropped bind param (would fail the `::jsonb` cast or return `null` not `'null'`), and (c) the bare value `null` (JS) vs. the JSON text `'null'`. The `JSON.parse(...).toBeNull()` follow-up is the right second leg. Implementation in `client.ts` matches: `event.payload === undefined ? 'null' : JSON.stringify(event.payload)`.

**3. Opt-in `unknownHandlerPolicy='ack'` (#136) — proves batch advanced**

`clients/typescript/test/consumer.test.ts` (env-gated suite):

- `pgque.retry_queue` count = 0 for the queue → not routed to retry.
- `pgque.dead_letter` count = 0 for the queue → not routed to DLQ.
- After a fresh `advanceQueue` + `receive(..., 100)`, every returned message satisfies `m.msgId !== unknownId` → the unknown msg_id was acked and is not being redelivered.

That triplet is sufficient evidence that the batch ack absorbed the unknown type. Tests live in `clients/typescript/test/`, no public Client API additions — the policy is a `ConsumerOptions` field consumed in `consumer.ts`, exactly the right seam.

**4. README accuracy**

`clients/typescript/README.md`:

- Consumer-options table documents `maxMessages: 500` (with the rationale tying it to PgQ's `ticker_max_count` ceiling — nice cross-reference) and `unknownHandlerPolicy: 'ack' | 'nack'` with the correct default (`'nack'`). 
- Dedicated "Payload coercion: `undefined` → JSON `null`" section shows all three equivalent forms (`payload: null` / `payload: undefined` / omitted) and correctly distinguishes top-level coercion from in-object property drop (standard `JSON.stringify` behavior). Matches code.
- v0.2.1-deferred items (#147 LISTEN, #148 ack return, #151 ticker return, #145 bigint parser) are correctly **absent** from the README — no premature documentation of unshipped behavior.

### Non-blocking nits

- **Mock test poll loop** uses `while (Date.now() < deadline && nack.mock.calls.length === 0) await sleep(10)` with a 4 s deadline. On a heavily loaded CI runner this could in theory race, but once `nack` is observed the assertions are deterministic. Optional follow-up: gate on `vi.waitFor(...)` for cleaner intent. Not a blocker.
- **Opt-in `ack` policy test** asserts the unknown `msgId` is not redelivered, which is the right outcome, but doesn't directly assert that a `pgq.finish_batch`-equivalent fired. The retry+DLQ=0 plus the post-`advanceQueue` re-receive miss is sufficient transitive evidence; a future test could assert `pgque.consumer` `last_tick`/`sub_last_tick` advanced, but that's gold-plating.
- **`tryNack` reason argument**: the new `reason: string` parameter to `tryNack` is wired through to `client.nack(batchId, msg, { reason })`. That looks like a behavior expansion (now passing a reason where v0.1 did not) — confirm `Client.nack` already accepts `NackOptions.reason` and that this isn't accidentally widening the public contract beyond what's documented. Quick scan of `types.ts` shows `NackOptions` exists, so this is fine; flagging only so it doesn't get lost in v0.2.1 release notes.

### Summary

- 26/26 vitest pass (reported), `tsc --noEmit` clean (reported).
- All four claimed tests landed and assert the right invariants with the right strength.
- README delta matches code delta; deferred items correctly omitted.
- No public API change, no architectural drift, tests in the right directory.

Lifting the BLOCKING flag from the previous round. Ready to merge from a REV perspective; defer formal approval to MERGE per project workflow.

— REV (re-review)

---

NikolayS:

So this should either:

NikolayS:
Second-pass review with local live-DB testing.

What I ran:

```bash
cd clients/typescript
rm -rf node_modules
npm install --no-audit --no-fund
npm run check
# PASS

# fresh DB with current sql/pgque.sql
PGQUE_TEST_DSN=postgresql:///pgque_pr154_review_... npm test
# PASS: 3 files, 26 tests
```

Extra probes I ran against this PR:

- `payload: undefined` now stores JSON text `null`, not SQL NULL:
  ```json
  {"len":1,"payload":"null","type":"string"}
  ```
- in-memory fake client with `nack()` throwing does **not** call `ack()`:
  ```json
  ["receive","nack","receive",...]
  ```
  No `ack` entry. Good.
- 600-event batch with `receive(...,500)` still demonstrates the underlying #134 hazard:
  ```json
  {"first":500,"after":0}
  ```

Verdict: the PR correctly fixes the TypeScript-specific bugs it targets (#135 TS part, #136 TS behavior/knob, #149). The new tests are meaningful and the implementation is simple.

One required correction before calling #134 closed: README/types say `500` matches `ticker_max_count` “ceiling” / “upper bound on how many events end up in one batch”. That is not true. It is a ticker threshold, not a hard batch-size cap. If 600 events arrive before ticker runs, a tick can contain 600 events; `receive(...,500)` + `ack()` skips the remaining 100 for that consumer.

So please change the wording to something like:

> Default `500` matches PgQue's default `ticker_max_count`, covering the normal/default tick threshold. It does not guarantee full-batch draining if the queue is configured differently or if more than 500 events accumulate before a tick; set `maxMessages` high enough for your workload, or use lower-level receive/ack carefully.

And do **not** mark #134 fully closed by this PR. This PR mitigates the common TS default case but does not solve the general SQL/client footgun.

Non-blocking notes:

- It would be nice to add a mock test for `unknownHandlerPolicy='nack'` + `nack()` failure too, not only handler-error + `nack()` failure. The same code path is used, so this is not a blocker.
- The existing unknown-type live test only asserts one retry row; adequate for this PR.

Merge verdict after README/type comment wording is corrected: yes for TS. Keep #134 open.

NikolayS:
Wording correction pushed in `96b6256` and PR body updated.

The previous README/JSDoc called `500` (`ticker_max_count`) a "ceiling" / "upper bound on how many events end up in one batch." That is wrong: `ticker_max_count` is the *threshold* at which the ticker fires, not a hard cap. A 600-event batch fetched with `receive(..., 500)` still loses the trailing 100 rows after `ack` (`{"first":500,"after":0}`), as confirmed in review.

What changed in this commit:

- **README** (`clients/typescript/README.md`): the `maxMessages` row now states explicitly that `500` is a threshold, not a ceiling; that this mitigates the common case but does **not** prevent loss when batches exceed `maxMessages`; that bursts via `ticker_max_lag` and operator changes to `ticker_max_count` can produce larger batches; that ack finishes the entire underlying batch; and that #134 is still open pending a SQL-side partial-ack fix.
- **JSDoc** (`clients/typescript/src/types.ts`, `ConsumerOptions.maxMessages`): same WARNING in TSDoc form, with link to #134.
- **PR body**: replaced "Closes #134 (TS)" with "Mitigates #134 (TS portion)"; the issue stays open.

No source/test changes. Plain `git push` (no force).

---

NikolayS:
## REV review — PR #154 (`fix/clients-typescript-v0.2.0` @ `b07e2a7`)

**Verdict: PASS** — no blocking issues found after the follow-up fixes.

### What was reviewed

Fresh REV lanes were run against head `b07e2a7` after the documentation/version cleanup and safety follow-ups:

- Security
- Bug hunter
- Test analyzer
- Guidelines
- Docs

### Findings

#### BLOCKING

None.

- `send()` serialization rejection tests currently cover top-level function payloads; add low-cost coverage for `Symbol()`, `BigInt`, and circular refs later.
- A couple of unknown-handler integration tests use fixed sleeps; polling-to-deadline would be less timing-sensitive.
- Docs/API wording nits only:
  - circular refs are wrapped as `PgqueSqlError` with the original error as `cause`, not surfaced as raw `TypeError`;
  - nested array `undefined` serializes as `null` while object properties are dropped;
  - `unknownHandlerPolicy='ack'` logs and discards after batch ack, so “silent discard” can be worded more precisely.

### Validation evidence

- GitHub CI green on head `b07e2a7`:
  - PG14–18 SQL tests
  - Python client tests
  - Go client tests
  - TypeScript client tests
  - `verify`
- Local TypeScript validation using Bun:
  - `bun install --frozen-lockfile`
  - `bun run check` ✅
  - `bun test` ✅ (`5 pass`, DB-gated tests skipped locally)

### Notes

The high-level consumer now defaults `maxMessages` to PostgreSQL `int` max to avoid ack-skipping unreturned rows. REV noted the expected operational tradeoff: huge PgQ batches can now produce large in-memory result sets. That is intentional and documented.

Issue #137, opened 2026-04-30
Python client: tests fail on fresh PgQue install and encode weak transaction assertions


NikolayS:
## Problem

Python client tests currently fail on a fresh PgQue install.

## Evidence

Fresh temp DB, installed current `sql/pgque.sql`, temp venv with `clients/python[dev]`:

```bash
python -m pytest -q clients/python/tests
# 39 passed, 2 failed
```

Failures observed:

1. `test_send_batch_none_payload_produces_json_null`
   - Test does `send_batch()` then `force_tick()`/`ticker()` in the same transaction before commit.
   - PgQue docs/semantics require send, tick, receive in separate transactions for visibility.
   - The test expected one message but got zero.

2. `test_consumer_warns_and_acks_unhandled_event_type`
   - Did not capture the expected warning.
   - The test also only checks `retry_queue` count is zero, which does not prove the message was actually consumed/acked; it could also be not delivered.

## Why this matters

These tests are supposed to validate client semantics, but currently encode/permit transaction mistakes and weak assertions. That makes future driver changes hard to trust.

## Suggested fix

- Fix transaction discipline in tests: commit after `send`/`send_batch`, then run `force_tick(queue)` + per-queue `ticker(queue)`, then receive/consume.
- Strengthen unknown-type test to prove the batch advanced / message is not redelivered after ack.
- Prefer per-queue `pgque.ticker(queue)` in tests to avoid cross-queue side effects.
- Keep a regression test for `send_batch([None])` storing JSON null, but make it transaction-correct.

## Acceptance criteria

- `python -m pytest -q clients/python/tests` passes against a fresh DB with `sql/pgque.sql`.
- Tests fail if send/tick/receive are accidentally collapsed into one transaction where visibility should not happen.

NikolayS:
Closing as completed. Acceptance criteria met by #155 (consumer warn+ack default reversed to nack) and #139 (transaction discipline fix in `test_consumer_warns_and_acks_unhandled_event_type` and `test_send_batch_none_payload_produces_json_null`). Both merged.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #155, opened 2026-04-30
fix(clients/python): nack unknown types by default + safe max_messages default

### LINKED ISSUE #136: Clients: align unknown event type policy across Python, Go, and TypeScript consumers

NikolayS:
## Summary

- Reverse PR #121's warn+ack default: the Python `Consumer` now **nacks** any message whose type has no registered handler and no `"*"` catch-all. Unknown types follow the standard retry / dead-letter path instead of being silently acknowledged.
- Add an explicit `Consumer(..., unknown_handler="nack" | "ack")` knob; default `"nack"`. `"ack"` opts back into the legacy "log WARNING and ack" behavior.
- Bump default `Consumer.max_messages` from 100 to **500** to match the queue's default `ticker_max_count`, so a single `receive()` drains a typical tick.
- Strengthen the unknown-type tests: prove the message is routed to `retry_queue` (or `dead_letter`) **and** that a follow-up `receive()` does not return the same `msg_id` -- the batch actually advanced.
- Drive-by: fix a snapshot-isolation slip in `test_send_batch_none_payload_produces_json_null` (commit between `send_batch` and `force_tick`).

Closes #136 (Python part), #134 (Python part), #137.

## Test plan

- [x] `PGQUE_TEST_DSN=... pytest -v clients/python/tests/` -- 43 passed
- [x] New `test_consumer_nacks_unhandled_event_type` asserts retry_queue / dead_letter row + msg_id does not reappear
- [x] New `test_consumer_acks_unhandled_event_type_when_opt_in` asserts no retry_queue row, WARNING logged, msg_id does not reappear
- [x] New `test_consumer_rejects_invalid_unknown_handler` covers the constructor validation path
- [x] CI matrix (PG 14-18) on push

---


NikolayS:
## REV review — PR #155

**Verdict: PASS (REQUEST_CHANGES_NON_BLOCKING)** — ship-ready; one transaction-state concern worth filing as a follow-up.

### 1. Security
`reason=f"no handler for type={msg.type}"` flows from `pgque.message.type` into a Python f-string and lands in `pgque.dead_letter.dl_reason` (parameterized via psycopg, no SQL injection). `msg.type` is the producer-supplied event type string; logs and DLQ rows accept arbitrary text. No log-injection or query-injection risk. Clean.

### 2. Bugs
- **Default change is correct**: Literal-typed param + explicit `if not in (...)` ValueError covers the static-typer-bypass path. Test `test_consumer_rejects_invalid_unknown_handler` confirms.
- **Non-blocking concern — transaction state on nack failure**: The new nack runs inside the outer `with conn.transaction():` block. The `except Exception` catches a `client.nack` failure and lets the loop continue to `client.ack(batch_id)`. If the nack raised a `psycopg.Error`, the underlying transaction is in an aborted state and the subsequent ack will also raise (and the transaction context will rollback the whole batch — net effect: messages get re-delivered, not lost, which is actually safe). Worth a comment or a savepoint per nack in a follow-up, but not a blocker for v0.2.0.
- **Strong assertions**: New tests assert presence in `retry_queue ∪ dead_letter` AND that a follow-up `receive()` does not return the same `msg_id`. The "doesn't reappear" check is partially load-bearing on `retry_after=60s` not having elapsed during the 3s consumer window — which works but is implicit. Consider asserting `rq.ev_retry_after` directly in a future iteration. Acceptable as-is.

### 3. Test analysis
- Genuinely red→green: removed `test_consumer_warns_and_acks_unhandled_event_type` was the canonical test for #121's now-reverted default; nothing else in `tests/` depends on the warn+ack default (verified via PR diff scope).
- Column reference `dl.ev_id` is **correct** — `pgque.dead_letter` defines `ev_id bigint not null` (the original event id), with `dl_id` as PK and `dl_queue_id` as FK. The test query joins on `dl.dl_queue_id` and filters `dl.ev_id = %s` — both column names match the schema.
- Both new behavior tests `commit()` between `send` and `force_tick` — addresses #137 snapshot-isolation issue correctly. The drive-by fix in `test_send.py` applies the same pattern to a pre-existing latent bug.
- `caplog` switched from `r.message` to `r.getMessage()` — correct fix; `LogRecord.message` is only populated after `Formatter.format()` runs, while `getMessage()` always renders the args.

### 4. Guidelines
- Conventional commit prefix `fix(clients/python):` ✓
- No narrative source comments added; existing PgQ ISC notice preserved in `consumer.py` header ✓
- `unknown_handler` is constructor config, **not** a new public SQL composite — CLAUDE.md rule 7 (1:1 driver/SQL mirror) preserved ✓
- Reversal of #121 explicitly called out in PR body opener ("Reverse PR #121's warn+ack default") ✓

### 5. Docs
- `clients/python/README.md`: stale "logged at WARNING and acked" wording replaced; new "Handling unknown event types" section documents both default and opt-in. ✓
- Class docstring in `consumer.py` updated to match new default semantics and references `unknown_handler="ack"` opt-in. ✓
- One nit: the README opt-in example uses `unknown_handler="ack"` but doesn't mention the WARNING-only-no-DLQ tradeoff that the docstring covers. Minor.

### 6. Architecture
- No new public composite types. `unknown_handler` parameterizes existing receive/ack/nack wrappers — the warn+ack heritage path remains reachable via opt-in, so #121's intent is preserved as a knob rather than erased.
- `max_messages` 100→500 aligns the driver default with `ticker_max_count` default; reasonable behavioral tweak, documented in PR body.
- Maintainer-accepted reversal per Leo's #136 cross-driver alignment audit; consistent with nack-default direction for go/ts drivers.

### Top findings
1. (non-blocking) Nack-inside-transaction failure path may abort the batch txn; worth a SAVEPOINT or explicit comment in v0.2.x.
2. (nit) "Batch advanced" check leans on `retry_after=60s` window; explicit `ev_retry_after` assertion would be more robust.
3. (nit) README opt-in section could mirror the docstring's clearer "WARNING only, no DLQ" framing.

Closes #136 (Python), #134 (Python), #137 — all addressed by this change. 43/43 local pass on PG16 reported; CI matrix (PG14–18) should confirm.

---

NikolayS:

So this should either:

NikolayS:
Additional Python-specific blocker from second-pass audit: `Consumer.start()` still appears to use the same potentially hanging `conn.notifies(timeout=...)` generator loop. I filed #158 with the repro shape.

Also, in this PR unknown-handler default changes to `nack`, but the unknown-handler `nack()` path catches/logs exceptions and then still reaches final `client.ack(batch_id)`. That recreates the #135 failure mode for Python unknown events: if the required `nack()` fails, the batch should not be acked.

NikolayS:
Pushed two follow-up commits addressing the deeper second-pass audit findings:

- `45af965` `fix(clients/python): do not ack after failed nack on unknown-type` — same data-loss class as the Go/TS fixes (#153/#154). `_poll_once` now tracks a per-batch `nack_failed` flag; on any nack failure (unknown-handler path or handler-error path) the batch is NOT acked, so PgQ redelivers on the next poll.
- `59da923` `fix(clients/python): bounded LISTEN/NOTIFY wait for prompt stop` — replaces the unbounded `conn.notifies(timeout=poll_interval)` generator with a bounded-slice loop (200ms slices, `stop_after=1`) that checks `self._running` between slices. `stop()` returns within ~200ms regardless of `poll_interval`, and a real `pg_notify` wakes the consumer on the next slice boundary.

New tests:
- `test_consumer_does_not_ack_when_unknown_type_nack_fails` (mock asserts `client.ack` is never called when `client.nack` raises)
- `test_consumer_does_not_ack_when_handler_error_nack_fails` (same, handler-error path)
- `test_consumer_stop_returns_within_2s_while_waiting` (`poll_interval=30`, `stop()` returns in <2s)
- `test_consumer_wakes_on_pg_notify_before_poll_interval` (producer-thread NOTIFY wakes consumer well under the 30s polling deadline)

Local: `PGQUE_TEST_DSN=... pytest -v clients/python/tests/` — **47 passed**.

---

NikolayS:
## REV review — PR #155 (`fix/clients-python-v0.2.0` @ `4f27c5e`)

**Verdict: PASS after follow-up fixes** — no known blocking issues remain.

### What was reviewed

REV lanes were run on the Python client PR and follow-up findings were addressed:

- Security
- Bug hunter
- Test analyzer
- Guidelines
- Docs

### Findings fixed during this pass

- Restored/imported the `select`-based bounded LISTEN/NOTIFY wait path.
- Changed high-level `Consumer.max_messages` default to PostgreSQL `int` max (`2_147_483_647`) so the consumer requests the whole PgQ batch before acking.
- Removed stale/dead `_wait_for_notify()` helper and duplicate slice constant.
- Reworded README/API comments to final-state wording (no historical “legacy/previous behavior” framing).
- Added/strengthened tests:
  - default/configured `max_messages` constructor state;
  - `_poll_once()` passes default/configured `max_messages` into `PgqueClient.receive()`;
  - `unknown_handler="ack"` does not create retry or DLQ rows;
  - mixed `send_batch` payload order/content including JSON null.

### Current findings

#### BLOCKING

None.

None that need to block merge.

#### POTENTIAL / operational note

The high-level consumer default now requests the whole PgQ batch. This avoids ack-skipping unreturned rows, but a very large PgQ batch can materialize a large Python list via `fetchall()` and increase memory/latency. This is the same intentional safety tradeoff used in the Go and TypeScript client fixes.

### Validation evidence

- GitHub CI green on head `4f27c5e`:
  - PG14–18 SQL tests
  - Python client tests
  - Go client tests
  - TypeScript client tests
  - `verify`
- Local Python validation against PostgreSQL 18:
  - installed current `sql/pgque.sql`
  - `PGQUE_TEST_DSN=... pytest -q clients/python/tests/`
  - result: `54 passed` ✅
- Local syntax check:
  - `python3 -m compileall -q clients/python/pgque clients/python/tests` ✅

### Final status

Ready for maintainer approval / merge.

PR #162, opened 2026-05-01
chore(clients/typescript): standardize on bun

NikolayS:
## Summary

Standardize the TypeScript client development/CI workflow on Bun while keeping the published package install docs npm-compatible for application users.

This is intentionally separate from PR #161 (batch API + benchmarks). This PR only changes TypeScript client package-manager/tooling and the TypeScript smoke gate.

## Before

The TypeScript client CI used npm commands:

```bash
npm install
npm run check
npx tsx src/smoke.ts
```

The package did not declare a package manager and had no Bun lockfile.

## After

The TypeScript client CI uses Bun:

```bash
bun install --frozen-lockfile
bun run check
bun run test
bun src/smoke.ts
```

Also:

- added `clients/typescript/bun.lock`
- added `"packageManager": "bun@1.3.8"`
- pinned `oven-sh/setup-bun` by commit SHA
- removed now-unused `tsx`
- README documents Bun for repo/client development while still showing npm/bun/pnpm/yarn install options for application consumers
- TypeScript smoke now explicitly calls both `forceTick()` and `ticker()` before receive; `forceTick()` bumps the event-seq threshold, while `ticker()` creates the tick that makes events visible
- CI sets `PGQUE_TEST_DSN`, so TypeScript smoke/tests run against the live Postgres container instead of silently skipping

## Why this is better

- deterministic TypeScript dependency install via Bun lockfile
- faster and standardized TypeScript client CI commands
- no confusion that library consumers must use Bun: published package remains installable with npm-compatible package managers
- smoke test now exercises the real DB path in CI

## Tests

Local:

```bash
cd clients/typescript
bun install --frozen-lockfile
bun run check
bun run test
PGQUE_TEST_DSN=postgresql://postgres:pgque_test@localhost/pgque_batch_test bun src/smoke.ts
```

All passed locally.

CI: pending after latest cleanup commit.

## REV

REV was run. Initial findings were fixed:

- restored npm-compatible install docs for application users
- clarified Node.js runtime vs Bun development tooling
- pinned setup-bun action by commit SHA
- added live `PGQUE_TEST_DSN` for client-smoke
- added `bun run test` to CI
- documented `forceTick()` + `ticker()` in smoke
- standardized smoke execution on `bun src/smoke.ts`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -132,6 +132,7 @@ jobs:
           cd clients/typescript
           bun install --frozen-lockfile
           bun run check
+          bun run test
           bunx tsx src/smoke.ts
 
       - name: Cleanup client smoke DB
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -120,12 +120,17 @@ jobs:
           cd clients/go
           go test -run TestSmoke ./...
 
+      - name: Setup Bun
+        uses: oven-sh/setup-bun@v2
+        with:
+          bun-version: 1.3.8
+
       - name: TypeScript smoke
         run: |
           cd clients/typescript
-          npm install
-          npm run check
-          npx tsx src/smoke.ts
+          bun install --frozen-lockfile
+          bun run check
+          bunx tsx src/smoke.ts
 
       - name: Cleanup client smoke DB
         if: always()
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -79,6 +79,8 @@ jobs:
 
   client-smoke:
     runs-on: ubuntu-latest
+    env:
+      PGQUE_TEST_DSN: postgresql://postgres:pgque_test@localhost/pgque_test
     steps:
       - uses: actions/checkout@v4
         with:
@@ -121,7 +123,7 @@ jobs:
           go test -run TestSmoke ./...
 
       - name: Setup Bun
-        uses: oven-sh/setup-bun@v2
+        uses: oven-sh/setup-bun@0c5077e51419868618aeaa5fe8019c62421857d6 # v2
         with:
           bun-version: 1.3.8
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -133,7 +133,7 @@ jobs:
           bun install --frozen-lockfile
           bun run check
           bun run test
-          bunx tsx src/smoke.ts
+          bun src/smoke.ts
 
       - name: Cleanup client smoke DB
         if: always()
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -123,7 +123,7 @@ jobs:
           go test -run TestSmoke ./...
 
       - name: Setup Bun
-        uses: oven-sh/setup-bun@0c5077e51419868618aeaa5fe8019c62421857d6 # v2
+        uses: oven-sh/setup-bun@0c5077e51419868618aeaa5fe8019c62421857d6 # v2.2.0
         with:
           bun-version: 1.3.8
 
--- a/clients/typescript/README.md
+++ b/clients/typescript/README.md
@@ -109,11 +109,13 @@ package is imported.
 
 ## Tests
 
-The integration tests need a running PostgreSQL with the PgQue schema
-installed and `pgque_admin`-equivalent privileges:
+The repository standardizes on Bun 1.3.8 for TypeScript client development.
+The integration tests need a running PostgreSQL with the PgQue schema installed
+and `pgque_admin`-equivalent privileges:
 
 ```bash
-PGQUE_TEST_DSN=postgres://postgres:pgque_test@localhost/pgque_test \
+bun install --frozen-lockfile
+PGQUE_TEST_DSN=postgresql://postgres:pgque_test@localhost/pgque_test \
   bun run test
 ```
 
--- a/clients/typescript/README.md
+++ b/clients/typescript/README.md
@@ -18,8 +18,7 @@ npm install pgque
 ```
 
 Runtime requirements: Node.js 20+ and PostgreSQL 14+ with the PgQue schema
-installed (`\i pgque.sql` — no extension required). This repository uses Bun
-for TypeScript client development and CI commands.
+installed (`\i pgque.sql` — no extension required).
 
 ## Quickstart
 
@@ -109,9 +108,9 @@ package is imported.
 
 ## Tests
 
-The repository standardizes on Bun 1.3.8 for TypeScript client development.
-The integration tests need a running PostgreSQL with the PgQue schema installed
-and `pgque_admin`-equivalent privileges:
+The repository standardizes on Bun 1.3.8 for TypeScript client development and
+CI commands. The integration tests need a running PostgreSQL with the PgQue
+schema installed and `pgque_admin`-equivalent privileges:
 
 ```bash
 bun install --frozen-lockfile
--- a/clients/typescript/README.md
+++ b/clients/typescript/README.md
@@ -7,12 +7,19 @@ PgQ-based universal PostgreSQL queue. Thin, idiomatic wrapper over the
 
 ## Install
 
+For application code, install the published package with any npm-compatible
+package manager:
+
 ```bash
-bun add pgque
+npm install pgque
+# or: bun add pgque
+# or: pnpm add pgque
+# or: yarn add pgque
 ```
 
-Requires Node.js 20+ and PostgreSQL 14+ with the PgQue schema installed
-(`\i pgque.sql` — no extension required).
+Runtime requirements: Node.js 20+ and PostgreSQL 14+ with the PgQue schema
+installed (`\i pgque.sql` — no extension required). This repository uses Bun
+for TypeScript client development and CI commands.
 
 ## Quickstart
 
--- a/clients/typescript/README.md
+++ b/clients/typescript/README.md
@@ -108,9 +108,9 @@ package is imported.
 
 ## Tests
 
-The repository standardizes on Bun 1.3.8 for TypeScript client development and
-CI commands. The integration tests need a running PostgreSQL with the PgQue
-schema installed and `pgque_admin`-equivalent privileges:
+The repository standardizes on Bun for TypeScript client development and CI
+commands. The integration tests need a running PostgreSQL with the PgQue schema
+installed and `pgque_admin`-equivalent privileges:
 
 ```bash
 bun install --frozen-lockfile
--- a/clients/typescript/README.md
+++ b/clients/typescript/README.md
@@ -8,7 +8,7 @@ PgQ-based universal PostgreSQL queue. Thin, idiomatic wrapper over the
 ## Install
 
 ```bash
-npm install pgque
+bun add pgque
 ```
 
 Requires Node.js 20+ and PostgreSQL 14+ with the PgQue schema installed
@@ -107,7 +107,7 @@ installed and `pgque_admin`-equivalent privileges:
 
 ```bash
 PGQUE_TEST_DSN=postgres://postgres:pgque_test@localhost/pgque_test \
-  npm test
+  bun run test
 ```
 
 Without `PGQUE_TEST_DSN` the integration tests skip.
--- a/clients/typescript/README.md
+++ b/clients/typescript/README.md
@@ -64,6 +64,8 @@ try {
 | `client.nack(batchId, msg, opts?)` | Single-message retry/DLQ. |
 | `client.subscribe(queue, consumer)` | Wraps `pgque.register_consumer`. |
 | `client.unsubscribe(queue, consumer)` | Wraps `pgque.unregister_consumer`. |
+| `client.forceTick(queue)` | Bump the event-seq threshold so the next ticker run produces a tick. |
+| `client.ticker(queue?)` | Run pgque ticker globally or for one queue; makes eligible events visible to consumers. |
 | `client.newConsumer(queue, name, opts?)` | High-level poll loop. |
 | `consumer.handle(eventType, fn)` | Register a handler. |
 | `consumer.start(signal?)` | Run; resolves when `AbortSignal` aborts. |
--- a/clients/typescript/bun.lock
+++ b/clients/typescript/bun.lock
@@ -10,64 +10,57 @@
       "devDependencies": {
         "@types/node": "^22.15.3",
         "@types/pg": "^8.15.2",
-        "tsx": "^4.19.3",
         "typescript": "^5.8.3",
         "vitest": "^2.1.9",
       },
     },
   },
   "packages": {
-    "@esbuild/aix-ppc64": ["@esbuild/aix-ppc64@0.27.7", "", { "os": "aix", "cpu": "ppc64" }, "sha512-EKX3Qwmhz1eMdEJokhALr0YiD0lhQNwDqkPYyPhiSwKrh7/4KRjQc04sZ8db+5DVVnZ1LmbNDI1uAMPEUBnQPg=="],
+    "@esbuild/aix-ppc64": ["@esbuild/aix-ppc64@0.21.5", "", { "os": "aix", "cpu": "ppc64" }, "sha512-1SDgH6ZSPTlggy1yI6+Dbkiz8xzpHJEVAlF/AM1tHPLsf5STom9rwtjE4hKAF20FfXXNTFqEYXyJNWh1GiZedQ=="],
 
-    "@esbuild/android-arm": ["@esbuild/android-arm@0.27.7", "", { "os": "android", "cpu": "arm" }, "sha512-jbPXvB4Yj2yBV7HUfE2KHe4GJX51QplCN1pGbYjvsyCZbQmies29EoJbkEc+vYuU5o45AfQn37vZlyXy4YJ8RQ=="],
+    "@esbuild/android-arm": ["@esbuild/android-arm@0.21.5", "", { "os": "android", "cpu": "arm" }, "sha512-vCPvzSjpPHEi1siZdlvAlsPxXl7WbOVUBBAowWug4rJHb68Ox8KualB+1ocNvT5fjv6wpkX6o/iEpbDrf68zcg=="],
 
-    "@esbuild/android-arm64": ["@esbuild/android-arm64@0.27.7", "", { "os": "android", "cpu": "arm64" }, "sha512-62dPZHpIXzvChfvfLJow3q5dDtiNMkwiRzPylSCfriLvZeq0a1bWChrGx/BbUbPwOrsWKMn8idSllklzBy+dgQ=="],
+    "@esbuild/android-arm64": ["@esbuild/android-arm64@0.21.5", "", { "os": "android", "cpu": "arm64" }, "sha512-c0uX9VAUBQ7dTDCjq+wdyGLowMdtR/GoC2U5IYk/7D1H1JYC0qseD7+11iMP2mRLN9RcCMRcjC4YMclCzGwS/A=="],
 
-    "@esbuild/android-x64": ["@esbuild/android-x64@0.27.7", "", { "os": "android", "cpu": "x64" }, "sha512-x5VpMODneVDb70PYV2VQOmIUUiBtY3D3mPBG8NxVk5CogneYhkR7MmM3yR/uMdITLrC1ml/NV1rj4bMJuy9MCg=="],
+    "@esbuild/android-x64": ["@esbuild/android-x64@0.21.5", "", { "os": "android", "cpu": "x64" }, "sha512-D7aPRUUNHRBwHxzxRvp856rjUHRFW1SdQATKXH2hqA0kAZb1hKmi02OpYRacl0TxIGz/ZmXWlbZgjwWYaCakTA=="],
 
-    "@esbuild/darwin-arm64": ["@esbuild/darwin-arm64@0.27.7", "", { "os": "darwin", "cpu": "arm64" }, "sha512-5lckdqeuBPlKUwvoCXIgI2D9/ABmPq3Rdp7IfL70393YgaASt7tbju3Ac+ePVi3KDH6N2RqePfHnXkaDtY9fkw=="],
+    "@esbuild/darwin-arm64": ["@esbuild/darwin-arm64@0.21.5", "", { "os": "darwin", "cpu": "arm64" }, "sha512-DwqXqZyuk5AiWWf3UfLiRDJ5EDd49zg6O9wclZ7kUMv2WRFr4HKjXp/5t8JZ11QbQfUS6/cRCKGwYhtNAY88kQ=="],
 
-    "@esbuild/darwin-x64": ["@esbuild/darwin-x64@0.27.7", "", { "os": "darwin", "cpu": "x64" }, "sha512-rYnXrKcXuT7Z+WL5K980jVFdvVKhCHhUwid+dDYQpH+qu+TefcomiMAJpIiC2EM3Rjtq0sO3StMV/+3w3MyyqQ=="],
+    "@esbuild/darwin-x64": ["@esbuild/darwin-x64@0.21.5", "", { "os": "darwin", "cpu": "x64" }, "sha512-se/JjF8NlmKVG4kNIuyWMV/22ZaerB+qaSi5MdrXtd6R08kvs2qCN4C09miupktDitvh8jRFflwGFBQcxZRjbw=="],
 
-    "@esbuild/freebsd-arm64": ["@esbuild/freebsd-arm64@0.27.7", "", { "os": "freebsd", "cpu": "arm64" }, "sha512-B48PqeCsEgOtzME2GbNM2roU29AMTuOIN91dsMO30t+Ydis3z/3Ngoj5hhnsOSSwNzS+6JppqWsuhTp6E82l2w=="],
+    "@esbuild/freebsd-arm64": ["@esbuild/freebsd-arm64@0.21.5", "", { "os": "freebsd", "cpu": "arm64" }, "sha512-5JcRxxRDUJLX8JXp/wcBCy3pENnCgBR9bN6JsY4OmhfUtIHe3ZW0mawA7+RDAcMLrMIZaf03NlQiX9DGyB8h4g=="],
 
-    "@esbuild/freebsd-x64": ["@esbuild/freebsd-x64@0.27.7", "", { "os": "freebsd", "cpu": "x64" }, "sha512-jOBDK5XEjA4m5IJK3bpAQF9/Lelu/Z9ZcdhTRLf4cajlB+8VEhFFRjWgfy3M1O4rO2GQ/b2dLwCUGpiF/eATNQ=="],
+    "@esbuild/freebsd-x64": ["@esbuild/freebsd-x64@0.21.5", "", { "os": "freebsd", "cpu": "x64" }, "sha512-J95kNBj1zkbMXtHVH29bBriQygMXqoVQOQYA+ISs0/2l3T9/kj42ow2mpqerRBxDJnmkUDCaQT/dfNXWX/ZZCQ=="],
 
-    "@esbuild/linux-arm": ["@esbuild/linux-arm@0.27.7", "", { "os": "linux", "cpu": "arm" }, "sha512-RkT/YXYBTSULo3+af8Ib0ykH8u2MBh57o7q/DAs3lTJlyVQkgQvlrPTnjIzzRPQyavxtPtfg0EopvDyIt0j1rA=="],
+    "@esbuild/linux-arm": ["@esbuild/linux-arm@0.21.5", "", { "os": "linux", "cpu": "arm" }, "sha512-bPb5AHZtbeNGjCKVZ9UGqGwo8EUu4cLq68E95A53KlxAPRmUyYv2D6F0uUI65XisGOL1hBP5mTronbgo+0bFcA=="],
 
-    "@esbuild/linux-arm64": ["@esbuild/linux-arm64@0.27.7", "", { "os": "linux", "cpu": "arm64" }, "sha512-RZPHBoxXuNnPQO9rvjh5jdkRmVizktkT7TCDkDmQ0W2SwHInKCAV95GRuvdSvA7w4VMwfCjUiPwDi0ZO6Nfe9A=="],
+    "@esbuild/linux-arm64": ["@esbuild/linux-arm64@0.21.5", "", { "os": "linux", "cpu": "arm64" }, "sha512-ibKvmyYzKsBeX8d8I7MH/TMfWDXBF3db4qM6sy+7re0YXya+K1cem3on9XgdT2EQGMu4hQyZhan7TeQ8XkGp4Q=="],
 
-    "@esbuild/linux-ia32": ["@esbuild/linux-ia32@0.27.7", "", { "os": "linux", "cpu": "ia32" }, "sha512-GA48aKNkyQDbd3KtkplYWT102C5sn/EZTY4XROkxONgruHPU72l+gW+FfF8tf2cFjeHaRbWpOYa/uRBz/Xq1Pg=="],
+    "@esbuild/linux-ia32": ["@esbuild/linux-ia32@0.21.5", "", { "os": "linux", "cpu": "ia32" }, "sha512-YvjXDqLRqPDl2dvRODYmmhz4rPeVKYvppfGYKSNGdyZkA01046pLWyRKKI3ax8fbJoK5QbxblURkwK/MWY18Tg=="],
 
-    "@esbuild/linux-loong64": ["@esbuild/linux-loong64@0.27.7", "", { "os": "linux", "cpu": "none" }, "sha512-a4POruNM2oWsD4WKvBSEKGIiWQF8fZOAsycHOt6JBpZ+JN2n2JH9WAv56SOyu9X5IqAjqSIPTaJkqN8F7XOQ5Q=="],
+    "@esbuild/linux-loong64": ["@esbuild/linux-loong64@0.21.5", "", { "os": "linux", "cpu": "none" }, "sha512-uHf1BmMG8qEvzdrzAqg2SIG/02+4/DHB6a9Kbya0XDvwDEKCoC8ZRWI5JJvNdUjtciBGFQ5PuBlpEOXQj+JQSg=="],
 
-    "@esbuild/linux-mips64el": ["@esbuild/linux-mips64el@0.27.7", "", { "os": "linux", "cpu": "none" }, "sha512-KabT5I6StirGfIz0FMgl1I+R1H73Gp0ofL9A3nG3i/cYFJzKHhouBV5VWK1CSgKvVaG4q1RNpCTR2LuTVB3fIw=="],
+    "@esbuild/linux-mips64el": ["@esbuild/linux-mips64el@0.21.5", "", { "os": "linux", "cpu": "none" }, "sha512-IajOmO+KJK23bj52dFSNCMsz1QP1DqM6cwLUv3W1QwyxkyIWecfafnI555fvSGqEKwjMXVLokcV5ygHW5b3Jbg=="],
 
-    "@esbuild/linux-ppc64": ["@esbuild/linux-ppc64@0.27.7", "", { "os": "linux", "cpu": "ppc64" }, "sha512-gRsL4x6wsGHGRqhtI+ifpN/vpOFTQtnbsupUF5R5YTAg+y/lKelYR1hXbnBdzDjGbMYjVJLJTd2OFmMewAgwlQ=="],
+    "@esbuild/linux-ppc64": ["@esbuild/linux-ppc64@0.21.5", "", { "os": "linux", "cpu": "ppc64" }, "sha512-1hHV/Z4OEfMwpLO8rp7CvlhBDnjsC3CttJXIhBi+5Aj5r+MBvy4egg7wCbe//hSsT+RvDAG7s81tAvpL2XAE4w=="],
 
-    "@esbuild/linux-riscv64": ["@esbuild/linux-riscv64@0.27.7", "", { "os": "linux", "cpu": "none" }, "sha512-hL25LbxO1QOngGzu2U5xeXtxXcW+/GvMN3ejANqXkxZ/opySAZMrc+9LY/WyjAan41unrR3YrmtTsUpwT66InQ=="],
+    "@esbuild/linux-riscv64": ["@esbuild/linux-riscv64@0.21.5", "", { "os": "linux", "cpu": "none" }, "sha512-2HdXDMd9GMgTGrPWnJzP2ALSokE/0O5HhTUvWIbD3YdjME8JwvSCnNGBnTThKGEB91OZhzrJ4qIIxk/SBmyDDA=="],
 
-    "@esbuild/linux-s390x": ["@esbuild/linux-s390x@0.27.7", "", { "os": "linux", "cpu": "s390x" }, "sha512-2k8go8Ycu1Kb46vEelhu1vqEP+UeRVj2zY1pSuPdgvbd5ykAw82Lrro28vXUrRmzEsUV0NzCf54yARIK8r0fdw=="],
+    "@esbuild/linux-s390x": ["@esbuild/linux-s390x@0.21.5", "", { "os": "linux", "cpu": "s390x" }, "sha512-zus5sxzqBJD3eXxwvjN1yQkRepANgxE9lgOW2qLnmr8ikMTphkjgXu1HR01K4FJg8h1kEEDAqDcZQtbrRnB41A=="],
 
-    "@esbuild/linux-x64": ["@esbuild/linux-x64@0.27.7", "", { "os": "linux", "cpu": "x64" }, "sha512-hzznmADPt+OmsYzw1EE33ccA+HPdIqiCRq7cQeL1Jlq2gb1+OyWBkMCrYGBJ+sxVzve2ZJEVeePbLM2iEIZSxA=="],
+    "@esbuild/linux-x64": ["@esbuild/linux-x64@0.21.5", "", { "os": "linux", "cpu": "x64" }, "sha512-1rYdTpyv03iycF1+BhzrzQJCdOuAOtaqHTWJZCWvijKD2N5Xu0TtVC8/+1faWqcP9iBCWOmjmhoH94dH82BxPQ=="],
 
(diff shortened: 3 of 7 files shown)


NikolayS:
## REV Code Review Report

- **PR:** [NikolayS/pgque#162](https://github.com/NikolayS/pgque/pull/162) — chore(clients/typescript): standardize on bun
- **Author:** @NikolayS
- **Branch:** `chore/typescript-bun`
- **CI:** ✅ all checks passed (test PG14-18, client-smoke, verify, claude-review)

---

### BLOCKING ISSUES (2)

**HIGH** `clients/typescript/README.md:11` — Install command changed to Bun-only, excluding npm/yarn/pnpm users
> The README now shows `bun add pgque` as the sole install instruction. The package is published to npm and declares `"engines": {"node": ">=20"}` — consumers using npm, yarn, or pnpm have no documented path.
> **Fix:** Show `npm install pgque` as the primary command (or alongside bun). If Bun-only is intentional, update `engines` in package.json to `{"bun": ">=1.3"}` and drop the Node.js engine requirement.

**MEDIUM** `clients/typescript/README.md:14` — Node.js requirement line contradicts the Bun-centric install instruction
> Line 14 reads "Requires Node.js 20+ and PostgreSQL 14+" immediately after `bun add pgque`. Bun is not a Node.js version — running under Bun does not require Node.js to be installed. Creates confusing mixed signals for new users.
> **Fix:** Update to "Requires Bun 1.x+ (or Node.js 20+) and PostgreSQL 14+", or revert the install command to `npm install pgque` and keep the existing requirement line.

---

### NON-BLOCKING (1)

**LOW** `.github/workflows/ci.yml:133` — `bunx tsx src/smoke.ts` adds unnecessary wrapper; Bun runs TypeScript natively
> After switching to Bun, invoking `bunx tsx` adds a second runtime layer (esbuild via tsx) over Bun's own native TypeScript execution. This works but wastes CI time and obscures which runtime actually executes the test.
> **Suggestion:** Use `bun run src/smoke.ts` directly. This also removes the need for `tsx` as a devDependency if it is not used elsewhere.

---

### POTENTIAL ISSUES (5)

**MEDIUM** `clients/typescript/README.md:109` — `bun run test` invokes vitest under Bun runtime, which is not officially supported *(confidence: 7/10)*
> The README test command is now `bun run test` → `vitest run`. Bun uses itself as the JS runtime for package scripts, and vitest officially targets Node.js — known compatibility gaps exist around worker threads and module resolution under Bun. CI only runs the smoke script (`bunx tsx src/smoke.ts`), not the full test suite, so this path is unvalidated in CI.
> **Suggestion:** Add a `bun run test` step to CI to validate this path, or document that `npm test` remains the supported contributor test command.

**LOW** `clients/typescript/src/smoke.ts:1` — Shebang `#!/usr/bin/env node` inconsistent with `bunx tsx` usage after migration *(confidence: 7/10)*
> The usage comment was updated to `bunx tsx src/smoke.ts`, but if the file retains a `#!/usr/bin/env node` shebang it now misleads editors, linters, and contributors about the intended runtime.
> **Suggestion:** Change shebang to `#!/usr/bin/env bun`, or drop it since the file is always invoked via `bunx tsx`.

**LOW** `.github/workflows/ci.yml:123` — No Bun install cache configured; packages re-downloaded every CI run *(confidence: 6/10)*
> `setup-bun` is added but without `cache: true` or an `actions/cache` step for `~/.bun/install/cache`. Previously `setup-node` provided implicit npm caching. This is a performance regression for CI.
> **Suggestion:** Add `cache: true` to the `oven-sh/setup-bun@v2` step, or add `actions/cache` for `~/.bun/install/cache`.

**LOW** `clients/typescript/package.json:N/A` — `engines.node: ">=20"` retained but CI no longer sets up Node *(confidence: 6/10)*
> `package.json` still declares `"engines": {"node": ">=20"}`, but the CI TypeScript job now only runs `setup-bun` with no `setup-node`. If any tooling falls back to Node (e.g. `tsc`), it will use whatever Node version the GitHub runner image happens to provide.
> **Suggestion:** Either add `setup-node` back to the CI job or update `engines` to reflect the actual requirement.

**LOW** `.github/workflows/ci.yml:123` — `Setup Bun` step placed mid-job, between Go and TypeScript smoke steps *(confidence: 5/10)*
> All other environment setup (Python pip, Go module cache) is done inline before test steps. The Bun setup appears between the Go smoke test and the TypeScript smoke step, breaking the visual pattern and making the job harder to reason about.
> **Suggestion:** Move `Setup Bun` to the top of the job alongside other tool setups, before any language-specific smoke tests.

---

### Summary

| CI | 0 | 0 | — |
| Security | 0 | 0 | 0 |
| Bugs | 1 | 3 | 0 |
| Tests | 0 | 0 | 0 |
| Guidelines | 0 | 1 | 0 |
| Docs | 2 | 1 | 0 |

---

NikolayS:
## REV Code Review Report

- **PR:** [NikolayS/pgque#162](https://github.com/NikolayS/pgque/pull/162) — chore(clients/typescript): standardize on bun
- **Author:** @NikolayS
- **Branch:** `chore/typescript-bun`
- **CI:** ⏳ running (verify ✅, client-smoke + test matrix in progress)

---

### BLOCKING ISSUES (1)

**HIGH** `.github/workflows/ci.yml:124` — `oven-sh/setup-bun@v2` is pinned to a mutable tag, not a commit SHA

> This PR introduces the only third-party (non-`actions/*`) action in the workflow. All existing `uses:` lines reference `actions/checkout@v4`, a first-party GitHub action with much lower supply-chain risk. A mutable tag on a third-party action means a force-push by the action author — malicious or accidental — silently executes new code with full access to runner secrets and the checkout.
>
> **Fix:** Pin to the SHA resolved for the current `v2` release:
> ```yaml
> uses: oven-sh/setup-bun@0c5077e51419868618aeaa5fe8019c62421857d6  # v2
> ```
> Then add a Dependabot `github-actions` config or use `pin-github-actions` to keep it current automatically.

---

### NON-BLOCKING (1)

**LOW** `.github/workflows/ci.yml:128-133` — TypeScript smoke step never exercises a live DB (pre-existing gap)

> `PGQUE_TEST_DSN` is never set in the `client-smoke` job, so `bunx tsx src/smoke.ts` exits 0 immediately (line 17-21 of smoke.ts: `if (!dsn) { process.exit(0) }`). The job already starts a Postgres 18 container and installs the schema — the infrastructure is there. Go smoke hardcodes a fallback DSN and actually connects. `bun run test` (vitest) is also absent from the step, so the integration test suite in `test/` is never run in CI.
>
> **Suggestion:** Add to the TypeScript smoke step:
> ```yaml
> env:
>   PGQUE_TEST_DSN: postgres://postgres:pgque_test@localhost/pgque_test
> ```
> and add `bun run test` before `bunx tsx src/smoke.ts`. This is out of scope for this PR but the step is being actively touched — ideal time to close the gap.

---

### POTENTIAL ISSUES (5)

**MEDIUM** `clients/typescript/package.json:6` — `"packageManager": "bun@1.3.8"` may create Corepack friction for contributors *(confidence: 5/10)*

> The `packageManager` field is enforced by Node.js Corepack. Contributors who `git clone` the repo and then run `npm install` or `yarn install` inside `clients/typescript/` will get a Corepack error demanding Bun. This is correct behavior for enforcing Bun, but it's a silent footgun for anyone not expecting it. The README's Install section does say to use Bun for dev, but only after this PR.
>
> **Suggestion:** Add a CONTRIBUTING note (or README dev-setup line) explaining the Corepack behavior, or add an `.npmrc` with `engine-strict=true` to give a clearer error message.

**LOW** `clients/typescript/tsconfig.json:15` — `"types": ["node"]` creates a fragile invariant *(confidence: 6/10)*

> Adding `"types": ["node"]` to the root `tsconfig.json` (inherited by `tsconfig.test.json`) restricts global ambient types to `@types/node` only. Currently safe because `vitest.config.ts` has `globals: false` and all vitest symbols are explicitly imported. If someone flips `globals: true` in the vitest config, type-checking for tests silently breaks — vitest ambient types (`describe`, `it`, `expect`) would not be found globally.
>
> **Suggestion:** Scope `"types": ["node"]` to `tsconfig.test.json` only (or add a comment explaining the `globals: false` dependency), so the invariant is documented and not invisible.

**LOW** `clients/typescript/src/smoke.ts:1` — shebang `#!/usr/bin/env node` is inconsistent with the Bun migration *(confidence: 7/10)*

> The shebang was not updated to match the new invocation (`bunx tsx src/smoke.ts`). Functionally harmless since the script is never executed directly, but misleads contributors who read the file header and try `./src/smoke.ts`.
>
> **Suggestion:** Remove the shebang or change it to `#!/usr/bin/env -S bunx tsx`.

**LOW** `clients/typescript/README.md:20-22` — Bun dev-tool note is placed in the end-user Install section *(confidence: 7/10)*

> "This repository uses Bun for TypeScript client development and CI commands." appears immediately after the Node.js runtime requirement in the Install section. A first-time package consumer reads this and may conclude they need Bun to use `pgque` as a library. Moving this to the Tests/Contributing section (which is already contributor-facing) would remove the ambiguity.
>
> **Suggestion:** Relocate the sentence to the Tests section or a separate "Contributing" subsection.

**LOW** `clients/typescript/README.md:117` — `bun run test` vs `bun test` not distinguished *(confidence: 6/10)*

> Bun ships two different test runners: `bun test` (Bun's native runner) and `bun run test` (runs the `test` npm script, here `vitest run`). They are not interchangeable. A contributor reading the README and using `bun test` instead of `bun run test` would get unexpected results.
>
> **Suggestion:** Add a one-line clarification: `bun run test  # runs vitest — 'bun test' uses a different runner and won't work here`.

---

### Summary

| Security | 1 | 0 | 1 |
| Bugs | 0 | 1 | 2 |
| Tests | 1 | 1 | 1 |
| Guidelines | 0 | 2 | 1 |
| Docs | 0 | 2 | 1 |

---

NikolayS:
## REV Code Review Report

- **PR:** [NikolayS/pgque#162](https://github.com/NikolayS/pgque/pull/162) — chore(clients/typescript): standardize on bun
- **Author:** @NikolayS
- **Branch:** `chore/typescript-bun`
- **CI:** ⏳ running (`test (14-18)`, `client-smoke`, `claude-review` in progress; `verify` ✅)

---

### BLOCKING ISSUES (3)

**HIGH** `.github/workflows/ci.yml:130-135` — vitest test suite never invoked in CI
> The CI `TypeScript smoke` step runs `bun install --frozen-lockfile`, `bun run check`, and `bunx tsx src/smoke.ts` — but never `bun run test`. Now that `PGQUE_TEST_DSN` is set at the job level (line 83), the full vitest integration suite (`test/client.test.ts`, `test/consumer.test.ts`, `test/nack.test.ts`) would run successfully, but it is never triggered. Nack routing, Consumer dispatch, abort behavior, concurrent producers, and error mapping are not CI-gated despite a live database being available.
> **Fix:** Add `bun run test` to the TypeScript smoke step, after `bun run check` and before `bunx tsx src/smoke.ts`.

**HIGH** `clients/typescript/src/smoke.ts:36-37` — `forceTick` + `ticker` pair added without explanation; `forceTick` alone is insufficient
> `await client.forceTick(queue)` only bumps the event-seq threshold; it does not insert a tick. A subsequent `ticker()` call is required to actually rotate the batch, and `ticker()` can return without producing a tick if timing conditions aren't met (though a fresh queue always forces it). `test/helpers.ts::advanceQueue()` already encapsulates this exact two-step idiom with an explanatory comment; smoke.ts now duplicates it silently.
> **Fix:** Add an inline comment: `// forceTick bumps the seq threshold; ticker() produces the batch boundary (mirrors advanceQueue in test/helpers.ts)`. The sequence is correct for a fresh queue but fragile if queue state changes.

**MEDIUM** `clients/typescript/src/smoke.ts:37` — PR description claims "No batch API changes" but adds a functional call
> The PR description states: *"No batch API changes are included here."* Yet the diff adds `await client.ticker(queue)` — a functional, behavior-changing call to a public API method. This mismatch misleads reviewers into treating the change as cosmetic. The previous smoke test would fail silently on `receive()` without this call, so its addition is important and should be explained.
> **Fix:** Amend the PR description to note that a smoke test correctness fix (adding the required `ticker()` call after `forceTick()`) is included.

---

### POTENTIAL ISSUES (8)

**HIGH** `clients/typescript/src/client.ts:231` — stale JSDoc *(pre-existing; confidence: 9/10)*
> `ticker()` JSDoc says *"Returns the underlying tick result count"* but the method signature is `Promise<void>` and the SQL result is discarded.
> **Suggestion:** Remove "Returns the underlying tick result count." from the JSDoc — pre-existing issue surfaced during this review.

**HIGH** `clients/typescript/README.md:57` — `ticker` and `forceTick` absent from API table *(pre-existing; confidence: 9/10)*
> Both are public `Client` methods; neither appears in the `## API` section.
> **Suggestion:** Add rows for both, or annotate them `@internal` with a JSDoc tag to clarify test-only intent.

**MEDIUM** `clients/typescript/src/smoke.ts:1` — shebang `#!/usr/bin/env node` after Bun standardization *(confidence: 7/10)*
> The shebang still reads `#!/usr/bin/env node`. CI invokes the file via `bunx tsx src/smoke.ts` so the shebang is irrelevant in practice, but it contradicts the Bun standardization this PR delivers.
> **Suggestion:** Change to `#!/usr/bin/env bun` or remove it; the `bunx tsx` invocation in the usage comment is authoritative.

**MEDIUM** `.github/workflows/ci.yml:83` — `PGQUE_TEST_DSN` hardcoded activates mandatory integration tests *(confidence: 8/10)*
> Previously the TypeScript smoke test exited 0 immediately when `PGQUE_TEST_DSN` was unset (opt-in). Now it is hardcoded at the job level, making integration tests mandatory in CI. Any future flakiness in `client.test.ts`, `consumer.test.ts`, or `nack.test.ts` will block the entire `client-smoke` job. Also: the pre-created `smoke_ts` queue in the CI setup step is now orphaned — the TS test creates `smoke_ts_<timestamp>` dynamically.
> **Suggestion:** Document this behavioral shift in the PR description; consider removing `smoke_ts` from the CI queue-setup step.

**MEDIUM** `clients/typescript/README.md:114` — DSN scheme inconsistency *(confidence: 6/10)*
> README example uses `postgres://postgres:pgque_test@localhost/pgque_test`; CI env uses `postgresql://...`. Both are accepted by libpq but the inconsistency is confusing.
> **Suggestion:** Align README example to `postgresql://` or add a parenthetical noting both schemes work.

**MEDIUM** `clients/typescript/package.json:6` — `packageManager` field lacks Corepack integrity hash *(confidence: 7/10)*
> `"packageManager": "bun@1.3.8"` — Corepack's strict mode expects a SHA-512 hash (e.g. `bun@1.3.8+sha512.<hash>`). Without it, Corepack cannot verify the binary.
> **Suggestion:** Run `corepack use bun@1.3.8` to generate the canonical string with the hash included.

**HIGH** `.github/workflows/ci.yml:18,85,144` — `actions/checkout` pinned to tag, not SHA *(pre-existing; confidence: 9/10)*
> `uses: actions/checkout@v4` appears 3 times. This PR correctly pins `oven-sh/setup-bun` to a commit SHA but leaves `actions/checkout` on a mutable tag, creating an inconsistency in supply chain hardening.
> **Suggestion:** Pin to SHA: `actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4` — pre-existing issue, but worth addressing in a follow-up.

**LOW** `.github/workflows/ci.yml:83` — hardcoded test DSN (not a secret risk) *(confidence: 6/10)*
> `PGQUE_TEST_DSN: postgresql://postgres:pgque_test@localhost/pgque_test` — credentials are ephemeral, localhost-only, and well-known test values. Not a security concern, but inconsistent with secrets hygiene conventions.
> **Suggestion:** Acceptable as-is; alternatively move to `${{ secrets.PGQUE_TEST_DSN }}` for consistency.

---

### Summary

| CI | 1 | 2 | 0 |
| Security | 0 | 1 | 1 |
| Bugs | 1 | 1 | 0 |
| Tests | 1 | 1 | 0 |
| Guidelines | 0 | 3 | 1 |
| Docs | 1 | 3 | 0 |

---

NikolayS:
## REV Code Review Report

- **PR:** [NikolayS/pgque#162](https://github.com/NikolayS/pgque/pull/162) — chore(clients/typescript): standardize on bun
- **Author:** @NikolayS
- **Branch:** `chore/typescript-bun`
- **CI:** ⏳ running (test 14–18 ✅, verify ✅, client-smoke in progress)

---

### BLOCKING ISSUES (1)

**MEDIUM** `clients/typescript/src/smoke.ts:1` — Shebang contradicts documented invocation and CI step
> Changed from `#!/usr/bin/env node` to `#!/usr/bin/env bun`, but line 8 comment and the CI step both use `bunx tsx src/smoke.ts`. Two different execution paths: bun-native TS runner (shebang) vs tsx-on-Node (invocation). The shebang is dead code in CI.
> **Fix:** Pick one path. Either (a) change the CI step and comment to `bun src/smoke.ts` (bun runs TS natively, no tsx needed), or (b) restore the shebang to `#!/usr/bin/env -S bunx tsx` if tsx is intentional. The current state means `./src/smoke.ts` and `bunx tsx src/smoke.ts` silently use different runtimes.

---

### NON-BLOCKING (2)

**LOW** `clients/typescript/README.md:116` — DSN scheme mismatch between README and CI
> README shows `postgres://postgres:pgque_test@localhost/pgque_test`; CI `env:` block uses `postgresql://postgres:pgque_test@localhost/pgque_test`. Both work, but inconsistency confuses contributors copying from either source.
> **Suggestion:** Standardize on `postgresql://` (the RFC URI scheme, used in CI).

**LOW** `clients/typescript/README.md:114` — Tests section shows `bun run test` with no pointer to Bun install
> A contributor hitting this section for the first time has no indication they need Bun (version 1.3.8 per `package.json`) before the command works.
> **Suggestion:** Add a note: "Requires Bun (`packageManager` field pins 1.3.8; install at bun.sh)." before the code block.

---

### POTENTIAL ISSUES (4)

**MEDIUM** `clients/typescript/README.md:20` — Bun runtime requirement may confuse end users *(confidence: 7/10)*
> "Runtime requirements: Node.js 20+…This repository uses Bun for TypeScript client development and CI commands" is in the same paragraph, which can read as though Bun is also needed to use the published npm package.
> **Suggestion:** Split into two sentences that explicitly state the npm package runs on Node.js only; Bun is contributor/CI tooling.

**LOW** `clients/typescript/README.md` — `forceTick()` and `ticker()` are undocumented in the API table *(confidence: 7/10)*
> Both are public `Client` methods exercised in smoke.ts, and this PR adds an explanatory comment for them — but neither appears in the API reference table in the README.
> **Suggestion:** Add rows for `client.forceTick(queue)` and `client.ticker(queue?)` to the API table.

**LOW** `clients/typescript/package.json:6` — `packageManager` field lacks integrity hash *(confidence: 7/10)*
> `"packageManager": "bun@1.3.8"` without a `+sha224…` suffix. Corepack spec requires the hash for verified installs; without it, corepack silently skips verification. Low practical impact since Bun does not use Corepack, but worth noting for spec compliance.
> **Suggestion:** Add the integrity hash or document the omission is intentional.

**LOW** `.github/workflows/ci.yml` — No CI job exercises vitest without `PGQUE_TEST_DSN` *(confidence: 6/10)*
> `PGQUE_TEST_DSN` is set at the job level, so all DB-gated tests always run in this job. There is no path that verifies the unconditional `PgqueConnectionError` test passes standalone (i.e., without a live DB).
> **Suggestion:** Add a step (or separate verify job) that runs `bun run test` without the env var to confirm gating logic and non-DB tests are healthy.

---

### Summary

| CI | — | — | — |
| Security | 0 | 0 | 0 |
| Bugs | 1 | 0 | 0 |
| Tests | 0 | 1 | 0 |
| Guidelines | 1 | 2 | 0 |
| Docs | 1 | 1 | 0 |

---

NikolayS:
## REV Code Review Report

- **PR:** [NikolayS/pgque#162](https://github.com/NikolayS/pgque/pull/162) — chore(clients/typescript): standardize on bun
- **Author:** @NikolayS
- **Branch:** `chore/typescript-bun`
- **CI:** ⏳ running (`verify` ✓, `test (14-18)` + `client-smoke` in progress)

---

### BLOCKING ISSUES (2)

**HIGH** `clients/typescript/README.md:20` — Runtime requirements paragraph conflates application user and contributor audiences
> The sentence "This repository uses Bun for TypeScript client development and CI commands." is appended to the runtime requirements block (aimed at package consumers), immediately after "Node.js 20+". A new contributor reading the `## Install` section cannot tell whether Bun is required to *use* the package or only to *develop* it — exactly the confusion the new intro paragraph was meant to prevent.
> **Fix:** Move the Bun sentence to the `## Tests` section (where Bun is already described) and leave `## Install`'s runtime line focused purely on application users: "Runtime requirements: Node.js 20+ and PostgreSQL 14+."

**MEDIUM** `clients/typescript/src/smoke.ts:36` — Comment mischaracterises what `forceTick()` does
> `// forceTick lowers the per-queue threshold` — `force_tick` does not lower a threshold. It calls `setval(queue_event_seq, nextval(...) + queue_ticker_max_count * 2 + 1000)`, bumping the event-seq counter far *above* `queue_ticker_max_count` so that `ticker()` sees a "full" batch and creates a tick immediately.
> **Fix:** Replace with: `// forceTick bumps the event-seq counter past queue_ticker_max_count,` / `// making ticker() treat the queue as full and produce a tick immediately.` (The existing JSDoc in client.ts already says "Bumps the event-seq threshold" — align with that.)

---

### POTENTIAL ISSUES (8)

**MEDIUM** `clients/typescript/package.json` — `tsx` remains in devDependencies after migration to Bun *(confidence: 7/10)*
> CI now uses `bun run check` + `bun src/smoke.ts`; `npx tsx` is gone. `tsx@^4.19.3` is a dead dependency.
> **Suggestion:** Remove `tsx` from devDependencies and regenerate `bun.lock`.

**MEDIUM** `clients/typescript/src/smoke.ts:39` — No dedicated test for the `forceTick()` + `ticker()` two-call contract *(confidence: 7/10)*
> The comment documents that both calls are required, but the only place this sequence is exercised with real assertions is `test/helpers.ts:advanceQueue` (a test helper, not a test case). A broken `ticker()` with a valid `forceTick()` (or vice versa) would not be caught by any named test.
> **Suggestion:** Add a test in `client.test.ts` asserting `receive()` returns 0 after `send()` alone, and 0 after only `forceTick()`, and 1 after the full sequence.

**MEDIUM** `clients/typescript/README.md:112` — Tests section pins Bun 1.3.8 without an install pointer *(confidence: 7/10)*
> A first-time contributor reading "The repository standardizes on Bun 1.3.8" has no guidance on how to install that specific version.
> **Suggestion:** Add a pointer: "Install via `curl -fsSL https://bun.sh/install | bash` or `mise use bun@1.3.8`; the version is also enforced by `packageManager` in `package.json`."

**MEDIUM** `clients/typescript/src/smoke.ts:52` — Smoke test does not assert the queue is empty after `ack` *(confidence: 6/10)*
> After `ack(msg.batchId)`, the test prints "OK" immediately. A silent double-delivery bug would not be detected.
> **Suggestion:** Add a second `receive()` call and assert the result is empty, mirroring the pattern already used in `client.test.ts`.

**LOW** `clients/typescript/tsconfig.json:15` — `"types": ["node"]` may conflict with the Bun runtime *(confidence: 6/10)*
> The package now runs under Bun, which ships its own type definitions. Pinning only `@types/node` could suppress Bun-specific globals or mask type mismatches.
> **Suggestion:** If `@types/node` is needed for the `pg` driver, add a comment explaining why. Otherwise consider removing the explicit override and letting Bun resolve types naturally.

**LOW** `clients/typescript/README.md:109` — `bun install --frozen-lockfile` appears without context *(confidence: 6/10)*
> A contributor might run `bun install` (without `--frozen-lockfile`) and unknowingly update `bun.lock` during a routine test run.
> **Suggestion:** Add a note: "Use `--frozen-lockfile` for reproducible installs; omit it only when intentionally updating dependencies."

**LOW** `clients/typescript/README.md:59` — `forceTick` and `ticker` are not in the API reference table *(confidence: 5/10)*
> Both are public client methods now explicitly called in smoke.ts, but the API table ends at `client.close()`.
> **Suggestion:** Add rows for `forceTick(queue)` and `ticker(queue?)`, marked as testing/low-level APIs.

**LOW** `clients/typescript/package.json` — Bun version is hardcoded in three places *(confidence: 4/10)*
> `ci.yml` (`bun-version: 1.3.8`), `package.json` (`"packageManager": "bun@1.3.8"`), and `README.md` ("Bun 1.3.8") must all be updated on a Bun version bump.
> **Suggestion:** Reference `package.json`'s `packageManager` field as the canonical source in README prose.

---

### Summary

| Area | Blocking | Potential | Filtered |
|------|----------|-----------|---------|
| CI | — | — | — |
| Security | 0 | 0 | 1 |
| Bugs | 0 | 0 | 0 |
| Tests | 0 | 3 | 1 |
| Guidelines | 0 | 2 | 0 |
| Docs | 2 | 3 | 0 |

---

NikolayS:
## REV Code Review Report

- **PR:** [NikolayS/pgque#162](https://github.com/NikolayS/pgque/pull/162) — chore(clients/typescript): standardize on bun
- **Author:** @NikolayS
- **Branch:** `chore/typescript-bun`
- **CI:** ⏳ running (verify ✅ — test/14–18 + client-smoke in progress)

---

### NON-BLOCKING (2)

**LOW** `clients/typescript/package.json` — `tsx` dev dependency is now dead

> After this PR, no script in `package.json`, no CI step, and no source file invokes `tsx`. Old path was `npx tsx src/smoke.ts`; new path is `bun src/smoke.ts` (Bun runs TypeScript natively). `tsx` remains pinned in `devDependencies` and locked in `bun.lock`, adding unnecessary install weight and supply chain surface.
>
> **Fix:** Remove `tsx` from `devDependencies` and regenerate `bun.lock`.

---

**LOW** `clients/typescript/README.md:111` — Bun version duplicated in prose creates a maintenance trap

> `README.md` says "The repository standardizes on Bun 1.3.8" while `ci.yml` independently pins `bun-version: 1.3.8`. When CI is bumped, the README silently diverges. `ci.yml` is already the authoritative source.
>
> **Fix:** Drop the version number from the README sentence — "The repository standardizes on Bun for TypeScript client development" — and let `ci.yml` be the single pin.

---

### POTENTIAL ISSUES (3)

**LOW** `.github/workflows/ci.yml:124` — SHA comment says `# v2` but SHA resolves to `v2.2.0` *(confidence: 7/10)*

> `oven-sh/setup-bun@0c5077e51419868618aeaa5fe8019c62421857d6 # v2` — the floating `# v2` comment implies the SHA tracks the `v2` tag (which moves), whereas the SHA is the immutable `v2.2.0` commit. When someone bumps the SHA in future they may write `# v2` again and not notice the version changed, undermining the auditability that SHA pinning provides.
>
> **Suggestion:** Change comment to `# v2.2.0`.

---

**INFO** PR description vs code mismatch *(confidence: 9/10)*

> The "After" block in the PR body shows `bunx tsx src/smoke.ts`, but the committed CI step runs `bun src/smoke.ts`. These take different execution paths (tsx wrapper vs. Bun's native TS runtime). No runtime regression expected, but the description is misleading for future `git log` readers.
>
> **Suggestion:** Update the PR description to match the actual command.

---

**LOW** `clients/typescript/package.json` — Bun pinned to 1.3.8; current stable is 1.3.13 *(confidence: 5/10)*

> Bun 1.3.9–1.3.13 include bug fixes. Both the `packageManager` field and `bun-version` in CI pin 1.3.8 (consistently). Worth checking the 1.3.9–1.3.13 changelog for anything relevant before this lands.
>
> **Suggestion:** Evaluate upgrading to 1.3.13; keep `packageManager` and `bun-version` in sync.

---

### Summary

| CI | 0 | 1 | 0 |
| Security | 0 | 1 | 1 |
| Bugs | 1 | 0 | 1 |
| Tests | 0 | 0 | 0 |
| Guidelines | 1 | 1 | 2 |
| Docs | 0 | 1 | 0 |

The `forceTick()` + `ticker()` fix in `smoke.ts` is correct per PgQ semantics and the comment explaining it is appropriate. The `"types": ["node"]` tsconfig addition is safe — `@types/pg` is module-based (no globals) and unaffected by the explicit `types` array. Overall a clean PR; the two non-blocking items (`tsx` removal and README version depin) are the ones most worth addressing before merge.

---

NikolayS:
## REV Code Review Report

- **PR:** [NikolayS/pgque#162](https://github.com/NikolayS/pgque/pull/162) — chore(clients/typescript): standardize on bun
- **Author:** @NikolayS
- **Branch:** `chore/typescript-bun`
- **CI:** ⏳ running (all checks in progress)

---

### BLOCKING ISSUES (1)

**MEDIUM** `clients/typescript/src/client.ts` — JSDoc for `ticker()` says "Returns the underlying tick result count" but the method signature is `Promise<void>`
> Pre-existing but surfaced by the new explicit `ticker()` call added in `smoke.ts`. The comment is a leftover from an earlier draft and directly contradicts the implementation.
> **Fix:** Remove "Returns the underlying tick result count." from the JSDoc (or replace with "Returns nothing; tick result is discarded.").

---

### POTENTIAL ISSUES (9)

**MEDIUM** `clients/typescript/README.md:109-111` — Dev tooling (Bun) vs runtime dependency distinction not explicit enough *(confidence: 7/10)*
> The Tests section says the repo "standardizes on Bun" but never explicitly states Bun is a *contributor/dev* requirement — application consumers still use Node.js. Readers installing with npm may be confused by a Tests section that only shows Bun commands.
> **Suggestion:** Add one sentence in the Tests section: "Node.js 20+ is the runtime requirement for application consumers; Bun is the development and CI tool for contributors."

**MEDIUM** `clients/typescript/test/client.test.ts` — `ticker()` and `forceTick()` only tested through `advanceQueue()` helper; no isolated tests for either method *(confidence: 7/10)*
> All DB-backed tests call `advanceQueue(env.client, env.queue)` which always issues both `forceTick` then `ticker`. If either method regresses in isolation (e.g. `ticker()` becomes a no-op without a prior `forceTick`), the suite would not catch it.
> **Suggestion:** Add a test that calls `ticker()` before `forceTick()` and asserts `receive()` returns empty, validating the two-step sequence is actually required.

**MEDIUM** `clients/typescript/test/client.test.ts` — No test for `client.ticker()` with no argument (global overload) *(confidence: 6/10)*
> `ticker(queue?: string)` has two code paths: per-queue `pgque.ticker($1)` and global no-arg `pgque.ticker()`. Only the per-queue path is exercised via `advanceQueue`.
> **Suggestion:** Add a test calling `client.ticker()` (no arg) against a live DB and asserting it doesn't throw.

**LOW** `clients/typescript/src/smoke.ts:1` — Shebang changed to `#!/usr/bin/env bun` but no indication the script is not runnable with `node` *(confidence: 7/10)*
> The README still says "Runtime requirements: Node.js 20+" — a contributor may try `node src/smoke.ts` and get a confusing TypeScript parse error.
> **Fix:** Add to usage comment: `//   (requires Bun; not directly runnable with node)`

**LOW** `clients/typescript/src/smoke.ts:36-38` — `forceTick`/`ticker` comment slightly mis-states `forceTick`'s role *(confidence: 6/10)*
> "bumps the event-seq threshold" is accurate but doesn't convey the causal dependency clearly — `forceTick` marks the queue dirty *so that* `ticker()` will produce a tick.
> **Fix:** Rephrase to: "forceTick marks the queue as having unseen events; ticker then creates the tick that makes those events visible to receive(). Both are required without an automatic ticker."

**LOW** `clients/typescript/tsconfig.json:15` — `"types": ["node"]` inconsistent with `#!/usr/bin/env bun` in `smoke.ts` *(confidence: 6/10)*
> Using `@types/node` is correct for the library's Node.js consumers but may mask Bun-specific API mismatches in the Bun-targeted smoke script.
> **Suggestion:** Acceptable for the library's `tsconfig.json` (it targets Node consumers). Consider a separate `tsconfig.smoke.json` with `bun-types` for the Bun-only script if type coverage there matters.

**LOW** `.github/workflows/ci.yml:82` — Hardcoded test credentials in workflow env block *(confidence: 5/10)*
> `PGQUE_TEST_DSN: postgresql://postgres:pgque_test@localhost/pgque_test` is acceptable for a local ephemeral CI DB, but there is no comment clarifying that.
> **Fix:** Add an inline comment: `# ephemeral local-only test instance — not a real secret`

**LOW** `.github/workflows/ci.yml:131-135` — All four TypeScript steps (`install`, `check`, `test`, `smoke`) run in a single `run:` block, making CI failure attribution ambiguous *(confidence: 5/10)*
> A vitest startup crash and an assertion failure produce identical CI step status. Splitting `bun run test` and `bun src/smoke.ts` into separate named steps would make failure triage faster.
> **Suggestion:** Split into two steps: `TypeScript tests` and `TypeScript smoke`.

**LOW** `clients/typescript/README.md:20` — `Runtime requirements` line omits Bun as a supported runtime *(confidence: 5/10)*
> `smoke.ts` now has `#!/usr/bin/env bun` and `packageManager` is set to Bun. Consumers using Bun as their app runtime are not covered by the stated requirements.
> **Fix:** Amend to: `Runtime requirements: Node.js 20+ (or Bun 1.3+) and PostgreSQL 14+`

---

### Summary

| Area       | Findings | Potential | Filtered |
|------------|----------|-----------|----------|
| CI         | —        | —         | —        |
| Security   | —        | 1         | 1        |
| Bugs       | —        | —         | —        |
| Tests      | —        | 3         | —        |
| Guidelines | —        | 3         | 1        |
| Docs       | 1        | 2         | —        |

---

NikolayS:
## REV Code Review Report

- **PR:** [NikolayS/pgque#162](https://github.com/NikolayS/pgque/pull/162) — chore(clients/typescript): standardize on bun
- **Author:** @NikolayS
- **Branch:** `chore/typescript-bun`
- **CI:** ⏳ running (test matrix + client-smoke in progress)

---

### POTENTIAL ISSUES (7)

**MEDIUM** `clients/typescript/README.md` — `ticker` and `forceTick` missing from API reference table *(confidence: 7/10)*
> Both methods are now exercised in the smoke test and explained in inline comments, but neither appears in the library's public API table. Consumers who discover them via autocomplete have no reference docs.
> **Suggestion:** Add rows to the API table for `client.ticker(queue?)` and `client.forceTick(queue)`, noting they are low-level/testing utilities.

**MEDIUM** `clients/typescript/src/smoke.ts` — No test asserts that `receive()` returns empty when `ticker()` is skipped *(confidence: 6/10)*
> The invariant "forceTick alone is insufficient; ticker() must follow" is now documented in a comment but is untested. A future refactor could silently break the ordering contract.
> **Suggestion:** Add an integration test that calls `send()` + `forceTick()` without `ticker()` and asserts `receive()` returns `[]`, then verifies visibility after `ticker()`.

**LOW** `.github/workflows/ci.yml:83` — `PGQUE_TEST_DSN` with plaintext password set at job-level `env` *(confidence: 6/10)*
> `postgresql://postgres:pgque_test@localhost/pgque_test` is visible to all steps and any third-party Actions in the job. Low actual risk (ephemeral CI credential), but the pattern is worth correcting before it's copied to a higher-stakes workflow.
> **Suggestion:** Move to a step-level `env` on the single step that needs it, or use a GitHub Actions secret.

**LOW** `clients/typescript/package.json:6` — `"packageManager": "bun@1.3.8"` is included in the published npm package *(confidence: 5/10)*
> This field signals corepack to enforce Bun for anyone running package manager commands inside this package's directory tree. Consumers with corepack enabled in a monorepo could see `npm install` / `pnpm install` rejected.
> **Suggestion:** Accept the risk (corepack only fires inside the package dir, not for dependents), or document the caveat in the README install section.

**LOW** `clients/typescript/README.md` — "Runtime requirements: Node.js 20+" conflates consumer runtime with Bun dev tooling *(confidence: 5/10)*
> The line appears directly after the multi-PM install block, which now lists `bun add pgque`. A reader can reasonably conclude Bun is also a production runtime requirement.
> **Suggestion:** Clarify that Node.js 20+ is the runtime requirement for end-users of the published package; Bun is the development/CI tooling requirement for contributors.

**LOW** `clients/typescript/src/client.ts:231` — `ticker` JSDoc doesn't mention its relationship to `forceTick` *(confidence: 5/10)*
> The JSDoc was corrected from "Returns the tick result count" to "Resolves after the ticker call completes" — accurate for `Promise<void>`, but the required sequencing (`forceTick` → `ticker`) that was explained in smoke.ts is absent from the library's own API docs.
> **Suggestion:** Add a note: "In manual/test mode, call `forceTick()` first to bump the event-seq threshold, then `ticker()` to make newly sent events visible to `receive()`."

**LOW** `clients/typescript/package.json` — `"engines": { "node": ">=20" }` vs `"packageManager": "bun@1.3.8"` sends mixed signals *(confidence: 4/10)*
> `engines.node` signals that Node.js 20+ is a valid runtime; `packageManager: bun` signals Bun is the required PM. These are not strictly contradictory (Bun can satisfy `engines.node` checks) but the combination may confuse tooling.
> **Suggestion:** Add `"bun": ">=1.3.8"` to `engines` if Bun is the intended development runtime, or document that `packageManager` is a contributor-only signal.

---

### Summary

| CI | 0 | — | — |
| Security | 0 | 1 | 0 |
| Bugs | 0 | 2 | 1 |
| Tests | 0 | 1 | 1 |
| Guidelines | 0 | 2 | 1 |
| Docs | 0 | 1 | 0 |

---

NikolayS:
Testing / review evidence for latest head `7693ca8422e435f69e139ee149d89cffde7ddb14`:

Local verification:

```bash
cd clients/typescript
bun install --frozen-lockfile
bun run check
bun run test
PGQUE_TEST_DSN=postgresql://postgres:pgque_test@localhost/pgque_batch_test bun src/smoke.ts
```

Result: all passed. Smoke output: `pgque TypeScript smoke test: OK`.

CI:

- `claude-review` ✅
- `client-smoke` ✅
- PG matrix `test (14)` ✅
- PG matrix `test (15)` ✅
- PG matrix `test (16)` ✅
- PG matrix `test (17)` ✅
- PG matrix `test (18)` ✅
- `verify` ✅

REV:

- REV run after latest changes: no blocking issues.
- Earlier REV findings fixed: npm-compatible install docs, Node/Bun wording, pinned setup-bun action, live TS smoke DSN, `bun run test` in CI, direct `bun src/smoke.ts`, `forceTick()` + `ticker()` comment, and stale `ticker()` JSDoc.

NikolayS:
## REV Code Review Report

- **PR:** [NikolayS/pgque#162](https://github.com/NikolayS/pgque/pull/162) — chore(clients/typescript): standardize on bun
- **Author:** @NikolayS
- **Branch:** `chore/typescript-bun`
- **CI:** ⏳ in_progress (`verify` passed; `test (14-18)`, `client-smoke`, `claude-review` still running)

---

### BLOCKING ISSUES (0)

None.

---

### NON-BLOCKING (1)

**LOW** `clients/typescript/src/smoke.ts:36` — 3-line comment block; CLAUDE.md asks for one short line max
> ```
> // forceTick bumps the event-seq threshold; ticker actually creates the tick
> //   that makes newly sent events visible to receive(). Both calls are required
> //   in manual/demo mode.
> ```
> **Suggestion:** Condense to one line: `// both calls required: forceTick bumps threshold, ticker makes events visible`

---

### POTENTIAL ISSUES (4)

**MEDIUM** `clients/typescript/tsconfig.json:15` — `"types": ["node"]` vs. Bun runtime mismatch *(confidence: 6/10)*
> Smoke script and CI now run under Bun (`#!/usr/bin/env bun`, `bun src/smoke.ts`), but `tsc --noEmit` only loads `@types/node`. Any future Bun-specific global (`Bun.file`, etc.) used in `src/` will pass type-checking but fail at runtime. Currently safe — `smoke.ts` only uses `process.env`/`process.exit` — but this is a latent footgun.
> **Suggestion:** Either add `bun-types` to the `types` array (+ devDependency) or document explicitly that `src/` must not use Bun-specific globals.

**MEDIUM** `clients/typescript/test/client.test.ts` — no-arg `client.ticker()` overload never exercised *(confidence: 6/10)*
> `ticker()` has two SQL paths: with-queue (`pgque.ticker(queue text)`) and no-arg (`pgque.ticker()`). The with-queue path is exercised indirectly via `advanceQueue()`. The no-arg path is never called in any test or in the smoke script — a regression there would only surface in production.
> **Suggestion:** Add one test calling `await client.ticker()` (no argument) after a queue exists, asserting it resolves without throwing.

**LOW** `.github/workflows/ci.yml:135` — vitest and smoke share the same DB without ordering guarantee *(confidence: 5/10)*
> Both `bun run test` (vitest) and `bun src/smoke.ts` target the same `PGQUE_TEST_DSN`. They run sequentially and use distinct queue names, so conflict is unlikely. However, a hard vitest failure leaving a queue in a half-created state could affect CI clarity (not correctness, since smoke uses a timestamped queue name).
> **Suggestion:** Low priority; acceptable as-is but worth watching in flaky-test triage.

**LOW** `clients/typescript/src/smoke.ts:24` — queue/consumer name collision under parallel CI *(confidence: 4/10)*
> `smoke_ts_${Date.now()}` uses millisecond granularity. Unlikely to collide, but the vitest helpers use `randomBytes(4)` for this exact reason.
> **Suggestion:** Add a short random suffix: `` `smoke_ts_${Date.now()}_${Math.random().toString(36).slice(2,6)}` ``

---

### Summary

| CI | 0 | 1 | 0 |
| Security | 0 | 0 | 3 |
| Bugs | 0 | 1 | 3 |
| Tests | 0 | 1 | 4 |
| Guidelines | 1 | 0 | 1 |
| Docs | 0 | 1 | 3 |

Overall: solid, well-scoped tooling migration. The `forceTick()` + `ticker()` fix in the smoke test is a meaningful correctness improvement. The SHA-pinned `setup-bun` action and `--frozen-lockfile` are good practices. No blocking issues.

---

NikolayS:
Final post-cleanup evidence for latest head `91b6b2a6459ec878783796a33e14eac63492b6d6`:

Small final improvements added:

- documented `client.forceTick(queue)` and `client.ticker(queue?)` in the TypeScript README API table
- changed setup-bun SHA comment to exact release: `# v2.2.0`

Local verification:

```bash
cd clients/typescript
bun run check
bun run test
PGQUE_TEST_DSN=postgresql://postgres:pgque_test@localhost/pgque_batch_test bun src/smoke.ts
```

Result: all passed. Smoke output: `pgque TypeScript smoke test: OK`.

CI:

- `claude-review` ✅
- `client-smoke` ✅
- PG matrix `test (14)` ✅
- PG matrix `test (15)` ✅
- PG matrix `test (16)` ✅
- PG matrix `test (17)` ✅
- PG matrix `test (18)` ✅
- `verify` ✅

REV:

- latest REV: no blocking issues
- comment: https://github.com/NikolayS/pgque/pull/162#issuecomment-4361858495

PR #168, opened 2026-05-02
fix(clients/python): bounded LISTEN/NOTIFY wait honors stop() and notify (#158)

### LINKED ISSUE #158: Python client: Consumer LISTEN/NOTIFY wait can hang and ignore stop
## Problem

Python `Consumer.start()` can hang in the LISTEN/NOTIFY wait loop and ignore `Consumer.stop()` / delayed notifications.

Current loop:

```python
gen = conn.notifies(timeout=self.poll_interval)
for _notify in gen:
    break
```

In adversarial testing with psycopg3, the thread remained alive past the requested timeout and `stop()` did not interrupt the wait promptly. A producer sending + ticking after the consumer had LISTENed also did not wake the handler within the expected window.

## Repro sketch

Against PR #155 / current Python Consumer shape:

```python
cons = pgque.Consumer(dsn, queue=q, name=c, poll_interval=10)
t = threading.Thread(target=cons.start, daemon=True)
t.start()
time.sleep(1)
cons.stop()
t.join(timeout=2)
assert not t.is_alive()  # fails / hangs in probe
```

A notify wake probe similarly hung until killed:

```python
# start consumer with poll_interval=10
# after LISTEN, send event, commit, force_tick(queue), ticker(queue), commit
# expected handler promptly; observed thread still alive/no handler call
```

## Why this matters

The Python Consumer advertises LISTEN/NOTIFY wakeup and graceful stop. A consumer loop that cannot be stopped promptly is operationally painful and makes tests flaky/hanging.

## Suggested fix

Use a bounded wait pattern that checks `_running` periodically and drains notifications without entering an uninterruptible generator iteration. For example, use psycopg's documented notify APIs with short timeouts, or `select.select()`/connection socket readiness with a small max wait slice.

Add tests proving:

- `stop()` returns promptly while waiting for notifications.
- a real `pg_notify('pgque_' || queue, ...)` wakes the consumer before `poll_interval` expires.

NikolayS:
## Summary

Fixes #158. The Python `Consumer.start()` LISTEN/NOTIFY wait used:

```python
gen = conn.notifies(timeout=self.poll_interval)
for _notify in gen:
    break
```

That generator blocks uninterruptibly inside libpq for up to the full
`poll_interval`, so:

- `stop()` is not honored until the timeout expires (10s with default
  `poll_interval=10`, longer with larger values).
- a producer's NOTIFY does not always wake the iteration within the
  expected window in adversarial timing.

## Root cause

`psycopg.Connection.notifies(timeout=N)` is implemented as a single
blocking read with no internal poll. Once entered, neither the GIL nor
a second thread setting `self._running = False` can interrupt it.

## Fix (before / after)

Replace the wait with `_wait_for_notify_or_stop()`, which loops on
`select.select([conn.fileno()], [], [], SLICE)` with `SLICE = 0.5s`,
re-checking `_running` between slices and draining notifications via
`conn.notifies(timeout=0)` when the socket is readable. Cumulative wait
is still capped at `poll_interval`.

Before: one uninterruptible blocking read up to `poll_interval`.
After: tight slice loop -- `stop()` is observed within ~0.5s, and a
NOTIFY wakes the consumer within ~0.5s of the server pushing it.

Smallest-possible diff: only the wait body changed; signal handling,
poll loop, dispatch, and ack/nack semantics are untouched.

## Evidence

New tests in `clients/python/tests/test_consumer_listen_stop.py`:

1. `test_stop_is_honored_within_two_seconds` -- starts the consumer with
   `poll_interval=10`, calls `stop()` after 1s, asserts thread joins
   within 2s.
2. `test_notify_wakes_consumer_before_poll_interval` -- starts the
   consumer with `poll_interval=10`, then sends + ticks from a separate
   connection, asserts the handler fires within 3s.

Red (without the fix, on this branch's first commit `18fbabe`):

```
FAILED tests/test_consumer_listen_stop.py::test_stop_is_honored_within_two_seconds
FAILED tests/test_consumer_listen_stop.py::test_notify_wakes_consumer_before_poll_interval
============================== 2 failed in 10.71s ==============================
```

Green (with the fix, on `38b294e`):

```
tests/test_consumer_listen_stop.py::test_stop_is_honored_within_two_seconds PASSED
tests/test_consumer_listen_stop.py::test_notify_wakes_consumer_before_poll_interval PASSED
============================== 2 passed in 2.21s ===============================
```

Full Python suite (live PG 16, `PGQUE_TEST_DSN` set):

```
============================= 43 passed in 16.41s ==============================
```

## Test plan

- [ ] Review `clients/python/pgque/consumer.py` -- specifically
      `_wait_for_notify_or_stop` and the `_WAIT_SLICE_SECONDS` constant.
- [ ] Confirm the new tests in `clients/python/tests/test_consumer_listen_stop.py`
      cover both axes (stop promptness + NOTIFY wakeup).
- [ ] Run `cd clients/python && PGQUE_TEST_DSN=... pytest -v tests/` --
      expect 43 passed.
- [ ] Sanity-check with a higher `poll_interval` (e.g. 60s) that
      `stop()` still returns within ~1s.

Closes #158.

---
_Generated by [Claude Code](https://claude.ai/code/session_01EUZEMeLZXuNvUCoUEgo7KB)_

### DIFF
--- a/clients/python/pgque/consumer.py
+++ b/clients/python/pgque/consumer.py
@@ -18,6 +18,10 @@ from .types import Message
 
 logger = logging.getLogger("pgque")
 
+# Maximum time the LISTEN wait blocks before re-checking the stop flag.
+# Bounds shutdown latency to roughly this many seconds. See issue #158.
+_WAIT_SLICE_SECONDS = 0.5
+
 
 class Consumer:
     """Synchronous polling consumer with LISTEN/NOTIFY wakeup.
@@ -134,15 +138,14 @@ class Consumer:
                     if not self._running:
                         break
 
-                    # Wait for NOTIFY or poll_interval timeout
-                    try:
-                        gen = conn.notifies(timeout=self.poll_interval)
-                        for _notify in gen:
-                            # Any notification means new events; break
-                            # to poll immediately.
-                            break
-                    except StopIteration:
-                        pass
+                    # Wait for NOTIFY or poll_interval timeout in short
+                    # bounded slices. psycopg's conn.notifies() can
+                    # block uninterruptibly for the full timeout, which
+                    # makes stop() slow and can miss prompt wakeups
+                    # (issue #158). Polling the underlying socket with
+                    # select() lets us re-check _running every SLICE
+                    # seconds and drain any pending NOTIFY immediately.
+                    self._wait_for_notify_or_stop(conn)
 
         finally:
             if in_main_thread:
@@ -154,6 +157,41 @@ class Consumer:
     def stop(self) -> None:
         """Request graceful shutdown (safe to call from another thread)."""
         self._running = False
+
+    def _wait_for_notify_or_stop(self, conn: psycopg.Connection) -> None:
+        """Wait up to ``poll_interval`` for a NOTIFY, in short slices.
+
+        Returns early on any of:
+          * a NOTIFY arrives (drained from the connection),
+          * ``stop()`` flips ``_running`` to False,
+          * ``poll_interval`` elapses cumulatively.
+
+        Each slice is at most ``_WAIT_SLICE_SECONDS`` so ``stop()`` is
+        observed within ~SLICE seconds of the call. Issue #158.
+        """
+        import time as _time
+
+        deadline = _time.monotonic() + self.poll_interval
+        fd = conn.fileno()
+        while self._running:
+            remaining = deadline - _time.monotonic()
+            if remaining <= 0:
+                return
+            slice_timeout = min(_WAIT_SLICE_SECONDS, remaining)
+            # select() returns when the socket is readable (notify
+            # delivered by the server) or when slice_timeout expires.
+            # It is a thin wrapper around the OS poll, so it is cheap
+            # and interruptible.
+            r, _w, _x = select.select([fd], [], [], slice_timeout)
+            if not self._running:
+                return
+            if r:
+                # Drain pending notifications without blocking. A
+                # zero timeout makes notifies() return immediately
+                # after consuming whatever is buffered.
+                for _notify in conn.notifies(timeout=0):
+                    pass
+                return
 
     def _poll_once(self, conn: psycopg.Connection) -> None:
         """Receive one batch and dispatch messages."""
--- a/clients/python/pgque/consumer.py
+++ b/clients/python/pgque/consumer.py
@@ -8,6 +8,7 @@ import logging
 import signal
 import select
 import threading
+import time
 from typing import Callable, Optional
 
 import psycopg
@@ -169,12 +170,22 @@ class Consumer:
         Each slice is at most ``_WAIT_SLICE_SECONDS`` so ``stop()`` is
         observed within ~SLICE seconds of the call. Issue #158.
         """
-        import time as _time
+        # Drain any NOTIFY already buffered in libpq from the prior
+        # _poll_once (e.g. delivered alongside query results). Without
+        # this, a buffered notify sits in libpq until the socket
+        # becomes readable for some other reason -- select() won't see
+        # it, and wakeup latency stretches out. Restores the implicit
+        # entry-drain semantics of the old conn.notifies(timeout=...).
+        drained = False
+        for _notify in conn.notifies(timeout=0):
+            drained = True
+        if drained:
+            return
 
-        deadline = _time.monotonic() + self.poll_interval
+        deadline = time.monotonic() + self.poll_interval
         fd = conn.fileno()
         while self._running:
-            remaining = deadline - _time.monotonic()
+            remaining = deadline - time.monotonic()
             if remaining <= 0:
                 return
             slice_timeout = min(_WAIT_SLICE_SECONDS, remaining)
--- a/clients/python/tests/test_consumer_listen_stop.py
+++ b/clients/python/tests/test_consumer_listen_stop.py
@@ -16,7 +16,7 @@ import pytest
 import pgque
 
 
-def test_stop_is_honored_within_two_seconds(dsn, setup_queue):
+def test_stop_is_honored_promptly(dsn, setup_queue):
     """`stop()` must interrupt the LISTEN/NOTIFY wait promptly.
 
     Regression for #158: the prior `for _notify in conn.notifies(timeout=...)`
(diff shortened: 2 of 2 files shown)


NikolayS (commented):
## REV Code Review Report

- **PR:** NikolayS/pgque#168 — fix(clients/python): bounded LISTEN/NOTIFY wait honors stop() and notify (#158)
- **Author:** @NikolayS
- **Branch:** `fix/python-consumer-listen-stop-158` → `main`
- **Verdict:** **COMMENT** (no must-fix; a few nice-to-haves)
- **SOC2 items:** skipped per maintainer policy.

### Tooling note

REV (https://gitlab.com/postgres-ai/rev) is GitLab-only by design — its `/review-mr` command depends on the `glab` CLI and validates `gitlab.com` URLs strictly. It cannot be pointed at a GitHub PR without modification. Rather than block on that, I ran REV's structured agent checklists (security-reviewer, bug-hunter, test-analyzer, guidelines-checker, docs-reviewer; sqitch-migration-checker is platform-only and N/A) against the diff manually and cross-checked against the actual files in the branch.

---

### BLOCKING ISSUES (0)

None.

---

### NON-BLOCKING (2)

**INFO** `clients/python/pgque/consumer.py:170` — `import time as _time` is done inside the function body, but `time` is not imported at module level. The aliasing isn't needed (no name collision with any other symbol in this file) and it's inconsistent with the file's existing style (`select`, `signal`, `threading` are all top-level imports). *(confidence: 8/10)*
> **Suggestion:** Add `import time` to the module-level imports near `import select` and replace `_time.monotonic()` with `time.monotonic()`. Minor style cleanup; not load-bearing.

**LOW** `clients/python/tests/test_consumer_listen_stop.py:38` — the `stop_elapsed < 2.0` assertion has narrow headroom. The fix's worst-case latency after `stop()` is ~`_WAIT_SLICE_SECONDS` (0.5s) plus whatever the in-flight `_poll_once` takes; 2.0s is comfortable on a healthy machine but can flake on slow CI runners or when the DB is contended. *(confidence: 6/10)*
> **Suggestion:** Either bump the threshold to ~3.0s (still well below the 10s `poll_interval` it's testing) or assert against `_WAIT_SLICE_SECONDS + small_constant` to make the bound's relationship to the implementation explicit. Optional.

---

### POTENTIAL ISSUES (1)

**LOW** `clients/python/pgque/consumer.py:174` — `fd = conn.fileno()` is captured once outside the inner loop. If the underlying psycopg connection is reset/reconnected mid-wait (e.g., server-side termination), the fd could in principle become stale. In practice the `with psycopg.connect(...)` context in `start()` holds the same socket for the consumer's lifetime, so this is unlikely to bite — but it's worth a comment or a per-iteration `conn.fileno()` if you want belt-and-braces. *(confidence: 4/10)*

---

### Test coverage

The two new regression tests exercise both axes called out in #158 (stop promptness + NOTIFY wakeup). The diff's behavior under the third return condition — cumulative `poll_interval` deadline elapsing without stop or NOTIFY — is exercised only indirectly via the outer `while self._running` loop. A small unit-style test (e.g. `poll_interval=1`, no NOTIFY, assert return ≤ 1.5s) would tighten coverage but is genuinely nice-to-have, not blocking.

### Sanity checks I ran beyond REV's checklists

- Verified `psycopg.Connection.fileno()` is a documented, stable API explicitly intended for use with `selectors` / `select` — the fix is idiomatic for psycopg3.
- Confirmed Python 3.5+ retries `select.select()` on EINTR per PEP 475, so signal delivery during a slice won't raise.
- Confirmed the diff is minimal and surgical: only the wait body changed; signal handling, the outer loop, `_poll_once`, and ack/nack semantics are untouched, matching the PR description's "smallest-possible diff" claim.
- Confirmed the new test file follows the existing `clients/python/tests/conftest.py` fixture conventions (`dsn`, `conn`, `setup_queue`).

### Summary

| Security | 0 | 0 | 0 |
| Bugs | 0 | 1 | 0 |
| Tests | 1 | 0 | 0 |
| Guidelines | 1 | 0 | 0 |
| Docs | 0 | 0 | 0 |

**Must-fix: 0. Nice-to-have: 2 (+1 potential).** The fix is correct, idiomatic, and well-tested. Ship-ready once the `import time` style nit is addressed (or deliberately deferred).

---
*REV-style review (agent checklists from [postgres-ai/rev](https://gitlab.com/postgres-ai/rev) applied manually since REV's CLI is GitLab-only). SOC2 items skipped per maintainer policy.*

---

NikolayS:
Follow-up after the second review pass — head is now `f2a3505`. Summary of changes + one explicit disagreement that warrants a maintainer call:

**Applied (commit `f2a3505`):**
- ✅ Entry-drain in `_wait_for_notify_or_stop`: drains pending notifies non-blockingly on entry and returns early if any were drained. Closes the bounded latency hole the reviewer flagged.
- ✅ `import time` hoisted to module scope; `_time.monotonic()` → `time.monotonic()`.
- ✅ Test renamed `test_stop_is_honored_within_two_seconds` → `test_stop_is_honored_promptly`.
- ✅ `fd = conn.fileno()` was already hoisted outside the slice loop.

**Did NOT do — reviewer's premise looks incorrect, want a maintainer ack before changing:**
The reviewer asked to restructure `test_notify_wakes_consumer_before_poll_interval` because they thought the producer's `send + commit` fires an insert-NOTIFY that races the tick commit. Engineer empirically tested both proposed restructures and both fail (consumer wakes to an empty batch — PgQ uses `pg_current_snapshot()` in tick inserts, so events committed in the same xact as the tick aren't visible in that tick's snapshot). They also `grep`d `sql/pgque.sql` and confirmed `pg_notify` is only called from inside `pgque.ticker()` — there is no insert-NOTIFY trigger. So the original two-commit ordering (send commit → no NOTIFY; tick commit → atomic NOTIFY+visibility) is already deterministic. The test now carries a comment block explaining this.

If the reviewer still wants a different test shape (e.g. add a third tick after send), please confirm the intent.

**CI:** re-running on `f2a3505`. Local full Python suite: `43 passed in 19.63s`.

---

NikolayS:
## Leo review — PR #168

Scope reviewed: `clients/python/pgque/consumer.py` and `clients/python/tests/test_consumer_listen_stop.py` at head `a6c2d0f`.

### BLOCKING

None found.

### NON-BLOCKING

None.

### POTENTIAL / watch later

- The wait loop uses `select.select()` on the psycopg connection fd, which is appropriate on Linux CI/runtime. If the Python client later promises first-class Windows support, this should be rechecked there. Not a blocker for current project/CI.

### Review notes

- The fix addresses the actual issue: `stop()` no longer waits for a full `poll_interval`; `_wait_for_notify_or_stop()` bounds stop latency by `_WAIT_SLICE_SECONDS` and exits early when a NOTIFY arrives.
- Entry-draining `conn.notifies(timeout=0)` before `select()` is important and present; otherwise already-buffered libpq notifications could be missed by fd readiness.
- Test coverage is targeted and useful:
  - stop promptness with `poll_interval=10`
  - NOTIFY wake before poll interval
- The producer/tick ordering in the NOTIFY test is correct for PgQ semantics: event commit first, then tick+notify commit. `pg_notify` call sites are only in ticker functions, not insert triggers.

### Validation

- GitHub CI on head `a6c2d0f`: all green, including new split client jobs.
- Local fresh PG18 full Python suite: `43 passed in 19.75s`.
- Local fresh PG18 targeted #158 tests, repeated 3x: all passed (`2 passed` each run).

Verdict: READY FOR MAINTAINER APPROVAL / MERGE.

PR #175, opened 2026-05-02
test: add receive two-session lock harness

alceops:
Adds an optional two-session `psql` harness for the same-consumer `pgque.receive()` race discussed in #125.

The existing SQL regression coverage is single-session, so this script opens two real backend sessions:

- session 1 receives one message and holds the transaction open
- session 2 concurrently calls `pgque.receive()` for the same `(queue, consumer)`
- the harness expects session 2 to wait on the row lock and return zero duplicate rows

I left it as an explicit script instead of wiring it into CI because it is meant as a red/green validator for the held #125 fix: current pre-fix code fails, while the row-lock fix should make it pass.

Verification run locally against current `main` install (expected red):

```text
bash -n tests/two_session_receive_lock.sh
PGQUE_TEST_DSN=postgresql://postgres:***@localhost/pgque_test tests/two_session_receive_lock.sh
# FAIL: two-session receive harness failed (session1=0, session2=3)
# session2 waited ~3.2s, then received 1 duplicate row
```

### DIFF
--- a/tests/two_session_receive_lock.sh
+++ b/tests/two_session_receive_lock.sh
@@ -1,0 +1,108 @@
+#!/usr/bin/env bash
+set -Eeuo pipefail
+
+# Validate same-consumer receive serialization with two real sessions.
+#
+# Usage:
+#   PGQUE_TEST_DSN=postgresql://postgres:***@localhost/pgque_test \
+#     tests/two_session_receive_lock.sh
+#
+# The target database must already have sql/pgque.sql installed. The harness
+# creates one temporary queue name, inserts one event, then proves that a second
+# concurrent pgque.receive(queue, consumer) call blocks behind the first session
+# and does not receive the same message while the first batch remains active.
+# It is intentionally useful as a red/green validator for the #97/#125 fix:
+# pre-fix code should fail by returning too quickly and/or duplicating the row;
+# the row-lock fix should make it wait and return zero duplicate rows.
+
+if [[ -z "${PGQUE_TEST_DSN:-}" ]]; then
+  echo "PGQUE_TEST_DSN is required" >&2
+  exit 2
+fi
+
+psql_base=(psql --no-psqlrc -v ON_ERROR_STOP=1 "${PGQUE_TEST_DSN}")
+queue_name="two_session_receive_${$}_$(date +%s)"
+workdir="$(mktemp -d)"
+cleanup() {
+  rm -rf "${workdir}"
+}
+trap cleanup EXIT
+
+cat >"${workdir}/setup.sql" <<SQL
+select pgque.create_queue('${queue_name}');
+select pgque.register_consumer('${queue_name}', 'c1');
+select pgque.insert_event('${queue_name}', 'test.concurrent', '{"n":1}');
+select pgque.force_tick('${queue_name}');
+select pgque.ticker();
+SQL
+
+cat >"${workdir}/session1.sql" <<SQL
+begin;
+create temp table s1_receive as
+  select * from pgque.receive('${queue_name}', 'c1', 10);
+do \$\$
+declare
+  v_count integer;
+begin
+  select count(*) into v_count from s1_receive;
+  assert v_count = 1, format('session1 expected 1 message, got %s', v_count);
+end \$\$;
+select pg_sleep(4);
+commit;
+SQL
+
+cat >"${workdir}/session2.sql" <<SQL
+\timing on
+begin;
+create temp table s2_receive as
+  select * from pgque.receive('${queue_name}', 'c1', 10);
+do \$\$
+declare
+  v_count integer;
+begin
+  select count(*) into v_count from s2_receive;
+  assert v_count = 0, format('session2 must not receive duplicate message, got %s', v_count);
+end \$\$;
+commit;
+SQL
+
+"${psql_base[@]}" -f "${workdir}/setup.sql"
+"${psql_base[@]}" -f "${workdir}/session1.sql" >"${workdir}/session1.out" 2>"${workdir}/session1.err" &
+session1_pid=$!
+
+print_debug() {
+  echo "--- session1.out ---" >&2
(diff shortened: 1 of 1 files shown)


tests/two_session_receive_lock.sh
```diff
@@ -0,0 +1,108 @@
...
+  echo "--- session1.err ---" >&2
+  cat "${workdir}/session1.err" >&2 || true
+  echo "--- session2.out ---" >&2
+  cat "${workdir}/session2.out" >&2 || true
+  echo "--- session2.err ---" >&2
+  cat "${workdir}/session2.err" >&2 || true
+}
+
+# Give session 1 enough time to enter receive() and hold its transaction open.
+sleep 1
```
NikolayS: can be racy on a slow runner; perhaps better to poll until session 1 is idle-in-tx or `sub_batch is not null`

smth like:

```suggestion
for i in $(seq 1 50); do
  "${psql_base[@]}" -tAc "
      select from pgque.subscription s
      join pgque.queue q on q.queue_id = s.sub_queue
      where q.queue_name = '${queue_name}' and s.sub_batch is not null
    " | grep -q 1 && break
  sleep 0.2
done
```

tests/two_session_receive_lock.sh
```diff
@@ -0,0 +1,108 @@
...
+\timing on
+begin;
+create temp table s2_receive as
+  select * from pgque.receive('${queue_name}', 'c1', 10);
+do \$\$
+declare
+  v_count integer;
+begin
+  select count(*) into v_count from s2_receive;
+  assert v_count = 0, format('session2 must not receive duplicate message, got %s', v_count);
```
NikolayS: I guess this assert is going to be changed once #125 lands.

thinking:

the harness is meant to be RED on current main (which it is — your run shows session 2 returns 1 row after waiting ~3.2s) and GREEN once a fix lands

there is a concern related to the GREEN target.

Both the next_batch_custom() early-return at sql/pgque.sql:2127 (`if batch_id is not null then return;`) and your own description of the intended fix in the #125 thread — "assert session 2 waits and returns the same `batch_id` after session 1 commits" — point to `v_count = 1`, not `0`. A `FOR UPDATE OF s` fix serializes the two `next_batch_custom` calls; the second one sees `sub_batch = batch_A` and returns it idempotently, then `get_batch_events(batch_A)` yields the same row.

If the assertion stays at `v_count = 0`, the harness will stay RED even after the intended fix lands and read as "fix is broken" to whoever runs it next.

how could resolve it:

```suggestion
  assert v_count = 1, format('session2 expected idempotent re-fetch (1 row), got %s', v_count);
  -- and assert s2_receive.batch_id = s1_receive.batch_id
```

The "no double-delivery" guarantee here is same batch_id, same rows (single-worker contract), not zero rows in the second session

tests/two_session_receive_lock.sh
```diff
@@ -0,0 +1,108 @@
...
+set -e
+
+if (( session1_status != 0 || session2_status != 0 )); then
+  echo "FAIL: two-session receive harness failed (session1=${session1_status}, session2=${session2_status})" >&2
+  print_debug
+  exit 1
+fi
+
+elapsed=$((end_epoch - start_epoch))
+if (( elapsed < 2 )); then
```
NikolayS: thinking maybe `< 3` is better; with `pg_sleep(4)` starting near T+0 and session 2 launching at T+1, the wait is ~3s. The `< 2` threshold lets a 2.0s response (which would mean the lock didn't actually serialize) pass.

tests/two_session_receive_lock.sh
```diff
@@ -0,0 +1,108 @@
...
+  exit 2
+fi
+
+psql_base=(psql --no-psqlrc -v ON_ERROR_STOP=1 "${PGQUE_TEST_DSN}")
+queue_name="two_session_receive_${$}_$(date +%s)"
+workdir="$(mktemp -d)"
+cleanup() {
+  rm -rf "${workdir}"
+}
+trap cleanup EXIT
```
NikolayS, line 36: for "absolute cleanup", worth adding to `cleanup`:

```sql
select pgque.unregister_consumer('${queue_name}', 'c1');
select pgque.drop_queue('${queue_name}', true);
```

tests/two_session_receive_lock.sh
```diff
@@ -0,0 +1,108 @@
+#!/usr/bin/env bash
```
NikolayS, line 1: please add the Apache-2.0 / PgQ derivation header that the rest of tests/ carries (see e.g. tests/test_concurrent_receive.sql).

tests/two_session_receive_lock.sh
```diff
@@ -0,0 +1,108 @@
...
+create temp table s1_receive as
+  select * from pgque.receive('${queue_name}', 'c1', 10);
+do \$\$
+declare
+  v_count integer;
+begin
+  select count(*) into v_count from s1_receive;
+  assert v_count = 1, format('session1 expected 1 message, got %s', v_count);
+end \$\$;
+select pg_sleep(4);
```
NikolayS: a very minor note:

Tightly coupled to the elapsed-time threshold on line 102. If you bump one, bump the other. Worth pulling both into a single HOLD_SECONDS=4 shell variable at the top so the contract is obvious.

NikolayS:
thanks @alceops!! left some comments above

NikolayS:
Follow-up status for #175:

- #175 was merged to add the standalone two-session receive harness before the 0.2 release.
- #220 fixed harness synchronization. The original version polled `sub_batch`, but that state is uncommitted while session 1 holds the transaction, so session 2 could start only after the lock was released. The fixed harness waits for session 1 to be inside `pg_sleep()` via `pg_stat_activity` before starting session 2.
- #125 has now been rebased on current `main` and wires this harness into CI as the RED/GREEN proof for the same-consumer receive-lock fix.

Thanks @alceops for the original harness contribution.

Issue #146, opened 2026-04-30
Umbrella: deep audit findings for Python, Go, and TypeScript drivers


NikolayS:
## Scope

Umbrella for the 2026-04-30 deep driver audit across:

- `clients/python`
- `clients/go`
- `clients/typescript`

Review goals:

- correctness against PgQue SQL semantics;
- alignment across all three drivers;
- “thin wrappers over SQL primitives / don't reinvent the wheel”;
- ease of use;
- tests that actually catch semantic regressions.

## Local test evidence

TypeScript:

```bash
cd clients/typescript
npm run check
# ok

# Fresh temp DB with sql/pgque.sql installed:
PGQUE_TEST_DSN=postgresql:///... npm test
# 22 passed
```

Python:

```bash
# Fresh temp DB with sql/pgque.sql installed, temp venv:
python -m pytest -q clients/python/tests
# 39 passed, 2 failed
```

Go:

```bash
go test ./clients/go
# not run locally: go binary missing on review host
```

Go findings are from code inspection + SQL repros; CI should be used as the execution gate.

## Filed issues

High/data-safety:

- #134 — high-level consumers can skip unreturned batch rows when `max_messages < batch size`.
- #135 — Go/TypeScript consumers ack batch even when `nack()` fails.
- #136 — unknown event type policy differs across Python/Go/TypeScript.
- #137 — Python tests fail on fresh install / weak transaction assertions.

Driver/API alignment and ergonomics:

- #138 — Go client missing thin wrappers/options for parity/ease-of-use.
- #140 — typed error mapping inconsistent/under-specified.
- #141 — explicit text-vs-JSON payload APIs missing/inconsistent.
- #142 — Go Consumer docs contradict runtime semantics.
- #143 — define behavior for SQL-valid NULL event type/payload.
- #144 — define/implement cross-driver API parity matrix.
- #145 — TypeScript client mutates process-global `pg` bigint parser at import time.

## Suggested order

1. Fix #134 first. It affects all three high-level Consumers and can silently skip work.
2. Fix #135 before relying on Go/TS consumers for retry/DLQ correctness.
3. Decide #136 once, then update docs/tests in all drivers.
4. Fix #137 so Python CI is trustworthy.
5. Use #144 as the parity roadmap; individual implementation issues can branch from it.

## Non-goals

This audit did not attempt to rewrite drivers or merge PRs. It documents defects and alignment gaps for red/green TDD fixes.

## Additional TypeScript/pass-3 findings filed

[...]

PR #179, opened 2026-05-03
fix(receive): re-queue unreturned batch rows on ack (#134)

### LINKED ISSUE #134: Clients: high-level consumers can skip unreturned batch rows when max_messages < batch size

NikolayS:
Closes #134.

## RCA

`pgque.receive(queue, consumer, max_return)` opens a PgQ batch covering the entire next-tick window via `pgque.next_batch`, then yields at most `max_return` rows. `pgque.ack()` unconditionally calls `pgque.finish_batch()`, which advances `sub_last_tick` past the whole window — so any batch row that `receive()` did not yield becomes unreachable for that consumer. Silent data loss.

Confirmed with the canonical repro from the issue:

```sql
select pgque.create_queue('q');
select pgque.register_consumer('q','c');
select pgque.send('q','default',jsonb_build_object('i',g))
  from generate_series(1, 105) g;
select pgque.force_tick('q'); select pgque.ticker('q');

create temp table got as select * from pgque.receive('q','c',100);
-- returned=100
select pgque.ack((select batch_id from got limit 1));
select count(*) from pgque.receive('q','c',100);  -- bug: returns 0 (lost 5 rows)
```

The umbrella audit (#146) ranked this #1 priority. The in-flight client PRs (#153/#154/#155) only mitigate the common-default case by raising `max_messages` to 500 — a single tick under burst can still exceed 500 (`ticker_max_count` is a *threshold*, not a hard cap), so the structural fix has to be SQL-side.

## Fix

The mismatch is between `max_return` (yield cap) and `finish_batch` (window-scoped). `pgque.receive()` now records the msg_ids it actually yielded; `pgque.ack()` re-queues whatever the batch contained but the caller never saw, then closes the batch as before.

- New tracking table `pgque.batch_returned (batch_id pk, returned_msg_ids bigint[])`.
- `pgque.receive()` accumulates yielded `ev_id`s and upserts the row before returning.
- `pgque.ack()` reads the row, inserts the unreturned events into `pgque.retry_queue` with `ev_retry_after = now()` and `ev_retry` **preserved** (these events were never delivered to a handler — they must not count as a retry attempt), deletes the tracking row, then calls `finish_batch`.
- `ON CONFLICT (ev_owner, ev_id) DO NOTHING` makes the requeue idempotent against any pre-existing retry_queue entry from a prior `nack()`.

### Backward compatibility

Callers that opened the batch via lower-level primitives (`next_batch` + `finish_batch` directly, or any path that bypasses `pgque.receive`) leave no row in `pgque.batch_returned`, so `ack()` falls through to plain `finish_batch` — legacy behavior preserved. No SECURITY DEFINER privileges were widened; the new table is touched only inside the existing SECURITY DEFINER functions.

### Why not the alternatives

- **Bump driver `max_messages` defaults** (#153/#154/#155): doesn't close the structural gap. `ticker_max_count` is a threshold, not a cap.
- **Don't advance the cursor on ack with unread rows**: would change the meaning of `ack(batch_id)` and require driver changes; also re-delivers everything the caller already processed.
- **Per-message ack API**: invasive client API change; doesn't help existing `ack(batch_id)` callers.

## Tests

**`tests/test_receive_partial_no_skip.sql`** — new red/green regression. Sends 105 events, receives 100, acks, runs `maint_retry_events` + tick, then asserts the remaining msg_ids 101..105 come back on the next `receive()`. Includes a final-drain check.

Red/green proof:

| install | test result |
| --- | --- |
| pre-fix (HEAD~) | `ERROR: after ack, the 5 unreturned events MUST be re-delivered, got 0 (issue #134)` |
| post-fix | `PASS: receive/ack does not skip unreturned batch rows (#134)` |

**`tests/test_api_receive.sql` Step 6** — previously codified the bug as expected behavior (`assert v_count = 0, 'ack(batch_id) should finish the whole batch...'`). Updated to assert `v_count = 2` after pumping the retry path, with a comment pointing at the new regression file.

**`tests/run_all.sql`** — wires the new test in. Full suite passes (78/78). `tests/acceptance/run_acceptance.sql` also green.

## Operational note

Re-queued events go through `pgque.retry_queue` and become deliverable again on the next `pgque.maint_retry_events()` cycle (scheduled by `pgque.start()` via pg_cron, or invoked manually). Time-to-redelivery is bounded by the maint cron period.

## Files

- `sql/pgque-api/receive.sql` — fix (source of truth)
- `sql/pgque.sql` — rebuilt via `bash build/transform.sh`
- `tests/test_receive_partial_no_skip.sql` — new
- `tests/test_api_receive.sql` — Step 6 updated
- `tests/run_all.sql` — register new test
- `docs/reference.md` — replace "batch-ownership caveat" with the new partial-receive safety guarantee

## Test plan

- [x] Red proof on HEAD~: assertion fires
- [x] Green proof on the fix: PASS
- [x] `tests/run_all.sql` — all green
- [x] `tests/acceptance/run_acceptance.sql` — all green
- [x] `tests/test_install_idempotency.sql` — re-install on top of an existing install is clean (the new `create table if not exists` survives)
- [ ] CI matrix (PG 14-18)

---
_Generated by [Claude Code](https://claude.ai/code/session_016GhBNhajFv3JkQsooHYLMz)_

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -171,6 +171,32 @@ test first, then the implementation that turns it green.
    the implementation. Applies to pgque-api, observability, client libraries,
    CLI. Does NOT apply to pgque-core repackaging (PgQ already has tests).
    See SPECx.md section 13.2.
+7. **Separate-transactions rule (snapshot visibility).** PgQue is
+   snapshot-based. The following operations MUST run in distinct, committed
+   transactions to be observable downstream — combining them in one tx
+   silently produces empty batches and dropped messages:
+   - `pgque.send` / `pgque.insert_event` (producer commit)
+     → `pgque.ticker` / `pgque.force_tick` (snapshot is taken here)
+     → `pgque.receive` / `pgque.next_batch` (sees only what committed
+        before the snapshot).
+   - `pgque.maint_retry_events` (re-inserts retry rows into event tables
+     with `pg_current_xact_id()`) → `pgque.ticker` (must run in a later
+     tx so the new ev_txids are visible in its snapshot) → `pgque.receive`.
+   - `pgque.maint_rotate_tables_step1` → `pgque.maint_rotate_tables_step2`
+     (PgQ design requirement; already documented in `sql/pgque-api/maint.sql`).
+
+   In SQL tests, this means one `do $$ ... $$` block per logical step
+   (see `tests/test_api_receive.sql` for the canonical pattern). In
+   client code, default-mode `pgxpool` / `pg.Pool` / `psycopg(autocommit)`
+   already runs each call in its own implicit tx — the rule is
+   transparently satisfied. The footgun is reaching for `Pool()` /
+   `client.conn` to wrap send + receive in one explicit tx; the consumer
+   side won't see what the producer just sent.
+
+   When adding a new test that crosses these boundaries, mark each
+   transaction boundary in a `-- separate transaction (snapshot
+   visibility)` comment. When adding new client examples, do not place
+   send/tick/receive inside a shared `BEGIN`/`COMMIT`.
 
 ## Copyright
 
--- a/clients/go/integration_test.go
+++ b/clients/go/integration_test.go
@@ -153,14 +153,33 @@ func TestNack_ToDLQAtRetryLimit(t *testing.T) {
 	const maxCycles = 5
 	for i := 0; i < maxCycles; i++ {
 		// Expire any pending retry delays so maint_retry_events picks them up
-		// immediately; in production these would expire naturally.
-		if _, err := client.Pool().Exec(ctx,
-			"update pgque.retry_queue set ev_retry_after = now() - interval '1 second'"); err != nil {
+		// immediately; in production these would expire naturally. Scope the
+		// update to this test's queue — pgque.ack() now re-queues unreturned
+		// batch rows globally (#134), so a global update would race with rows
+		// from sibling tests.
+		if _, err := client.Pool().Exec(ctx, `
+			update pgque.retry_queue rq
+			   set ev_retry_after = now() - interval '1 second'
+			  from pgque.queue q
+			 where q.queue_id = rq.ev_queue
+			   and q.queue_name = $1`, queue); err != nil {
 			t.Logf("retry_queue update unavailable: %v", err)
 		}
-		// Re-queue retry_queue rows for redelivery.
-		if _, err := client.Pool().Exec(ctx, "select pgque.maint_retry_events()"); err != nil {
-			t.Logf("maint_retry_events unavailable, using ticker fallback: %v", err)
+		// Re-queue retry_queue rows for redelivery. maint_retry_events()
+		// processes at most 10 rows per call; loop until drained so that
+		// past-due rows left behind by sibling tests (which can pile up under
+		// the global retry_queue now that pgque.ack() re-queues unreturned
+		// batch events for #134) do not starve this test's row.
+		for {
+			var n int
+			if err := client.Pool().QueryRow(ctx,
+				"select pgque.maint_retry_events()").Scan(&n); err != nil {
+				t.Logf("maint_retry_events unavailable, using ticker fallback: %v", err)
+				break
+			}
+			if n == 0 {
+				break
+			}
 		}
 		tick(t, client, queue)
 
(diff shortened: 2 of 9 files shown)


NikolayS:
Rebased onto current main and force-pushed `claude/review-open-issues-KFn9A` at `746b60d`.

Validation run:
- `./build/transform.sh` ✅
- Installed `sql/pgque.sql` on Postgres 18 and ran `tests/test_api_receive.sql` + `tests/test_receive_partial_no_skip.sql` ✅
- Fresh Postgres 18 install + `tests/run_all.sql` ✅
- `git diff --check` ✅

Note: attempted Go client tests for the touched Go test, but `go` is not installed in this environment.

NikolayS:
## samorev Code Review Report

- **PR:** NikolayS/PgQue#179 - fix(receive): re-queue unreturned batch rows on ack (#134)
- **Author:** @NikolayS
- **AI-Assisted:** Unknown

| Pipeline | Coverage |
|----------|----------|
| PASS | Not reported |

---

### BLOCKING ISSUES (1)

---

### Summary

| CI/Pipeline | 0 | 0 | 0 |
| Security | 0 | 0 | 0 |
| Bugs | 0 | 0 | 0 |
| Tests | 0 | 0 | 0 |
| Guidelines | 0 | 0 | 0 |
| Docs | 0 | 0 | 0 |
| Metadata | 1 | 0 | 0 |

Note:

<details>

```text
provider=github
kind=pr
project=NikolayS/PgQue
number=179
target=github:NikolayS/PgQue#179
state=OPEN
draft=true
diff_lines=626
diff_added=409
diff_removed=10
diff_bytes=25945
comments_count=1
commits_count=4
ci_status=success
ci_summary=total=11 success=11 failure=0 pending=0 other=0
blocking=false
posted_by=gh
no_comment=false
live_posting=posted
```

</details>

---

PR #186, opened 2026-05-03
fix(clients/typescript): nack reasons, retryAfter seconds, coverage

NikolayS:
## Summary

Follow-up to merged #154 addressing review feedback. Repo has no users yet, so the public API change to `nack`'s `retryAfter` (now `number` seconds) is unconstrained.

## Changes

- **Thread real error message into nack reason** (consumer.ts):
  - Unknown type → `` `unknown event type: ${msg.type}` ``
  - Handler error → `` `handler error: ${formatErr(err)}` ``
- **Document units explicitly**:
  - `ConsumerOptions.pollInterval` JSDoc says **milliseconds**.
  - `Client.nack` `retryAfter` is now **`number` seconds** (was `string` interval). Picked seconds to match the Python client (`retry_after: int = 60`) and align the cross-driver convention.
  - `nack` validates non-negative finite numbers; builds `'<n> seconds'` interval text internally.
- **Export `DEFAULT_MAX_MESSAGES`** constant (re-exported from `index.ts`); tests now import it instead of the magic literal `2_147_483_647`.
- **Three new env-gated live-PG tests**:
  - `dl_reason` lands the threaded reason string in `pgque.dead_letter`.
  - Default consumer drains a 105-message batch in one poll.
  - Partial-batch `ok`/`boom`/`ok` asserts only `boom` is redelivered.

## Test plan

- [x] `tsc --noEmit` (src + test tsconfigs) clean
- [x] `vitest --run` without `PGQUE_TEST_DSN`: 5 passed, 35 skipped (env-gated)
- [x] `vitest --run` with `PGQUE_TEST_DSN`: 40 passed, 0 failed
- [ ] CI green on PG 14–18

## Notes

- `nack(... { retryAfter: '5 seconds' })` callers must update to `nack(... { retryAfter: 5 })`. Acceptable per the no-users policy.
- The orphaned source branch `fix/clients-typescript-v0.2.0` was resurrected by an automated step. The three commits on this branch are the cherry-picked, freshly-signed versions. Safe to delete the resurrected branch.

https://claude.ai/code/session_0153x6qY6bLADom5d5wwMYRm

---
_Generated by [Claude Code](https://claude.ai/code/session_0153x6qY6bLADom5d5wwMYRm)_

### DIFF
--- a/clients/typescript/src/client.ts
+++ b/clients/typescript/src/client.ts
@@ -237,12 +237,27 @@ export class Client {
    * Negatively acknowledge a single message. Routes to the retry queue if
    * `retry_count < queue_max_retries`, otherwise to the dead-letter queue.
    * Other messages in the same batch are not affected.
+   *
+   * `opts.retryAfter` is the retry delay in **seconds** (default `60`).
+   * The driver binds it as a PostgreSQL `interval` (`'<n> seconds'`) for
+   * `pgque.nack`'s `i_retry_after` parameter; cross-driver parity with
+   * `pgque-py` and `pgque-go`.
    */
   async nack(batchId: bigint, msg: Message, opts: NackOptions = {}): Promise<void> {
     if (typeof batchId !== 'bigint') {
       throw new PgqueSqlError('nack', { cause: new Error('batchId must be bigint') });
     }
-    const retryAfter = opts.retryAfter ?? '60 seconds';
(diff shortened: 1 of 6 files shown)


NikolayS:
Rebased onto current `main` (1581f39) and force-pushed with lease.\n\nValidation run:\n- `bun run check`\n- `bun run test` (18 passed, 41 skipped)\n- `git diff --check`

Issue #183, opened 2026-05-03
Distribute TypeScript client (pgque-ts) to npm


NikolayS:
## Goal

Publish `pgque-ts` (the TypeScript client at `clients/typescript/`) so users can `npm install pgque` (or `bun add ...`) instead of cloning the repo.

**0.2.0 priority.**

## Current state

- `clients/typescript/` exists as a Bun-managed package (node-postgres based; see `clients/typescript/package.json`).
- CI runs the test suite against PG 18 (`typescript-client-tests` job in `.github/workflows/ci.yml`).
- README shows `new PgqueClient('postgresql://...')` usage; no install command.

## Scope

1. Confirm the **npm package name** (`pgque` if available, otherwise `pgque-ts`, `@pgque/client`, etc.). If we want to claim a scope we own the org for, decide that now.
2. Audit `clients/typescript/package.json` for npm publish requirements: `name`, `version`, `description`, `keywords`, `repository`, `homepage`, `bugs`, `license`, `author`, `engines`, `files` (or `.npmignore`) so we don't ship `node_modules` / tests.
3. Decide on the **build artifact**: dual ESM + CJS? Type declarations (`.d.ts`)? Source maps? `exports` field in `package.json`?
4. Pick the **publish toolchain**: `npm publish` directly, `tsup` / `pkgroll` for the build, `changesets` for changelog and version bumps.
5. Decide the **release process**: GitHub Action with npm Trusted Publisher (OIDC) so we don't store long-lived `NPM_TOKEN` in CI secrets.
6. Versioning policy: client version independent of server `pgque.version()`.
7. Confirm the runtime support matrix: Node version range, Deno/Bun support, browser non-support.
8. Update `clients/typescript/README.md` with the published install command once stamped.

## Out of scope

- Major API changes to the TS client.
- Migrating off `bun` for development; this is about distribution to npm, which is the Node consumer surface.

## Sister issues

- Python client distribution
- Go client distribution
- Distribution & discoverability of the SQL install (separate issue, lower priority)

NikolayS:
Release-readiness check for TypeScript client:

- npm name check: `pgque` currently returns 404 from the npm registry API, so the name appears available.
- `clients/typescript/package.json` already has npm publish metadata, ESM exports, declarations, `files`, `engines`, provenance config, repository/bugs/homepage, and scripts.
- Local package checks passed:

```text
bun install --frozen-lockfile
bun run check
bun run build
npm pack --dry-run
```

[...]

PR #189, opened 2026-05-03
fix(clients/typescript): scope bigint parser to pgque pool (#145)

### LINKED ISSUE #145: TypeScript client: avoid process-global pg bigint parser mutation at import time
## Problem

The TypeScript client mutates node-postgres' process-global bigint parser at module import time:

```ts
types.setTypeParser(20, (val) => BigInt(val));
```

This changes how **all** `pg` clients in the same Node.js process parse PostgreSQL `bigint`, not only PgQue's pool.

## Why this matters

This is surprising in library code. An application may already rely on node-postgres' default `int8` behavior (`string`) or its own parser. Importing `pgque` changes global behavior outside PgQue and can break unrelated queries.

The comment says it is documented, but this is still a risky default for a client library.

## Suggested fix options

- Avoid global parser mutation; parse PgQue result fields explicitly (`BigInt(row.msg_id)`, etc.).
- Or make it opt-in via connect option, e.g. `connect(dsn, { parseBigInt: true })`.
- At minimum, put a prominent warning in README/API docs and tests proving the side effect is intentional.

## Acceptance criteria

- Importing `pgque` does not silently mutate global `pg` type parsers by default, or the mutation is explicit/opt-in and heavily documented.
- Tests cover bigint parsing for PgQue rows without relying on ambient process-global side effects.

NikolayS:
## Root cause analysis

`clients/typescript/src/client.ts` called `types.setTypeParser(20, (val) => BigInt(val))` at module-import time (line 23). This mutates the **process-global** `pg-types` parser table — every other `pg.Pool` / `pg.Client` in the same Node process suddenly receives `bigint` instead of the default `string` for `int8` columns, silently breaking unrelated application code.

## The fix

Remove the module-level `types.setTypeParser(20, ...)` call entirely. In its place, define a `pgqueTypes` `CustomTypesConfig` object that overrides OID 20 (int8) to `BigInt` and delegates all other OIDs to the global defaults. Pass `types: pgqueTypes` to the `pg.Pool` constructor inside `connect()`.

This scopes the bigint parser to pgque's own pool only. All other `pg` clients in the process are unaffected.

Key change in `clients/typescript/src/client.ts`:

```ts
// Before (global mutation):
types.setTypeParser(20, (val) => BigInt(val));

// After (per-pool, no global mutation):
const pgqueTypes: pg.CustomTypesConfig = {
  getTypeParser(oid: number, format?: string) {
    if (oid === 20) return (val: string) => BigInt(val);
    return types.getTypeParser(oid, format as 'text' | 'binary');
  },
};
// ...
const pool = new Pool({ connectionString: dsn, types: pgqueTypes, ...poolOptions });
```

## Red → green proof

**Red** (on `HEAD~` / unfixed code): `bun run test` output:

```
× after importing pgque, a second pg.Pool still gets string for int8
  → expected 'bigint' to be 'string'
```

**Green** (on this PR):

```
✓ global int8 parser is string-returning before pgque import
✓ after importing pgque, a second pg.Pool still gets string for int8 (global parser is unchanged)
✓ pgque pool uses per-pool bigint parser for OID 20 (internal CustomTypesConfig)
Test Files  3 passed | 1 skipped (4)
Tests  8 passed | 32 skipped (40)
```

## Test plan

- [x] `test/bigint-global-leak.test.ts` — pure-vitest, no live PG required
  - [x] global OID-20 parser is `string`-returning before import
  - [x] global OID-20 parser remains `string`-returning after pgque import
  - [x] pgque's per-pool config returns `bigint` for OID-20 internally
- [x] Existing tests pass (`rejects bad DSN with PgqueConnectionError`, all consumer mock tests)
- [x] `bun run check` (type-check) passes
- [x] `bun run build` succeeds
- [x] README Caveats section updated: removed global-mutation warning, replaced with note that the parser is scoped to pgque's pool

Closes #145

https://claude.ai/code/session_015Tf54iAd24uQPKCBaXeiTV

---
_Generated by [Claude Code](https://claude.ai/code/session_015Tf54iAd24uQPKCBaXeiTV)_

### DIFF
--- a/clients/typescript/README.md
+++ b/clients/typescript/README.md
@@ -125,28 +125,13 @@ All errors derive from `PgqueError`:
 
 ## Caveats
 
-### Global BIGINT parser mutation
+### bigint columns
 
-Importing `pgque` calls `types.setTypeParser(20, ...)` at module load
-time. This mutates the process-global `pg-types` parser table so that
-**all** `pg.Pool` / `pg.Client` instances in the same Node.js process
-will return PostgreSQL `bigint` columns as JS `bigint` instead of the
-default string representation.
-
-Practical impact:
-
-- If other code in your process uses `pg` and relies on `bigint` coming
-  back as a string (the `pg` default), those columns will silently change
-  type after `pgque` is imported.
-- The change is intentional: JS `bigint` is the correct representation for
-  PostgreSQL `bigint` and avoids silent precision loss above
-  `Number.MAX_SAFE_INTEGER`. The Go and Python pgque drivers behave the
-  same way.
-- If you cannot accept this side effect, do not import this package.
-
-There is no opt-out once the module is loaded — Node.js module caches
-mean the parser is set exactly once, regardless of how many times the
-package is imported.
+`Message.msgId`, `Message.batchId`, and the return values of `send()` /
+`sendBatch()` are JS `bigint`. The `int8` → `bigint` parser is registered
+only on pgque's internal pool via a per-pool `CustomTypesConfig` — it
+does **not** touch the process-global `pg-types` table. Other `pg.Pool`
+or `pg.Client` instances in the same process are unaffected.
 
 ## Tests
 
--- a/clients/typescript/src/client.ts
+++ b/clients/typescript/src/client.ts
@@ -13,14 +13,24 @@ import type { ConsumerOptions, Event, Message, NackOptions } from './types.js';
 
 const { Pool, types } = pg;
 
-// PostgreSQL `bigint` (oid 20) is parsed by `pg` as string by default to
+// PostgreSQL `bigint` (OID 20) is parsed by `pg` as string by default to
 // avoid silent precision loss above Number.MAX_SAFE_INTEGER. We promote to
 // JS `bigint` for safety AND ergonomics — `bigint` is the natural type for
 // PG `bigint`, matches the Go driver's `int64`, and round-trips losslessly.
 //
-// This mutates a process-global parser table. We do it once at module load
-// and document the side effect in the README.
-types.setTypeParser(20, (val) => BigInt(val));
+// We do NOT touch the process-global parser table. Instead, each Pool created
+// by pgque is given a per-pool CustomTypesConfig that overrides OID 20 only
+// for queries issued by pgque. Unrelated pg pools in the same process are
+// unaffected.
+const pgqueTypes: pg.CustomTypesConfig = {
+  getTypeParser(oid: number, format?: string) {
+    if (oid === 20) {
+      // int8 / bigint → JS bigint
+      return (val: string) => BigInt(val);
+    }
+    return types.getTypeParser(oid, format as 'text' | 'binary');
+  },
+};
 
 const PG_RAISE_EXCEPTION_CODE = 'P0001';
 
@@ -313,7 +323,7 @@ export async function connect(
   dsn: string,
   poolOptions: Omit<pg.PoolConfig, 'connectionString'> = {},
 ): Promise<Client> {
-  const pool = new Pool({ connectionString: dsn, ...poolOptions });
+  const pool = new Pool({ connectionString: dsn, types: pgqueTypes, ...poolOptions });
   let probe: pg.PoolClient;
   try {
     probe = await pool.connect();
--- a/clients/typescript/src/client.ts
+++ b/clients/typescript/src/client.ts
@@ -333,7 +333,14 @@ export async function connect(
   dsn: string,
   poolOptions: Omit<pg.PoolConfig, 'connectionString' | 'types'> = {},
 ): Promise<Client> {
-  const pool = new Pool({ connectionString: dsn, ...poolOptions, types: pgqueTypes });
+  // Defensively strip `types` from poolOptions before spreading. The
+  // `Omit<..., 'types'>` already rejects this at compile time, but JS
+  // callers (or `as` casts) can still smuggle one in. Dropping it here
+  // makes the pgque types config impossible to override regardless of
+  // spread order — see REV review on PR #189.
+  const { types: _userTypes, ...restPoolOptions } = poolOptions as pg.PoolConfig;
+  void _userTypes;
+  const pool = new Pool({ connectionString: dsn, ...restPoolOptions, types: pgqueTypes });
   let probe: pg.PoolClient;
   try {
     probe = await pool.connect();
--- a/clients/typescript/src/client.ts
+++ b/clients/typescript/src/client.ts
@@ -22,7 +22,14 @@ const { Pool, types } = pg;
 // by pgque is given a per-pool CustomTypesConfig that overrides OID 20 only
 // for queries issued by pgque. Unrelated pg pools in the same process are
 // unaffected.
-const pgqueTypes: pg.CustomTypesConfig = {
+/**
+ * Per-pool type overrides used by pgque's `connect()`.
+ * Parses PostgreSQL `bigint` (OID 20) as JS `bigint` without touching the
+ * process-global `pg.types` table.
+ *
+ * @internal — exported for testing only; do not depend on this in application code.
+ */
+export const pgqueTypes: pg.CustomTypesConfig = {
   getTypeParser(oid: number, format?: string) {
     if (oid === 20) {
       // int8 / bigint → JS bigint
@@ -309,6 +316,9 @@ export class Client {
  * the connection eagerly; rejects with {@link PgqueConnectionError} on
  * failure.
  *
+ * The `types` parser config is reserved for pgque's internal bigint parsing;
+ * user-supplied `types` is ignored.
+ *
  * @example
  * ```ts
  * const client = await connect('postgres://user:pass@localhost/mydb');
@@ -321,9 +331,9 @@ export class Client {
  */
 export async function connect(
   dsn: string,
-  poolOptions: Omit<pg.PoolConfig, 'connectionString'> = {},
+  poolOptions: Omit<pg.PoolConfig, 'connectionString' | 'types'> = {},
 ): Promise<Client> {
-  const pool = new Pool({ connectionString: dsn, types: pgqueTypes, ...poolOptions });
+  const pool = new Pool({ connectionString: dsn, ...poolOptions, types: pgqueTypes });
   let probe: pg.PoolClient;
   try {
     probe = await pool.connect();
--- a/clients/typescript/src/index.ts
+++ b/clients/typescript/src/index.ts
@@ -29,10 +29,9 @@
  * }
  * ```
  *
- * **Side effect on import:** registers a global `pg-types` parser for
- * `bigint` (oid 20) that promotes the column to JS `bigint`. This avoids
- * silent precision loss but also affects any other `pg`-using code in
- * the same process. Documented here for transparency.
+ * **bigint columns:** `msg_id`, `batch_id`, and `send()` / `sendBatch()`
+ * return values are JS `bigint`. The parser is scoped to pgque's own pool
+ * and does not affect other `pg` clients in the same process.
  */
 
 export { Client, connect } from './client.js';
--- a/clients/typescript/src/index.ts
+++ b/clients/typescript/src/index.ts
@@ -34,7 +34,7 @@
  * and does not affect other `pg` clients in the same process.
  */
 
-export { Client, connect } from './client.js';
+export { Client, connect, pgqueTypes } from './client.js';
 export { Consumer } from './consumer.js';
 export {
   PgqueConnectionError,
--- a/clients/typescript/test/bigint-global-leak.test.ts
+++ b/clients/typescript/test/bigint-global-leak.test.ts
@@ -10,6 +10,7 @@
 
 import { beforeAll, describe, expect, it, vi } from 'vitest';
 import pg from 'pg';
+import { pgqueTypes } from '../src/index.js';
 
 // ---------------------------------------------------------------------------
 // Test 1 (red on unfixed code, green on fix):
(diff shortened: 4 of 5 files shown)


NikolayS:
## REV review — PR #189

### Verdict
One blocking issue (user-supplied `types` in `poolOptions` silently overrides the bigint parser); several non-blocking findings worth addressing before merge.

### Summary
The PR correctly eliminates the global `types.setTypeParser(20, ...)` side-effect by introducing a per-pool `CustomTypesConfig` (`pgqueTypes`) passed to `pg.Pool`. The approach is idiomatic and the fix is sound for the common case. The main risk is that the `poolOptions` spread order makes it trivially easy for a caller to clobber `pgqueTypes` and silently break bigint parsing across all pgque API calls. The test suite verifies the global-leak invariant well but does not cover the override scenario.

### CI status
All 9 checks green: `TypeScript client tests`, `Go client tests`, `Python client tests`, `verify`, and `test (14/15/16/17/18)`. No red.

---

### Findings

#### Security
None. The change removes a global mutation and adds no new input-parsing surface. No credentials or secrets touched.

#### Bugs / correctness

- **[BLOCKING] severity HIGH, confidence 9/10 — `clients/typescript/src/client.ts`:323 — user-supplied `types` in `poolOptions` silently overwrites `pgqueTypes`, breaking bigint parsing**
  - Evidence: `const pool = new Pool({ connectionString: dsn, types: pgqueTypes, ...poolOptions });` — because `poolOptions` is spread last, any caller who passes `{ types: myCustomConfig }` (which `Omit<pg.PoolConfig, 'connectionString'>` still permits) completely replaces `pgqueTypes`. After that, OID 20 comes back as a string, and every call to `send()` / `receive()` returns or expects `bigint` from the DB but gets `string`, causing silent type errors or runtime `BigInt(undefined)` crashes.
  - Fix: Either (a) exclude `types` from the public `poolOptions` parameter with a stricter `Omit<pg.PoolConfig, 'connectionString' | 'types'>`, or (b) spread `poolOptions` first and then pin `types: pgqueTypes` last so pgque's config always wins. Option (a) is cleaner and surfaces the constraint to callers at compile time.

- **[NON-BLOCKING] severity LOW, confidence 6/10 — `clients/typescript/src/client.ts`:26 — `format` passed as `undefined` to `types.getTypeParser` for non-int8 OIDs**
  - Evidence: `return types.getTypeParser(oid, format as 'text' | 'binary');` — `format` is `undefined` when `pg` omits it. `pg`'s `getTypeParser` defaults `undefined` to `'text'`, so this is behaviorally safe today, but the cast hides the discrepancy between the declared `format?: string` and the expected `'text' | 'binary'`.
  - Fix: `return types.getTypeParser(oid, (format ?? 'text') as 'text' | 'binary');` to be explicit.

#### Tests

- **[NON-BLOCKING] severity MEDIUM, confidence 8/10 — `clients/typescript/test/bigint-global-leak.test.ts`:52-87 — third test does not exercise the production `pgqueTypes` object**
  - Evidence: The test reconstructs an equivalent `localPgqueTypes` inline rather than importing `pgqueTypes` from `client.ts`. If `client.ts` changed its OID guard (e.g. `oid === 21` by mistake), this test would still pass because it tests its own local copy, not the real one.
  - Fix: Export `pgqueTypes` (or a `getTypeParser` test-helper) from `client.ts` (or a dedicated internal module) so the test can import and exercise it directly.

- **[POTENTIAL] severity LOW, confidence 5/10 — missing concurrent-pool isolation test**
  - There is no test that creates both a pgque pool AND an unrelated `pg.Pool` concurrently in the same process and asserts they parse OID 20 differently. The current test only checks the global parser, not live pool behaviour.
  - Fix: Add a test (live-PG or via mocked `pool.query`) that verifies a foreign pool returns `string` while pgque's pool returns `bigint` for the same int8 value, in the same test scope.

#### Guidelines (CLAUDE.md)

- **[NON-BLOCKING] severity INFO, confidence 9/10 — commit/PR title exceeds 50-char subject limit**
  - Evidence: PR title `fix(clients/typescript): scope bigint parser to pgque pool (#145)` is 63 characters. CLAUDE.md: "Subject < 50 chars".
  - Fix: e.g. `fix(ts-client): scope bigint parser to pool` (43 chars). Issue reference belongs in the body, not the subject.

- **[NON-BLOCKING] severity INFO, confidence 8/10 — `Omit<pg.PoolConfig, 'connectionString'>` allows `types` through, inconsistent with the intent of the fix**
  - This is the same root cause as the blocking bug above, but from a type-hygiene angle: the public API signature implicitly promises callers they can pass `types`, even though doing so breaks pgque internals.
  - Rule: CLAUDE.md "Modern API functions must reduce cleanly to PgQ primitives" / "Keep changes surgical."

#### Docs

- **[NON-BLOCKING] severity MEDIUM, confidence 8/10 — `connect()` TSDoc does not mention the `types` interaction**
  - Evidence: `clients/typescript/src/client.ts`:307-321 — the JSDoc for `connect()` says nothing about `poolOptions.types` being reserved/ignored/overridden. A caller reading the signature will assume all `PoolConfig` fields work normally.
  - Fix: Add a `@param poolOptions` note: "`types` is reserved — pgque installs its own `CustomTypesConfig` for bigint handling and will override (or be overridden by) any `types` value here."

- **[POTENTIAL] severity LOW, confidence 5/10 — README Caveats: removed rationale for using `bigint` (precision-loss explanation)**
  - The old section explained *why* `bigint` is used (silent precision loss above `Number.MAX_SAFE_INTEGER`). The new section omits this. Developers unfamiliar with the issue may wonder why the driver doesn't just return numbers.
  - Fix (optional): Append one sentence, e.g. "JS `number` silently loses precision for integers above `Number.MAX_SAFE_INTEGER`; `bigint` round-trips losslessly."

---

### Recommended next step
**Resolve the blocking issue before merge:** change the `connect()` signature to `Omit<pg.PoolConfig, 'connectionString' | 'types'>` (preferred) or move `types: pgqueTypes` after the `poolOptions` spread. Update `connect()` TSDoc accordingly. The test gap (third test using a local copy) is worth fixing in the same pass given it's a small change.

---

---

NikolayS:
Addressed all REV findings in commit `cda7404` (green) on this branch, preceded by red-proof commit `39cd63b`:

**Blocking (spread order):** Moved `types: pgqueTypes` to after `...poolOptions` in the `Pool` constructor call — `{ connectionString: dsn, ...poolOptions, types: pgqueTypes }` — so pgque's bigint parser always wins regardless of what the caller passes.

**Type safety:** Narrowed `poolOptions` from `Omit<PoolConfig, 'connectionString'>` to `Omit<PoolConfig, 'connectionString' | 'types'>`, so `connect(dsn, { types })` is now a compile-time error (belt-and-suspenders with the runtime fix).

**pgqueTypes export:** Exported `pgqueTypes` with `@internal` TSDoc from `client.ts` / `index.ts`. Fixed `bigint-global-leak.test.ts` test 3 to import and use the production object instead of a locally-reconstructed copy — future changes to `pgqueTypes` are automatically reflected in that test.

**TSDoc:** Added a one-line note to `connect()` documenting that `types` is reserved: "The `types` parser config is reserved for pgque's internal bigint parsing; user-supplied `types` is ignored."

**New regression test:** `test/connect-types-precedence.test.ts` — uses `vi.mock('pg')` to capture the Pool constructor argument without a live DB; asserts `types === pgqueTypes` by identity, asserts the `@ts-expect-error` directive is consumed (TS rejects `{ types }` at compile time), and proves the JS-caller bypass case (cast through `as`) is also safe at runtime.

---

NikolayS:
## REV review

The fix correctly removes the module-level `types.setTypeParser(20, ...)` call and replaces it with a per-Pool `CustomTypesConfig` that overrides only OID 20 and delegates everything else to `pg.types.getTypeParser`. Verified: no other `setTypeParser` / global `pg-types` mutation remains under `clients/typescript/src/`. The delegation cast `format as 'text' | 'binary'` is sound — `pg-types`' `getTypeParser` signature accepts `'text' | 'binary'` (default `'text'`), and an `undefined` here passes through to pg's default. Tests provide red→green proof and an identity assertion against the production `pgqueTypes` object. One blocking concern remains: the option-merge order lets a caller silently override `pgqueTypes` at runtime, which re-opens an adjacent footgun.

FINDING (bug-hunter):
- severity: HIGH
- confidence: 9
- file: clients/typescript/src/client.ts
- line: ~334 (`new Pool({ connectionString: dsn, ...poolOptions, types: pgqueTypes })`)
- issue: The TS signature `Omit<…, 'types'>` blocks `types` at compile time, but a JS caller (or `as any`) can still pass `{ types: ... }` in `poolOptions`. Because `...poolOptions` is spread **before** `types: pgqueTypes`, pgque's parser does win here — but this depends entirely on key order in the object literal. If anyone reorders these fields later (e.g., to put `types` first for "consistency"), user input will silently overwrite the bigint parser and `msgId` / `batchId` will come back as strings, breaking every consumer downstream. There is one new test (`connect-types-precedence.test.ts`) that asserts the current order, which is good — but the safer construction is to spread user options first into a local, then explicitly assign `types`, OR to strip `types` defensively before the spread. Recommend: `const { types: _ignored, ...rest } = poolOptions as pg.PoolConfig; new Pool({ connectionString: dsn, ...rest, types: pgqueTypes });` — this makes the intent explicit and eliminates the order-dependence.
- evidence: `new Pool({ connectionString: dsn, ...poolOptions, types: pgqueTypes });`
- fix: destructure-and-strip `types` from `poolOptions` before constructing the Pool, so the override is order-independent.

### Non-blocking

FINDING (test-analyzer):
- severity: MEDIUM
- confidence: 7
- file: clients/typescript/test/bigint-global-leak.test.ts
- line: 12, 36
- issue: The "no global mutation" test imports `pgqueTypes` from `../src/index.js` at the top of the file, which already loads the pgque module before the first `it()` runs. The second test then does `await import('../src/index.js')` — but Node's module cache means the side-effect (if any) already happened during the top-level import, not inside the `it`. The first test ("string-returning before pgque import") is therefore mislabeled: pgque is already imported by then. The assertion still proves the global parser is unchanged after pgque is loaded, so the test is not a *false* safety signal — but the "before / after" framing is wrong. Either move the `pgqueTypes` import into a separate file that the assertion test does not load, or reword the test names to drop "before import".
- untested_scenarios: a true "before" snapshot would require running pgque in a separate worker / `vm` context.

FINDING (bug-hunter):
- severity: LOW
- confidence: 6
- file: clients/typescript/src/client.ts
- line: ~33 (`return types.getTypeParser(oid, format as 'text' | 'binary')`)
- issue: When `format` is `undefined`, this delegates with `undefined` cast to a string-union — pg-types treats undefined as `'text'`, so behavior is correct, but the cast is misleading. Either default the param (`format: 'text' | 'binary' = 'text'`) or drop the cast (pg-types accepts `format?`).
- fix: cosmetic; tighten the signature.

FINDING (docs-reviewer):
- severity: LOW
- confidence: 6
- file: clients/typescript/README.md
- line: 127–134 (new "bigint columns" section)
- issue: The new section is correctly scoped and avoids PR/version refs, satisfying CLAUDE.md. It would help to add one sentence: "If your application code was relying on the previous global behavior — e.g. another `pg.Pool` returning bigint after pgque was imported — that side effect is gone; install your own `types.setTypeParser(20, …)` if you need it." Backward-compat note for users coming from the broken behavior.
- suggestion: add a one-liner pointing users who depended on the side-effect to the explicit opt-in.

### Skipped (SOC2)

Skipped per repo policy.

Verdict: REQUEST CHANGES

---

NikolayS:
REV blocking finding addressed in 6754bd3: `types` is now defensively destructured out of `poolOptions` before the spread, so a JS caller (or an `as`-cast) can't smuggle in a parser config that overrides pgque's bigint parser. Fix is no longer spread-order-dependent.

---

NikolayS:
## REV review (round 2)

Re-reviewed at `6754bd3`. The prior round-1 blocker — `connect()` relying purely on spread order to win against a user-supplied `types` (the `Omit<..., 'types'>` only protects TS callers; a JS or `as`-cast caller could still smuggle one in) — is resolved. `client.ts` lines 341–343 now defensively destructure `types` out of `poolOptions` before spreading, then re-add `types: pgqueTypes` last. The `connect-types-precedence.test.ts` runtime test (identity check `capturedPoolConfig.types === pgqueTypes` even when a `types` is forced through via `as`-cast) covers this and passes in CI. No other code path constructs a `Pool` from `poolOptions` (only one `new Pool(...)` site exists in `src/`). The `bigint-global-leak.test.ts` rename is correct: pgque is statically imported at the top of the file, so the assertion is genuinely "after import", not "before"; the new wording matches reality. All 10 CI checks green.

### Blocking
None.

### Non-blocking
- **test-analyzer / guidelines (confidence 6, LOW):** `clients/typescript/test/bigint-global-leak.test.ts:11` imports `beforeAll` and `vi` from vitest but neither is used in the file (only `describe`, `it`, `expect` are referenced). Dead imports; trivial cleanup. Not worth a re-spin on its own — fold into the next typescript-client touch.
- **guidelines-checker (confidence 5, INFO):** `clients/typescript/src/client.ts:340` comment says `see REV review on PR #189`. CLAUDE.md forbids PR numbers in README/`docs/`; source-code comments aren't explicitly listed, so this is borderline. Consider rewording to "see the round-1 REV review" or dropping the reference, since the destructure is self-explanatory with the surrounding rationale.
- **docs-reviewer (confidence 4, INFO):** the rewritten README "bigint columns" caveat is clean and new-user-friendly. No action.

### Skipped (SOC2)
Per project policy, SOC2 findings are not applied to this repo.

Bug-hunter, security-reviewer: NO_FINDINGS. The fix removes the global `pg-types` mutation cleanly; the per-pool `CustomTypesConfig` delegates non-OID-20 calls back to `types.getTypeParser`, preserving default behaviour for everything else. No prototype-pollution surface, no unsafe casts that survive the destructure, no async/cleanup regressions in `connect()` (the existing `pool.end().catch(...)` on probe failure is unchanged).

Verdict: APPROVE

---

NikolayS:
## Step 3 — real testing

Walked the user-facing claim: importing pgque must not break BIGINT parsing for unrelated `pg.Pool` instances in the same Node process.

Setup:
- PostgreSQL 16 fresh install of `sql/pgque.sql`
- `bun install --frozen-lockfile` in `clients/typescript`

Test script:
```ts
// /tmp/pr189-walkthrough.ts
import pg from 'pg';
import { connect } from '/home/user/pgque/.claude/worktrees/agent-a24b941f9b969de8a/clients/typescript/src/index.js';

// 1. Show the global int8 parser is the default `string` BEFORE we import pgque (already happened — `connect` is imported above, so the import has run. Test the result of the global parser RIGHT NOW.)
const globalParser = pg.types.getTypeParser(20);
console.log('global int8 parser type:', typeof globalParser('42'));  // expect "string"

// 2. A SECOND pg.Pool — like a separate library doing its own queries — should still return string for int8.
const otherPool = new pg.Pool({ connectionString: dsn });
const otherResult = await otherPool.query<{ v: unknown }>('select 1::bigint as v');
console.log('other pool int8 row.v:', typeof otherResult.rows[0].v, otherResult.rows[0].v);  // expect "string", "1"
await otherPool.end();

// 3. pgque's own pool MUST get bigint.
const client = await connect(dsn);
const m = await client.send('walkthrough_q', { type: 'pr189', payload: { ok: true } });
console.log('pgque send returned:', typeof m, m);  // expect "bigint", some non-zero value
await client.close();
```

Output:
```
global int8 parser type: string
other pool int8 row.v: string 1
pgque send returned: bigint 1n
```

Verdict: ✓ Global `pg-types` int8 parser remains `string`-returning. Independent `pg.Pool` instances are unaffected. pgque's own pool returns `bigint`. The release-blocker hazard for npm publish (#183) is gone.

---

PR #193, opened 2026-05-03
fix(clients/python): tighten test transactions + assertions (#137)

### LINKED ISSUE #137: Python client: tests fail on fresh PgQue install and encode weak transaction assertions

NikolayS:
## Summary

- **Switch all `pgque.ticker()` → `pgque.ticker(queue)`** across six test files (`test_consumer.py`, `test_consumer_listen_stop.py`, `test_nack.py`, `test_receive.py`, `test_send.py`, `test_smoke.py`) to avoid cross-queue side effects and make per-queue intent explicit.
- **Audit of transaction discipline**: all 24 `send`/`send_batch` → `force_tick`/`ticker` sequences already had an explicit `conn.commit()` in between. No additional missing commits were found.
- **New test file `test_transaction_visibility.py`** with two tests:
  1. `test_collapsed_transaction_returns_no_messages`: intentionally collapses `send → force_tick → receive` into one transaction without a commit and asserts `len(received) == 0`. Codifies the PgQ snapshot-isolation visibility contract loudly.
  2. `test_unhandled_event_nack_assertion_catches_stale_cursor`: red/green regression guard. Patches `_poll_once` to a no-op (simulates a consumer that receives but never acks), then verifies the follow-up receive still returns the message — proving the batch cursor did not advance without an explicit ack. Documents what the stronger `follow_up receive returns 0` assertion in the nack tests is protecting against.

## Audited tests

| File | ticker calls fixed | commit-before-tick verified |
|---|---|---|
| `test_receive.py` | 5 → per-queue | all correct |
| `test_send.py` | 5 → per-queue | all correct |
| `test_consumer.py` | 10 → per-queue | all correct |
| `test_consumer_listen_stop.py` | 1 → per-queue | all correct |
| `test_nack.py` | 1 → per-queue | all correct |
| `test_smoke.py` | 1 → per-queue | all correct |

## Before / after pass counts

- **Before** (on main, fresh DB): 54 passed
- **After** (this branch, fresh DB): 56 passed (+2 new tests)

## Test plan

- [ ] Spin up Postgres 16: `docker run --rm -d --name pgque-test -e POSTGRES_PASSWORD=postgres -p 55432:5432 postgres:16`
- [ ] Install schema: `psql "$PGURL" -f sql/pgque.sql`
- [ ] `cd clients/python && pip install -e '.[dev]'`
- [ ] `PGQUE_TEST_DSN="$PGURL" python -m pytest -q tests/` — expect 56 passed

Closes #137


NikolayS:
## REV review — PR #193

### Verdict
Solid tests-only PR with one potential correctness issue in the new collapsed-transaction test and one latent flakiness risk; no blockers.

### Summary
PR switches 23 `pgque.ticker()` (no-arg, ticks all queues) calls to `pgque.ticker(queue)` (per-queue) across 6 test files — confirmed correct: `pgque.ticker(text)` exists in `sql/pgque.sql` as a distinct overload from `pgque.ticker()`. The PR also adds `test_transaction_visibility.py` with two new tests codifying PgQ snapshot-isolation semantics. The transaction-discipline audit (24 send→tick sequences) is confirmed complete by reading every modified file. 54→56 pass count is plausible; no tests appear to have been removed or skipped.

### CI status
All 9 check runs passed: `test (14/15/16/17/18)` ✓, `Python client tests` ✓, `Go client tests` ✓, `TypeScript client tests` ✓, `verify` ✓. Clean green across all supported PG versions.

### Findings

#### Security
none

#### Bugs / correctness

- [NON-BLOCKING] 7/10 — `clients/python/tests/test_transaction_visibility.py`:44–55 — `test_collapsed_transaction_returns_no_messages` calls `conn.rollback()` at the end to clean up, but psycopg's default connection mode (`autocommit=False`) means the entire test runs inside one implicit transaction block. The `conn.rollback()` correctly aborts it. However, if `client.receive()` internally opens a savepoint or uses `conn.transaction()` (a nested block), the rollback will abort the outer transaction and leave the connection in a clean state, as intended. Verified against `consumer.py`: `_poll_once` uses `with conn.transaction()` — but here `receive()` is called directly on `PgqueClient`, not through `Consumer`. If `PgqueClient.receive()` does NOT use a nested `conn.transaction()`, the test is fine. If it does, the `conn.rollback()` after the nested block could raise `psycopg.errors.InFailedSqlTransaction`. Low-probability but worth a comment confirming the assumption.
  - Evidence: `msgs = client.receive(queue, consumer, max_messages=10)` followed by `conn.rollback()` with no explicit savepoint handling.
  - Fix: Add a brief comment: `# conn.rollback() is safe here because PgqueClient.receive() does not open a nested transaction block.`

- [NON-BLOCKING] 6/10 — `clients/python/tests/test_transaction_visibility.py`:101 — The collapsed-transaction test asserts `len(msgs) == 0` which correctly documents the PgQ isolation contract. However, it does not assert the converse: if somehow a tick DID produce a visible batch, the test would fail with a confusing "got N messages" message that could be mistaken for a test framework issue. The existing assertion message is good — this is minor.

#### Tests

- [POTENTIAL] 6/10 — `clients/python/tests/test_transaction_visibility.py`:85–95 — `test_unhandled_event_nack_assertion_catches_stale_cursor` uses `time.sleep(2.0)` to give the patched consumer time to "run" (even though `_poll_once` is a no-op). The consumer's `start()` loop calls `_poll_once` then enters `_wait_for_notify_or_stop()` with `poll_interval=1`. The sleep of 2.0s gives the loop ~2 cycles. This works in CI but is timing-dependent: on a heavily loaded machine the loop might not complete a cycle before the sleep expires, making the follow-up receive non-deterministic. The no-op patch means the cursor provably cannot advance regardless of timing, so the test assertion (`any(m.msg_id == msg_id for m in follow_up)`) is actually timing-safe — the sleep is only needed to let `cons.stop()` be reached. Low actual flake risk given the logic, but the comment should clarify this.
  - Evidence: `time.sleep(2.0)` with `poll_interval=1` — two iterations is not bounded.
  - Fix: Add comment: `# sleep only needs to let stop() be reached; the assertion is timing-safe because _poll_once is a no-op regardless of iteration count.`

- [NON-BLOCKING] 8/10 — `clients/python/tests/test_transaction_visibility.py`:88–89 — `import threading` and `import time` are placed inside the function body (inside the `with mock.patch.object(...)` block). This is unusual and makes the test harder to read; both are already imported at module level in `test_consumer.py`. These should be at the top of the file.
  - Evidence: Lines 88–89: `import threading` / `import time` inside `with mock.patch.object(type(cons), "_poll_once", return_value=None):`
  - Fix: Move both imports to the top of `test_transaction_visibility.py` alongside `from unittest import mock`.

- [NON-BLOCKING] 5/10 — Per-queue ticker switch: no test in the suite exercises the global `pgque.ticker()` (no-arg, all-queues) path directly. This is correct for the test suite's scope (each test owns one queue), but there is no regression guard that `pgque.ticker()` still processes multiple queues. This is out of scope for this PR but worth a follow-up issue.

#### Guidelines (CLAUDE.md)

- [NON-BLOCKING] 9/10 — `clients/python/tests/test_transaction_visibility.py`:1 — New test file has Apache-2.0 copyright header `# Copyright 2026 Nikolay Samokhvalov. Apache-2.0 license.` but does NOT include the PgQ ISC derivation notice. CLAUDE.md §Copyright: "Always preserve the PgQ copyright notice in source headers." The ISC notice is not strictly required in test files that contain no derived PgQ code, but for consistency with `consumer.py` (which does carry it), a brief note is advisable.
  - Fix: Add `# PgQue includes code derived from PgQ (ISC license, Marko Kreen / Skype Technologies OU).` — or a project decision to omit it from test-only files.

- [NON-BLOCKING] 7/10 — Commit subject: `fix(clients/python): tighten test transactions + assertions (#137)` — subject is 58 chars, exceeds the < 50 char limit from CLAUDE.md. Minor; no functional impact.

#### Docs
- The module docstring in `test_transaction_visibility.py` clearly explains PgQ snapshot isolation semantics and why the two-commit ordering is required. Per-function docstrings are present and accurate. No docstring rot found.
- The comment update in `test_consumer_listen_stop.py:79` (`pgque.ticker()` → `pgque.ticker(queue)`) is consistent with the code change.

### Recommended next step
has-potential — two non-blocking items worth addressing before undrafting: (1) move the mid-function imports to module level in the new test file, and (2) add a short comment in `test_collapsed_transaction_returns_no_messages` confirming the rollback assumption.

---

---

NikolayS:
## Lifecycle status — ready for your final look

- **CI**: 9/9 green (PG 14–18 + verify + Go/Python/TS client jobs).
- **REV**: clean — no blocking findings. Two non-blocking polish items (mid-function `import threading`/`import time` could move to module level; clarifying comment on `conn.rollback()` safety in the collapsed-transaction test).
- **Tests**: 54 → 56 passed against fresh PG with `\i sql/pgque.sql`. Per-queue ticker switch correct and complete.

Not merging per your "I'll look myself" instruction. Branch ready when you are.

---

NikolayS:
## Step 3 — real testing

Walked the user-facing claim: fresh PG install → install pgque → `python -m pytest` → 56 passed.

Setup:
```
PG 16 (local), pgque from this branch's sql/pgque.sql
python -m pip install -e 'clients/python[dev]'
```

Result:
```
collected 56 items

clients/python/tests/test_transaction_visibility.py::test_collapsed_transaction_returns_no_messages PASSED [ 98%]
clients/python/tests/test_transaction_visibility.py::test_unhandled_event_nack_assertion_catches_stale_cursor PASSED [100%]

============================= 56 passed in 30.57s ==============================
```

The two new transaction-visibility tests pass:
- `test_collapsed_transaction_returns_no_messages` — proves the bad pattern (no commit between send/tick) returns 0 messages.
- `test_unhandled_event_nack_assertion_catches_stale_cursor` — proves the stronger assertion catches a stale-cursor regression.

Verdict: ✓ The claim holds. Test count up to 56 from 54.

---

PR #203, opened 2026-05-04
fix(clients/python): warn on stale ack (#148)

NikolayS:
Closes the cross-driver Consumer asymmetry that #192 (TS+Go ack rowcount surfacing) left behind in Python.

## Background

PR #192 made TS and Go `client.ack(batchId)` return the integer rowcount from `pgque.finish_batch` (1 success / 0 stale-or-double-ack), and the high-level Consumers in those languages log a `warn` on `0`. Python's `client.ack()` already returned the int, but the high-level `Consumer` at `clients/python/pgque/consumer.py:299` discarded it:

```python
client.ack(batch_id)
```

So a stale/double ack — the same condition TS+Go now log — was silently swallowed by Python's Consumer.

## Fix

Capture the rowcount and log a `WARNING` on `0`, mirroring the wording used in `clients/typescript/src/consumer.ts` and `clients/go/consumer.go`.

```python
if client.ack(batch_id) == 0:
    logger.warning(
        "pgque: ack batch %d returned 0 -- stale or double ack "
        "(batch already finished or not found)",
        batch_id,
    )
```

The 1-line behavior change is additive — successful acks (`n == 1`) are unchanged. Errors (raised from `client.ack`) still propagate as before.

## Test

`tests/test_consumer.py::test_consumer_warns_on_stale_ack` — monkey-patches the per-poll `PgqueClient.ack` to return `0` and asserts a `WARNING` containing "stale" / "double" appears in `caplog`.

## TDD

```
$ # RED on origin/main:
$ python -m pytest tests/test_consumer.py::test_consumer_warns_on_stale_ack -x -v
AssertionError: expected a WARNING mentioning stale or double ack; got []

$ # GREEN on this branch:
PASSED

$ # Full suite:
$ python -m pytest -q
55 passed in 31.88s
```

## Out of scope

- The actual stale-ack scenario (race between two consumers ackking the same batch) is not exercised here — that requires a 2-session live test. The unit test locks the warn-log behavior; integration coverage can come later.

Refs #148.

https://claude.ai/code/session_01TwKyarGpkGKN8LY1CV8dsG

---
_Generated by [Claude Code](https://claude.ai/code/session_01TwKyarGpkGKN8LY1CV8dsG)_

### DIFF
--- a/clients/python/pgque/consumer.py
+++ b/clients/python/pgque/consumer.py
@@ -296,4 +296,13 @@ class Consumer:
                 # Do NOT ack -- redeliver on next poll.
                 return
 
-            client.ack(batch_id)
+            # pgque.ack returns 1 on success, 0 if the batch was already
+            # finished or not found (stale/double ack, cross-consumer
+            # race). Mirror the TS+Go consumers and log warn on 0; do
+            # not treat it as an error.
+            if client.ack(batch_id) == 0:
+                logger.warning(
+                    "pgque: ack batch %d returned 0 -- stale or double ack "
+                    "(batch already finished or not found)",
+                    batch_id,
+                )
--- a/clients/python/tests/test_consumer.py
+++ b/clients/python/tests/test_consumer.py
@@ -407,6 +407,59 @@ def test_consumer_does_not_ack_when_handler_error_nack_fails(
         conn.commit()
 
 
+def test_consumer_warns_on_stale_ack(dsn, conn, setup_queue, caplog):
+    """If ``client.ack(batch_id)`` returns 0 (stale/double ack), the
+    Consumer must log a WARNING -- matching the TS+Go consumers, which
+    log a warn on ``n == 0`` from the new ack-rowcount surface.
+
+    Simulated by monkey-patching the per-poll ``PgqueClient`` so its
+    ``ack`` returns 0 instead of delegating to the SQL function. The
+    test does not exercise the real stale path (which requires a race
+    between two consumers); it locks the warn-log behavior.
+    """
+    queue, consumer_name = setup_queue
+    client = pgque.PgqueClient(conn)
+    client.send(queue, {"x": 1}, type="order.created")
+    conn.commit()
+    conn.execute("select pgque.force_tick(%s)", (queue,))
+    conn.execute("select pgque.ticker()")
+    conn.commit()
+
+    cons = pgque.Consumer(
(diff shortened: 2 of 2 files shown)


NikolayS:
## REV review

Small, focused cross-driver follow-up to the TS+Go ack-rowcount work. The 10-line block in `consumer.py` correctly checks the existing int return from `client.ack` and emits a warn-level log via the module-level `logger = logging.getLogger("pgque")`, which matches the test's `caplog.at_level(logging.WARNING, logger="pgque")`. Exception semantics on `client.ack` are preserved: only the `== 0` branch is intercepted, raised `psycopg.Error`-derived exceptions still propagate as before. The new test monkey-patches `PgqueClient.__init__` via `mock.patch.object` as a context manager, so the patch is torn down on `__exit__` even if the inner block raises -- no leakage into adjacent tests. Commit subjects are 45 and 49 chars (both under 50), conventional types `fix:` and `test:`. No README/`docs/` diff, no PR or version refs in the new comments/docstrings. CLAUDE.md compliance is clean.

None.

### Non-blocking
- **Cross-driver wording parity (docs-reviewer, conf 5)** -- `clients/python/pgque/consumer.py:305-308`. Python emits `"pgque: ack batch %d returned 0 -- stale or double ack ..."` while TS uses `"... returned 0 — ..."` (em-dash) and Go uses `"... returned 0 rows — ..."` (em-dash + "rows"). All three communicate the same thing, but the punctuation and "rows" diverge. If you want strict parity, consider em-dash and adding "rows" to match Go (Go's wording is the most explicit). Not required -- the message is unambiguous and the test asserts on `"stale"`/`"double"`, which is robust to either choice.
- **Test assertion strength (test-analyzer, conf 4)** -- `clients/python/tests/test_consumer.py:454-459`. The test asserts `any(... "stale" or "double" ...)` across all WARNING records and only checks `ack_returns` is non-empty. It does not assert (a) exactly one stale-ack warning per zero return, (b) the formatted `batch_id` appears in the message, or (c) the warning came from the `pgque` logger specifically (caplog filters by level globally, the `logger="pgque"` arg only sets the level). Tightening to e.g. `assert sum(1 for r in caplog.records if r.name == "pgque" and "stale" in r.getMessage()) == len(ack_returns)` would lock the contract more tightly. Optional.

### Skipped (SOC2)
Skipped per repo policy.

Verdict: APPROVE

---

PR #215, opened 2026-05-06
feat(clients/go): experimental cooperative consumers

NikolayS:
## Summary

- Adds the **experimental cooperative consumer** client API to the Go driver, wrapping the SQL helpers shipped on `main` (`pgque.subscribe_subconsumer`, `pgque.unsubscribe_subconsumer`, `pgque.receive_coop`, `pgque.touch_subconsumer`).
- Adds `Consumer` options `WithSubconsumer(name)` and `WithDeadInterval(d)` so the high-level poll loop branches onto the cooperative path; without them the consumer behaves exactly as today.
- Documents the API as experimental in package docs, the client README, and the cross-client parity matrix.

## Experimental notice

> **Experimental in PgQue 0.2.** Function names, edge-case behavior, and client API shape may change before this feature is marked stable. Do not use this as the only processing path for critical workloads without idempotent handlers and stale-worker takeover tests.

## API table

| Surface | Wraps / Behavior |
| --- | --- |
| `Client.SubscribeSubconsumer(ctx, q, c, sc)` | `pgque.subscribe_subconsumer` — returns 1 on first call, 0 when already registered |
| `Client.UnsubscribeSubconsumer(ctx, q, c, sc, opts...)` | `pgque.unsubscribe_subconsumer` — default refuses if active batch; `WithBatchHandlingRetry()` routes through retry/DLQ |
| `Client.ReceiveCoop(ctx, q, c, sc, opts...)` | `pgque.receive_coop` — auto-registers main + subconsumer; `WithCoopMaxMessages(n)` (default 100) and `WithCoopDeadInterval(d)` |
| `Client.TouchSubconsumer(ctx, q, c, sc)` | `pgque.touch_subconsumer` heartbeat |
| `Consumer` opt `WithSubconsumer(name)` | High-level: switches the poll loop to `ReceiveCoop` |
| `Consumer` opt `WithDeadInterval(d)` | High-level: passes a stale-worker takeover window |
| `Consumer` opt `WithBatchHandlingRetry()` | `UnsubscribeSubconsumer` only — sets `batch_handling=1` |

Heartbeats are **not** auto-emitted from the poll loop; call `Client.TouchSubconsumer` manually if you want to advertise liveness ahead of a dead-interval takeover.

## Manual two-worker e2e (evidence)

Two `Consumer` goroutines, each with a different `WithSubconsumer` name, share one logical consumer. Forced ticks scatter messages across batches; each handler must observe a disjoint subset and the union must equal everything published.

```
[main] published 12 messages across 12 ticks
[worker-1] msg_id=1 batch_id=68 payload={"i": 0}
[worker-2] msg_id=2003 batch_id=69 payload={"i": 1}
[worker-1] msg_id=4005 batch_id=70 payload={"i": 2}
[worker-2] msg_id=6007 batch_id=71 payload={"i": 3}
[worker-1] msg_id=8009 batch_id=72 payload={"i": 4}
[worker-2] msg_id=10011 batch_id=73 payload={"i": 5}
[worker-1] msg_id=12013 batch_id=74 payload={"i": 6}
[worker-2] msg_id=14015 batch_id=75 payload={"i": 7}
[worker-1] msg_id=16017 batch_id=76 payload={"i": 8}
[worker-2] msg_id=18019 batch_id=77 payload={"i": 9}
[worker-1] msg_id=20021 batch_id=78 payload={"i": 10}
[worker-2] msg_id=22023 batch_id=79 payload={"i": 11}

[main] total delivered = 12 (want 12)
[main] worker-1 saw 6 messages
[main] worker-2 saw 6 messages
[main] OK: all 12 messages delivered exactly once across two workers
```

Each batch_id is distinct; worker-1 saw 6, worker-2 saw 6, no duplicates, no losses.

## Test plan

- [x] `cd clients/go && go vet ./...` clean
- [x] `cd clients/go && go test -count=1 -short .` (unit suite, no DB) passes
- [x] `PGQUE_TEST_DSN=postgres://nik@localhost/pgque_coop_go go test -count=1 ./...` passes (covers `SubscribeSubconsumer` idempotency, `ReceiveCoop` round-trip, two-subconsumer split, `UnsubscribeSubconsumer` active-batch guard + `WithBatchHandlingRetry`, `TouchSubconsumer` rowcount, high-level `Consumer` dispatch with `WithSubconsumer`, smoke test for non-coop path, regression guard for normal `Receive` on a coop main)
- [x] Manual two-worker e2e against a fresh queue (output above)

## Notes / concerns

- The `consumerBackend` interface stays unchanged. `ReceiveCoop` lives on a sibling `coopBackend` interface that the `Client` satisfies; the poll loop type-asserts only when `WithSubconsumer` is set, so existing stub-based tests keep compiling without edits.
- The high-level `Consumer` passes its configured `maxMessages` (default `math.MaxInt32`) into `ReceiveCoop`. The SQL function only validates `max_return >= 1` so this is fine, but lowering `maxMessages` below the real batch size remains a footgun (the same one as plain `Receive`).
- Heartbeats are intentionally manual; auto-heartbeat can be added later as an option once the contract stabilizes.


NikolayS:
## REV Code Review Report

- **PR:** [NikolayS/PgQue#215](https://github.com/NikolayS/PgQue/pull/215) - feat(clients/go): experimental cooperative consumers
- **Author:** @NikolayS
- **AI-Assisted:** No
- **CI:** ✅ success

---

### POTENTIAL ISSUES (4)

Issues with moderate confidence (4–7/10). Review manually — may be false positives.

**MEDIUM** `clients/go/coop_integration_test.go:350` - Test cleanup uses `drop_queue` without `force=true`, but subconsumers may still be registered at teardown *(confidence: 7/10)*
> `client.Pool().Exec(ctx, "select pgque.drop_queue($1)", queue)` — compare with `bench/coop_demo/main.go` which explicitly uses `drop_queue($1, true)` precisely because subconsumers remain. If PgQue's `drop_queue` raises when subscriptions exist, teardown fails silently, leaving stale state across tests.
> **Suggestion:** Change `setupFreshCoopGroup` cleanup to `drop_queue($1, true)` (matching the demo pattern), or explicitly unsubscribe all subconsumers before dropping.

**MEDIUM** `clients/go/consumer.go:294` - `coopBackend` type assertion repeated on every poll-loop iteration *(confidence: 6/10)*
> `coop, ok := c.backend.(coopBackend)` — this runs each tick. If the backend doesn't implement `ReceiveCoop`, the consumer spins logging errors until cancelled. Validating once at `NewConsumer` construction time would fail fast and eliminate per-tick overhead.
> **Suggestion:** Assert `coopBackend` in `NewConsumer` when `WithSubconsumer` is set and store the result, or return an error from `NewConsumer` if the backend is incompatible.

**MEDIUM** `clients/go/coop_integration_test.go:745` - `TestConsumer_TwoSubconsumers_DisjointDelivery` issues `tick()` inside the consumer polling loop *(confidence: 6/10)*
> The ticker goroutine issues ticks while both workers are concurrently consuming and acking. Under load, a tick can arrive before the previous ack commits, causing the allocator to skip a batch. The test could report fewer delivered messages than expected and fail spuriously.
> **Suggestion:** Pre-publish all ticks before starting consumers (as done in `TestConsumer_WithSubconsumer_DispatchesAndAcks`), or increase the expected-total tolerance.

**LOW** `clients/go/bench/coop_demo/main.go:151` - Cleanup defer uses raw SQL instead of the Go client API *(confidence: 5/10)*
> `client.Pool().Exec(bg, "select pgque.unsubscribe_subconsumer($1, $2, $3, 1)", ...)` is called in the cleanup `defer` instead of `client.UnsubscribeSubconsumer(..., pgque.WithBatchHandlingRetry())`. The demo is otherwise a showcase of the Go client API.
> **Suggestion:** Replace raw SQL with the wrapper method to demonstrate the complete API and keep the demo consistent with the README examples.

---

### Summary

| CI/Actions | 0 | 0 | 0 |
| Security | 0 | 0 | 0 |
| Bugs | 0 | 2 | 0 |
| Tests | 0 | 1 | 0 |
| Guidelines | 0 | 0 | 1 |
| Docs | 0 | 1 | 0 |
| Metadata | 0 | 0 | 0 |

---

PR #219, opened 2026-05-06
feat: add pg_timetable scheduler support

NikolayS:
## Summary

Adds first-class pg_timetable lifecycle support alongside pg_cron:

- `pgque.start_timetable(ticks_per_second default 10)` schedules PgQue jobs in pg_timetable
- `pgque.stop_timetable()` removes the PgQue pg_timetable jobs
- default ticker cadence remains 10/s via existing `CALL pgque.ticker_loop()`
- stores active scheduler in `pgque.config.scheduler`
- updates README and generated install artifacts

## Testing

- `./build/transform.sh`
- fresh install + `tests/test_pgtimetable_lifecycle.sql`
- fresh install + `tests/test_tick_period.sql`
- full fresh-install SQL regression: `tests/run_all.sql`
- fake pg_timetable schema/API smoke test verifying four scheduled jobs and `CALL pgque.ticker_loop()` command

## Notes

The pg_timetable path uses `@every 1 second` and delegates sub-second cadence to `pgque.ticker_loop()`, preserving the per-tick transaction boundary with `COMMIT` between iterations.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -231,9 +231,109 @@ jobs:
             exit 1
           fi
 
+      - name: Verify pg_timetable to pg_cron scheduler switch
+        run: |
+          set -Eeuo pipefail
+          # Run the pg_timetable lifecycle file directly in the pg_cron image.
+          # This creates the fake pg_timetable schema, starts PgQue with it,
+          # switches back to pg_cron, and must exercise the non-SKIP switch path.
+          out=$(PGPASSWORD=pgque_test psql -h localhost -U postgres -d pgque_test \
+            -v ON_ERROR_STOP=1 -f tests/test_pgtimetable_lifecycle.sql 2>&1)
+          echo "$out"
+          if echo "$out" | grep -q 'SKIP: pg_cron not installed (cannot test pg_timetable -> pg_cron switch)'; then
+            echo "FAIL: pg_timetable -> pg_cron switch test skipped despite pg_cron being installed"
+            exit 1
+          fi
+          if ! echo "$out" | grep -q 'PASS: start() switches pg_timetable to pg_cron cleanly'; then
+            echo "FAIL: pg_timetable -> pg_cron switch assertion did not run"
+            exit 1
+          fi
+
       - name: Cleanup
         if: always()
         run: docker rm -f pgque-pgcron-test
+
+
+  pgtimetable-real-test:
+    name: pg_timetable real worker
+    runs-on: ubuntu-latest
+    env:
+      PG_VERSION: '18'
+      PGTT_IMAGE: cybertecpostgresql/pg_timetable@sha256:6bed16f944ab91392fac9cd04aba82cd6690769e179d411748fd0bb538f55acd
+      PGTT_NET: pgque-pgtimetable-test
+      PG_CONTAINER: pgque-pgtimetable-pg
+      PGTT_CONTAINER: pgque-pgtimetable-worker
+      PGPASSWORD: pgque_test
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          submodules: recursive
+
+      - name: Build pgque
+        run: bash build/transform.sh
+
+      - name: Start PostgreSQL 18
+        run: |
+          set -Eeuo pipefail
+          docker network create "${PGTT_NET}"
+          docker run -d --name "${PG_CONTAINER}" --network "${PGTT_NET}" \
+            -e POSTGRES_PASSWORD=pgque_test \
+            -e POSTGRES_DB=pgque_test \
+            postgres:${PG_VERSION}
+
+          ready=0
+          for _ in $(seq 1 60); do
+            if docker exec "${PG_CONTAINER}" pg_isready -U postgres -d pgque_test; then
+              ready=1
+              break
+            fi
+            sleep 1
+          done
+          if [[ "${ready}" -ne 1 ]]; then
+            echo "PG not ready after 60 seconds"
+            docker logs "${PG_CONTAINER}"
+            exit 1
+          fi
+
+      - name: Initialize real pg_timetable schema
+        run: |
+          set -Eeuo pipefail
+          docker run --rm --network "${PGTT_NET}" "${PGTT_IMAGE}" \
+            --init \
+            --connstr="postgresql://postgres:pgque_test@${PG_CONTAINER}:5432/pgque_test" \
+            --clientname=pgque
+
+      - name: Install pgque and schedule pg_timetable jobs
+        run: |
+          set -Eeuo pipefail
+          docker cp sql/pgque.sql "${PG_CONTAINER}":/tmp/pgque.sql
+          docker cp tests/test_pgtimetable_real_setup.sql "${PG_CONTAINER}":/tmp/test_pgtimetable_real_setup.sql
+          docker exec -e PGPASSWORD "${PG_CONTAINER}" \
+            psql -U postgres -d pgque_test -v ON_ERROR_STOP=1 -f /tmp/pgque.sql
+          docker exec -e PGPASSWORD "${PG_CONTAINER}" \
+            psql -U postgres -d pgque_test -v ON_ERROR_STOP=1 -f /tmp/test_pgtimetable_real_setup.sql
+
+      - name: Start real pg_timetable worker
+        run: |
+          set -Eeuo pipefail
+          docker run -d --name "${PGTT_CONTAINER}" --network "${PGTT_NET}" "${PGTT_IMAGE}" \
+            --connstr="postgresql://postgres:pgque_test@${PG_CONTAINER}:5432/pgque_test" \
+            --clientname=pgque \
+            --log-level=debug
+
+      - name: Verify real pg_timetable executes PgQue ticker_loop
+        run: |
+          set -Eeuo pipefail
+          docker cp tests/test_pgtimetable_real_assert.sql "${PG_CONTAINER}":/tmp/test_pgtimetable_real_assert.sql
+          docker exec -e PGPASSWORD "${PG_CONTAINER}" \
+            psql -U postgres -d pgque_test -v ON_ERROR_STOP=1 -f /tmp/test_pgtimetable_real_assert.sql
+
+      - name: Cleanup
+        if: always()
+        run: |
+          docker logs "${PGTT_CONTAINER}" || true
+          docker rm -f "${PGTT_CONTAINER}" "${PG_CONTAINER}" || true
+          docker network rm "${PGTT_NET}" || true
 
   python-client-tests:
     name: Python client tests
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 <h1 align="center">PgQue – PgQ, universal edition</h1>
 
-<p align="center"><strong>Zero-bloat Postgres queue. One SQL file to install, <code>pg_cron</code> to tick.</strong></p>
+<p align="center"><strong>Zero-bloat Postgres queue. One SQL file to install, <code>pg_cron</code> or <code>pg_timetable</code> to tick.</strong></p>
 
 <p align="center">
   <a href="https://github.com/NikolayS/pgque/actions/workflows/ci.yml"><img src="https://github.com/NikolayS/pgque/actions/workflows/ci.yml/badge.svg" alt="CI"></a>
@@ -150,17 +150,33 @@ Or from the shell, same single-transaction guarantee via `psql --single-transact
 PAGER=cat psql --no-psqlrc --single-transaction -d mydb -f sql/pgque.sql
 ```
 
-With `pg_cron` available in the same database as PgQue, `pgque.start()` creates the default ticker and maintenance jobs:
+With `pg_cron` available in the same database as PgQue, `pgque.start()` creates the default ticker and maintenance jobs. The ticker uses a one-second pg_cron slot and calls `pgque.ticker_loop()`, which ticks every 100 ms by default (10 ticks/sec) with a commit between ticks:
 
 ```sql
 select pgque.start();
 ```
 
+With `pg_timetable`, run the external pg_timetable worker against the database where PgQue is installed, then schedule PgQue with the same 10 ticks/sec default. For example:
+
+```bash
+pg_timetable --dbname=mydb --clientname=pgque
+```
+
+The `--clientname=pgque` flag is required; pg_timetable only executes chains whose `job_client_name` matches the running worker's client name. Keep that worker running; unlike `pg_cron`, pg_timetable is an external scheduler process, not a Postgres extension background worker. See the [pg_timetable docs](https://github.com/cybertec-postgresql/pg_timetable) for production service setup.
+
+```sql
+select pgque.start_timetable();      -- default: 10 ticks/sec
+-- or explicitly:
+select pgque.start_timetable(10);
+```
+
+`pgque.stop_timetable()` removes the PgQue pg_timetable jobs. `pgque.stop()` also stops whichever PgQue scheduler is active. Calling `pgque.start_timetable()` automatically removes existing PgQue `pg_cron` jobs first; `pgque.start()` does the same for PgQue pg_timetable jobs. `pgque.set_tick_period_ms(ms)` still controls the loop cadence; for example `100` means 10 ticks/sec, `200` means 5 ticks/sec, and `1000` means 1 tick/sec.
+
 ### Tick rate
 
-PgQue ticks **10 times per second by default** (every 100 ms), even though `pg_cron`'s minimum schedule is 1 second. `pgque.start()` schedules a single 1-second pg_cron slot that calls `CALL pgque.ticker_loop()`; the procedure then re-invokes `pgque.ticker()` every `tick_period_ms` ms inside that slot, committing between iterations so each tick gets its own transaction (snapshot semantics; bounded held-xmin so rotation isn't blocked).
+PgQue ticks **10 times per second by default** (every 100 ms), even though `pg_cron`'s minimum schedule is 1 second. `pgque.start()` schedules a single 1-second pg_cron slot that calls `CALL pgque.ticker_loop()`; `pgque.start_timetable()` schedules the same loop through one pg_timetable `@every 1 second` job. The procedure then re-invokes `pgque.ticker()` every `tick_period_ms` ms inside that slot, committing between iterations so each tick gets its own transaction (snapshot semantics; bounded held-xmin so rotation isn't blocked).
 
-Tune at runtime — no need to call `start()` again, the change picks up on the next pg_cron slot (≤1 s):
+Tune at runtime — no need to call `start()` again, the change picks up on the next scheduler slot (≤1 s):
 
 ```sql
 select pgque.set_tick_period_ms(50);    -- 20 ticks/sec, ~25 ms median e2e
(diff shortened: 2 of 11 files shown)


NikolayS:
## REV Code Review Report

- **PR:** NikolayS/PgQue#219 — feat: add pg_timetable scheduler support
- **Author:** @NikolayS
- **CI Status:** ✅ success (11/11 checks passed)
- **Draft:** yes

---

### BLOCKING ISSUES (6)

Issues that must be addressed before merge.

---

**HIGH** `sql/pgque-additions/lifecycle.sql` — `stop_timetable()` WHERE clause destroys pg_cron job IDs on upgrade, causing double-scheduling

> After the migration adds `scheduler text` as NULL for existing rows, any installation that had pg_cron active has `scheduler = NULL` and `ticker_job_id = <real_id>`. When the user next calls `pgque.start()`, it calls `stop_timetable()` first, which matches the `scheduler is null` arm and NULLs out both job IDs. Then `stop()` reads those NULL IDs and skips `cron.unschedule()` — old pg_cron jobs keep running. `start()` then schedules new jobs. Result: ticker, retry, maint, and rotate_step2 all run twice.
>
> **Fix:** Change the WHERE clause in `stop_timetable()` from `where scheduler is null or scheduler = 'pg_timetable'` to `where scheduler = 'pg_timetable'`. There is nothing for stop_timetable to clean up when scheduler is NULL — either the IDs are already null (fresh install) or they belong to pg_cron which `stop()` handles.

---

**HIGH** `tests/test_pgtimetable_lifecycle.sql` — No test for invalid `ticks_per_second` values

> The validation block `(< 1 or > 1000 or 1000 % i_ticks_per_second <> 0)` is entirely uncovered. Values like `0`, `-1`, `7` (non-divisor), `1001`, and explicit `NULL` all raise an exception but are never exercised.
>
> **Fix:** Add a test block that calls `start_timetable(0)`, `start_timetable(7)`, `start_timetable(1001)`, and `start_timetable(NULL)` inside `BEGIN/EXCEPTION` and asserts each raises with a message mentioning the bad value.

---

**HIGH** `tests/test_pgtimetable_lifecycle.sql` — No test for the scheduler-switch path (pg_timetable → pg_cron)

> `pgque.start()` now calls `stop_timetable()` before `stop()`, but no test verifies that switching from pg_timetable back to pg_cron fully cleans up timetable state and correctly re-schedules pg_cron jobs.
>
> **Fix:** Add a test: call `start_timetable(10)`, then `start()` (pg_cron path), then assert `scheduler='pg_cron'`, no timetable chain rows remain, and pg_cron jobs were scheduled.

---

**HIGH** `sql/pgque-additions/lifecycle.sql` (stop function) — `stop()` silently skips clearing config when `scheduler='pg_timetable'`, and this behavior is untested

> `stop()` has `WHERE scheduler IS NULL OR scheduler = 'pg_cron'`, which means calling `stop()` while pg_timetable is active leaves `config.scheduler` set to `'pg_timetable'` and job IDs intact. An operator calling `stop()` directly expecting a clean state will be surprised.
>
> **Fix:** Add a test that calls `start_timetable(10)` then `stop()`, then asserts on the resulting config state — to either confirm this is intentional (pg_cron's stop is a no-op for pg_timetable) or reveal it as a bug.

---

**MEDIUM** `sql/pgque-additions/lifecycle.sql` — `pgque.status()` output is a breaking API change

> Old code returned separate rows with `component='ticker'` and `component='maintenance'`; the new code collapses them into a single `component='scheduler'` row. Any script or monitoring tool parsing `pgque.status()` by component name will silently get wrong data — no error, just missing rows.
>
> **Fix:** Either keep backward-compatible `'ticker'` and `'maintenance'` rows alongside the new `'scheduler'` row, or document this as a breaking change in the PR description and README and bump the version accordingly.

---

**MEDIUM** `README.md` — pg_timetable setup instructions too vague

> "Initialize pg_timetable in the database, run its worker" provides no concrete commands. Unlike pg_cron (a Postgres extension), pg_timetable is an external binary that must stay running as a daemon. This is a non-obvious operational difference.
>
> **Fix:** Add the concrete invocation (e.g. `pg_timetable --dbname=mydb --clientname=pgque`) or a link to pg_timetable docs, and note that the binary must remain running. The current wording implies it's a one-time SQL step.

---

### POTENTIAL ISSUES (12)

Moderate-confidence findings — review manually, may be context-dependent.

**MEDIUM** `sql/pgque-additions/lifecycle.sql` — SECURITY DEFINER functions trust externally-creatable `timetable` schema *(confidence: 6/10)*
> `start_timetable()` and `stop_timetable()` are `security definer` and call schema-qualified `timetable.add_job()` / `timetable.delete_job()` via EXECUTE. In a shared database environment, a user with `CREATE SCHEMA` privilege can create their own `timetable` schema matching the API signature (the test file is a complete recipe) and have those functions run under the function owner's privileges. The `set search_path = pgque, pg_catalog` doesn't help here since the calls are schema-qualified.
> **Suggestion:** `REVOKE EXECUTE ON FUNCTION pgque.start_timetable(integer), pgque.stop_timetable() FROM PUBLIC;` and grant only to the intended operator role.

**MEDIUM** `tests/test_pgtimetable_lifecycle.sql` — No idempotency test for `start_timetable()` called twice *(confidence: 7/10)*
> **Suggestion:** Call `start_timetable(10)` twice, then assert `count(*) = 4` in `timetable.chain` for pgque job names.

**MEDIUM** `README.md` — pg_timetable log retention not mentioned alongside pg_cron hygiene section *(confidence: 7/10)*
> The "pg_cron log hygiene" section warns about `cron.job_run_details` growing forever. pg_timetable also accumulates execution history; the README is silent on whether the same concern applies.
> **Suggestion:** Add a note clarifying pg_timetable's log retention behavior, or confirm it auto-purges.

**MEDIUM** `tests/test_pgtimetable_lifecycle.sql` — No test for the new `pgque.status()` output format *(confidence: 6/10)*
> **Suggestion:** Add assertions against `select * from pgque.status()` after `start_timetable()`, verifying the `scheduler` row shows `'pg_timetable'`.

**LOW** `sql/pgque-additions/lifecycle.sql` (stop) — Stale comment says "regardless" but UPDATE now has a WHERE clause *(confidence: 5/10)*
> `-- Clear job IDs regardless (even if pg_cron is gone)` is now misleading.
> **Suggestion:** Change to `-- Clear pg_cron job IDs (only when scheduler is pg_cron or unset)`.

**LOW** `sql/pgque-additions/lifecycle.sql` (start_timetable) — `to_regprocedure` pins to a specific 12-argument overload; a newer pg_timetable with a different signature would break even if the named-argument call would still work *(confidence: 4/10)*
> **Suggestion:** Consider `to_regproc('timetable.add_job') is null` — the subsequent `execute` will fail with a clear error if arguments are truly incompatible.

**LOW** `tests/test_pgtimetable_lifecycle.sql` — Fake schema persists if test assertions raise inside DO blocks *(confidence: 5/10)*
> **Suggestion:** Wrap teardown in `DO $$ BEGIN DROP SCHEMA IF EXISTS timetable CASCADE; EXCEPTION WHEN OTHERS THEN NULL; END $$`.

**INFO** `sql/pgque-additions/lifecycle.sql` — `v_retry_id` and `v_step2_id` captured from `add_job()` but not stored in config *(confidence: 7/10)*
> Only `ticker_job_id` and `maint_job_id` are persisted; the other two IDs are used only in `RAISE NOTICE`. Since `stop_timetable()` deletes by name (not ID), storing them isn't strictly needed — but consider either documenting why, or dropping the variables entirely.

**INFO** `README.md` — Public API documented as `ticks_per_second` but actual parameter is `i_ticks_per_second` *(confidence: 6/10)*
> Named-argument callers need `pgque.start_timetable(i_ticks_per_second => 5)`, not `ticks_per_second => 5`.
> **Suggestion:** Align README with the actual parameter name, or drop the `i_` prefix.

**INFO** `README.md` — Scheduler-switch auto-cleanup not documented *(confidence: 6/10)*
> **Suggestion:** Add one sentence: "Calling `pgque.start_timetable()` automatically removes any existing pg_cron PgQue jobs first; `pgque.start()` does the same for pg_timetable jobs."

**INFO** `README.md` — Sub-second ticking mechanism not explained for pg_timetable *(confidence: 5/10)*
> The pg_cron section clearly states it uses a single 1-second slot + `ticker_loop()` internally; the pg_timetable section omits this.
> **Suggestion:** Add: "As with pg_cron, pg_timetable uses a single `@every 1 second` slot and delegates sub-second cadence to `pgque.ticker_loop()`."

**INFO** `sql/pgque-additions/lifecycle.sql` — EXECUTE $sql$ blocks have no comment explaining why dynamic SQL is needed *(confidence: 4/10)*
> **Suggestion:** Add: `-- Dynamic SQL: timetable schema may not exist at function-definition time`

---

### Summary

| CI/Pipeline | 0 | 0 | 0 |
| Security | 0 | 1 | 0 |
| Bugs | 2 | 2 | 0 |
| Tests | 4 | 2 | 0 |
| Guidelines | 0 | 3 | 0 |
| Docs | 1 | 3 | 0 |
| Metadata | 0 | 0 | 0 |

---

NikolayS:
## REV Code Review Report

- **PR:** NikolayS/PgQue#219 — feat: add pg_timetable scheduler support
- **Author:** @NikolayS
- **CI Status:** ✅ success (11/11 checks passed)
- **Draft:** yes

All 6 blocking issues from the previous review have been addressed in this update. The PR is now in good shape.

---

### POTENTIAL ISSUES (5)

No blocking issues. The items below are low-severity for awareness.

**INFO** `sql/pgque-additions/lifecycle.sql` — `start()` calls `stop_timetable()` before `stop()`, but `stop()` now delegates to `stop_timetable()` when `scheduler='pg_timetable'` — double-call on pg_timetable→pg_cron switch *(confidence: 5/10)*
> Harmless (second call is a no-op after config is cleared), but redundant. Could remove the explicit `perform pgque.stop_timetable()` from `start()` and rely on `stop()` delegation alone.

**MEDIUM** `tests/test_pgtimetable_lifecycle.sql:1282` — Test 1b (untrusted schema owner) requires `CREATE ROLE` privilege *(confidence: 6/10)*
> The test executes `CREATE ROLE pgque_fake_timetable_owner` which requires `CREATEROLE` or superuser. If CI runs tests as a restricted database user, Test 1b will fail with a permission error, not the expected `untrusted pg_timetable schema owner` message.
> **Suggestion:** Guard the test with `IF current_setting('is_superuser') = 'on' OR pg_has_role(session_user, 'pg_role_with_createrole', 'member')`, or wrap in an outer `EXCEPTION WHEN insufficient_privilege` that skips with a NOTICE.

**INFO** `sql/pgque-additions/lifecycle.sql` — Security ownership check is duplicated between `start_timetable()` and `stop_timetable()` *(confidence: 4/10)*
> The 10-line ownership validation block appears verbatim in both functions. A private helper function (e.g. `pgque._assert_timetable_trusted()`) would be cleaner, but this is a minor maintenance concern.

**INFO** `sql/pgque-additions/lifecycle.sql` — `v_retry_id` and `v_step2_id` captured from `add_job()` but not stored in config *(confidence: 7/10)*
> Only `ticker_job_id` and `maint_job_id` are persisted; the retry and rotate_step2 IDs appear only in `RAISE NOTICE`. Since `stop_timetable()` deletes by name (not ID), this has no operational impact. Consider either documenting this explicitly or dropping the unused variables.

**INFO** `README.md` — Public API documented as `ticks_per_second` but actual parameter name is `i_ticks_per_second` *(confidence: 6/10)*
> Named-argument callers would need `pgque.start_timetable(i_ticks_per_second => 5)`, not `ticks_per_second => 5`. Consider dropping the `i_` prefix from the parameter to match the documentation style.

---

### Summary

| CI/Pipeline | 0 | 0 | 0 |
| Security | 0 | 0 | 0 |
| Bugs | 0 | 0 | 0 |
| Tests | 0 | 1 | 0 |
| Guidelines | 0 | 3 | 0 |
| Docs | 0 | 1 | 0 |

**Result: PASSED** — no blocking issues.

---

NikolayS:
## REV Code Review Report

- **PR:** NikolayS/PgQue#219 — feat: add pg_timetable scheduler support
- **Author:** @NikolayS

**CI Status:** ✅ success (12/12 checks passed)

---

### BLOCKING ISSUES (2)

Issues that must be addressed before merge.

**HIGH** `tests/test_pgtimetable_lifecycle.sql:1634` — pg_timetable→pg_cron scheduler switch test always skipped in CI
> Test 6 checks `if not exists (select 1 from pg_extension where extname = 'pg_cron')` — it only runs when pg_cron is installed. But the fake timetable harness is set up and torn down within the lifecycle test file itself, so on the `pg_cron install path` CI job, the fake schema doesn't exist when Test 6 runs. Result: the pg_timetable→pg_cron direction (highest double-scheduling risk) has zero automated coverage.
> **Fix:** Add a variant of Test 6 to the `pg_cron install path` CI job that installs the fake timetable schema, calls `start_timetable()`, then `start()`, and asserts `scheduler='pg_cron'` with zero timetable jobs remaining.

**MEDIUM** `README.md:~120` — `--clientname=pgque` requirement not explained
> The README shows `pg_timetable --dbname=mydb --clientname=pgque` but doesn't say why `--clientname=pgque` is mandatory. Operators who omit it get a silently broken setup — pg_timetable won't execute PgQue chains because it only runs chains whose `job_client_name` matches the worker's client name.
> **Fix:** Add one sentence after the CLI example: "The `--clientname=pgque` flag is required; pg_timetable only executes chains whose `job_client_name` matches the running worker's client name."

---

### NON-BLOCKING (2)

Minor issues. Can be addressed after merge.

**LOW** `README.md:~150` — `pgque.status()` output format change undocumented
> The new `scheduler` summary row in `status()` output changes row count from 1–2 to always ≥3. Scripts that parse `status()` by row count or order will break silently.
> **Suggestion:** Add a note showing the new `status()` output, or call out the change in release notes.

**INFO** `sql/pgque-additions/lifecycle.sql` — `status()` row count change under-documented in code
> Inline comment says "backward-compatible rows retained" but doesn't mention the additive `scheduler` row.
> **Suggestion:** Add `-- NOTE: status() now always returns ≥3 rows regardless of scheduler` near the function signature.

---

### POTENTIAL ISSUES (14)

Medium-confidence findings (4–7/10). Review manually — may be false positives.

**LOW** `tests/test_pgtimetable_real_assert.sql:14` — `exit when v_count = 1` hangs if count exceeds 1 *(confidence: 7/10)*
> If `receive()` returns 2 messages in one batch (e.g. leftover state), `v_count` becomes 2, the exit never fires, the loop runs all 60 iterations, and the final assert fails with the misleading message "real pg_timetable worker did not deliver event" — when it delivered too many.
> **Suggestion:** Change to `exit when v_count >= 1` and update the assert message accordingly.

**LOW** `sql/pgque-additions/lifecycle.sql` — `start_timetable()` generates pg_cron error-log noise on dual-scheduler environments *(confidence: 6/10)*
> After `stop_timetable()` nulls `scheduler`, the subsequent `stop()` call enters the pg_cron branch and tries to unschedule non-existent PgQue jobs by name, generating silently-swallowed exceptions in pg_cron logs on every `start_timetable()` call.
> **Suggestion:** Read `scheduler` before calling `stop_timetable()`, then skip `stop()` when the previous value was `'pg_timetable'`.

**LOW** `sql/pgque-additions/config.sql` — `config_scheduler_check` constraint unconditionally dropped and re-added on every install *(confidence: 5/10)*
> Inconsistent with the guarded `IF NOT EXISTS` pattern used for column and `tick_period_ms` additions. Creates a brief window where the constraint is absent mid-transaction.
> **Suggestion:** Wrap in an `IF NOT EXISTS` check matching the style used for the column addition directly above it.

**MEDIUM** `tests/test_pgtimetable_lifecycle.sql` — No test for upgrade path (existing install gaining the `scheduler` column) *(confidence: 7/10)*
> The `ALTER TABLE ADD COLUMN scheduler` + constraint drop/re-add block is untested against an existing populated config row.
> **Suggestion:** Add a test that simulates pre-upgrade state (config row without `scheduler` column) and verifies the upgrade lands correctly.

**MEDIUM** `tests/test_pgtimetable_lifecycle.sql` — `stop_timetable()` not tested when `scheduler='pg_cron'` *(confidence: 6/10)*
> When `scheduler='pg_cron'` and a timetable schema is present, `stop_timetable()` calls `delete_job()` for all four names (they return false — harmless) and skips the config UPDATE (`WHERE scheduler = 'pg_timetable'` doesn't match). Correct but unverified.
> **Suggestion:** Add a test that sets `scheduler='pg_cron'`, calls `stop_timetable()`, and asserts `scheduler` is still `'pg_cron'` afterwards.

**LOW** `sql/pgque-additions/lifecycle.sql` — `v_delete_job_reg` ownership verified but invocation uses string literals *(confidence: 5/10)*
> The trusted function OID is checked (`proowner <> v_timetable_owner`) but actual deletes use `execute $sql$select timetable.delete_job('pgque_ticker')$sql$` — resolving by name again rather than invoking the verified reference. The gap is small but the trust check and the invocation are mismatched.
> **Suggestion:** Document that the string-literal form is intentional (name-based dispatch is safe given the ownership check), or use the verified OID in the execute call.

**LOW** `.github/workflows/ci.yml` — Real integration test polling window (12s) may be tight on loaded runners *(confidence: 4/10)*
> No wait-for-worker-ready step before the assert script; if pg_timetable is slow to connect, the first tick may not fire within the budget.
> **Suggestion:** Add a brief log poll or health-check step confirming the pg_timetable worker has connected before running the assert.

**LOW** `sql/pgque-additions/lifecycle.sql` — `stop_timetable()` silently skips cleanup with no notice logged *(confidence: 6/10)*
> When a timetable schema exists with an untrusted owner and `scheduler != 'pg_timetable'`, cleanup is silently skipped with no operator feedback.
> **Suggestion:** Add `RAISE NOTICE 'skipping timetable cleanup: schema owner is untrusted'` so operators can diagnose unexpected scheduler state.

**LOW** `.github/workflows/ci.yml` — `PGPASSWORD` as plaintext env var *(confidence: 5/10)*
> Clearly a throwaway test credential scoped to an ephemeral Docker network (no real risk), but inconsistent with the pattern of using secrets.
> **Suggestion:** Add a comment: `# intentional throwaway test password — not a secret`.

**MEDIUM** `README.md:~150` — "Scheduler log hygiene" body is still pg_cron-only after the heading was generalized *(confidence: 7/10)*
> The purge recipe, `cron.log_run` tip, and `cron.job_run_details` references are all pg_cron-specific. The trailing pg_timetable note gives no actionable guidance.
> **Suggestion:** Add a concrete pg_timetable example (e.g. `DELETE FROM timetable.execution_log WHERE ...`) or keep separate pg_cron / pg_timetable subsections.

**MEDIUM** `README.md` — Cross-database pg_timetable usage not documented *(confidence: 7/10)*
> The "pg_cron in a different database" callout is present but the pg_timetable equivalent is missing — pg_timetable is actually simpler here (just point `--connstr` at the target DB directly).
> **Suggestion:** Add a short note: pg_timetable natively connects to the database specified by `--connstr`, so no cross-database workarounds are needed.

**LOW** `README.md` — pg_timetable version compatibility (12-arg vs 11-arg `add_job`) not surfaced to users *(confidence: 7/10)*
> The code handles both transparently, but users on old installs won't know if they're supported.
> **Suggestion:** Add "compatible with pg_timetable v4+" or note the minimum supported version.

**INFO** `tests/test_pgtimetable_lifecycle.sql:~10` — Fake-vs-real detection keyed on absence of `timetable.execution_log` *(confidence: 6/10)*
> If a future pg_timetable version renames or removes that table, the cleanup guard silently skips dropping the fake schema, leaving stale state.
> **Suggestion:** Insert a sentinel column or row at harness-install time and key teardown on that marker instead.

**INFO** `README.md:~150` — "Scheduler log hygiene" heading misrepresents pg_cron-only body *(confidence: 5/10)*
> Already covered above; worth tracking separately as a heading-level mismatch.

---

### Summary

| CI/Pipeline | 0 | 0 | 0 |
| Security | 0 | 2 | 0 |
| Bugs | 0 | 3 | 0 |
| Tests | 1 | 3 | 0 |
| Guidelines | 1 | 2 | 0 |
| Docs | 1 | 4 | 0 |

- **Findings**: High-confidence issues (8–10/10) — blocking or non-blocking per severity
- **Potential**: Medium-confidence issues (4–7/10) — review manually, may be false positives
- **Filtered**: Low-confidence issues (0–3/10) — excluded as likely false positives

---

NikolayS:
Both blocking issues from the previous REV review are resolved in the current tip:

1. **Test 6 (pg_timetable→pg_cron switch)** — already guarded in `ci.yml` with an explicit `FAIL` if the switch test is skipped despite pg_cron being installed (line 243). The test runs correctly in the `pg_cron install path` job.

2. **`--clientname=pgque` not explained** — already documented in README: *"The `--clientname=pgque` flag is required; pg_timetable only executes chains whose `job_client_name` matches the running worker's client name."*

The review was run against an earlier commit; both were fixed before CI passed. No remaining blockers. ✅

pashagolub:
We can use `@every 100 milliseconds` for sub-seconds intervals. Any postgres interval should work here

PR #292, opened 2026-06-10
fix: make transform.sh patches fail loudly

NikolayS:
# Build-pipeline integrity fixes (transform.sh)

Addresses findings C1, C2, C3, C5, C8 of #283.

### C1 — silent no-op sed patches (medium)
The critical correctness patches (batch_event_sql xid8 quoting, `batch_retry` `NULL::int8 → NULL::xid8`, `upgrade_schema` `return 0 → return cnt`, sqltriga header fix, annotation seds) were plain `sedi` calls: sed exits 0 whether or not it matched, and the script printed `PASS: ...` unconditionally. If upstream PgQ reworded an anchored line, the build would report success while the generated `batch_event_sql` emits unquoted txids → runtime `operator does not exist: xid8 = integer` on every batch fetch.

**Fix:** new `sedi_must` helper (cmp before/after; exits 1 with the failing expression and file when nothing changed). Every must-match `sedi` now goes through it. The `upgrade_schema` role-grant and `insert_event_raw` guard awk patches now count their substitutions/injections and fail on a count mismatch, matching the standard already set by the get_batch_cursor patch.

### C2 — missing source files only warned (medium)
- Transform loop: `WARNING ... continue` → hard error (`exit 1`).
- `FUNCTION_FILES` assembly loop: `WARNING` → hard error.
- `pgque-api` files: previously **silently skipped with no message** → hard error; a missing `sql/pgque-api/` directory is also a hard error.
- `pgque-additions` files: explicit hard error with a clear message.

A submodule bump that renames/drops a file can no longer ship a truncated install file that passes the CI drift check after regenerate-and-commit.

### C3 — vacuous pg_notify check + unverified ticker injection (medium)
- (a) The ticker pg_notify awk injection now counts both anchors (`return i_tick_id;`, `return currval(q.queue_tick_seq);`) and fails the build unless each fired exactly once. Same for the create_queue length-check injection.
- (b) The post-assembly check `grep -q 'pg_notify' "$INSTALL_FILE"` was vacuous — Section 6 (hardening.sql) contains its own pg_notify, so it passed even if BOTH ticker injections missed. Replaced with an awk that extracts each tick-emitting `pgque.ticker` definition from the generated file and requires `perform pg_notify(` inside each body.

### C5 — bogus annotation on hardening.sql's pg_notify line (low, changes generated files)
[...]

PR #291, opened 2026-06-10
fix(python): consumer loop drain, error survival, dead Event.extra

NikolayS:
## Bugs

**B2 (medium) — backlog drained at one batch per `poll_interval`.**
`Consumer.start()` unconditionally ran `_poll_once(conn)` then `_wait_for_notify_or_stop(conn)`. After a non-empty batch the consumer still waited up to `poll_interval` (default 30s) for a NOTIFY — but notifies fire only on new ticks, and the notifies for accumulated batches were emitted while the consumer was not listening. A consumer that was down with N batches accumulated drained them at ~1 batch per `poll_interval` and could never catch up.

**B3 (medium) — a transient SQL error killed the consumer.**
There was no try/except around the poll iteration, so a transient error from receive or ack (failover, restart, network blip) propagated out of `start()` and the consumer died, contradicting the documented "blocks until SIGTERM/SIGINT" contract. Nack failures were already handled, making this an oversight.

**B5 (low) — `Event.extra` was silently dropped.**
`pgque/types.py` declared `Event.extra: dict[str, str]`, but `client.send()` unpacks only `type` and `payload`; `extra` was referenced nowhere. Data placed there was silently lost.

## Fixes

- **B2:** `_poll_once` now returns whether it processed a batch; the loop re-polls immediately after a non-empty batch and only waits for NOTIFY/timeout when the queue came back empty (matches the Go consumer). A batch left unacked due to a nack failure reports `False`, so redelivery waits a poll cycle instead of hot-looping.
- **B3:** `start()` now wraps each connection session: on `psycopg.Error` / `PgqueError` it logs, sleeps `poll_interval` in short slices (so `stop()` stays prompt), reconnects, re-`LISTEN`s, and resumes — paralleling the Go consumer's log-and-retry on receive errors. `KeyboardInterrupt` (BaseException) and `stop()` semantics are unchanged; handler exceptions and nack failures remain contained inside `_poll_once`.
- **B5:** I checked the SQL surface first: the `pgque.send` overloads are `(queue, payload)` and `(queue, type, payload)` only — no `ev_extra1..4` parameters (those exist only on the PgQ primitive `pgque.insert_event(7)`), and a `dict[str, str]` has no natural mapping to four positional extra columns anyway. So there is no natural wiring through `pgque.send`; the dead field is removed. A never-working field that silently loses data is worse than a removed field — construction with `extra=` now fails loudly with `TypeError`.

[...]

PR #290, opened 2026-06-10
fix: harden uninstall paths (stop jobs, narrow guards)

NikolayS:
## Bugs

Four uninstall-path findings from #283:

- **C4 (medium):** `sql/pgque-tle-uninstall.sql` ran `drop extension if exists pgque cascade` without calling `pgque.stop()` first. pg_cron / pg_timetable jobs are catalog rows, not dependent objects, so `pgque_ticker` (every 1 s), `pgque_retry_events`, `pgque_maint`, and `pgque_rotate_step2` survived the drop and failed every 1–30 s forever ("schema pgque does not exist"), spamming logs and `cron.job_run_details`.
- **C6 (low):** `sql/pgque_uninstall.sql` wrapped `pgque.stop()` in `exception when others then null`, silently swallowing real failures (e.g. `cron.unschedule` permission errors, the "untrusted pg_timetable schema owner" exception) and dropping the schema anyway — same orphaned-jobs outcome, but silent.
- **C7 (low):** the generated `sql/pgque-tle.sql` header claimed the file "works in psql, GUI tools (DBeaver, etc.), JDBC, libpq-direct callers", but the script uses `\set ON_ERROR_STOP` and `\echo` psql meta-commands, so any non-psql client fails on the first one.
- **C9 (low):** `pgque.uninstall()` ran `drop schema pgque cascade`, which fails on extension (pg_tle) installs with a confusing dependency error ("extension pgque requires it").

## Fixes

- **C4:** `sql/pgque-tle-uninstall.sql` now calls `pgque.stop()` before `drop extension`, with the drop in the same `do` block, so a real `stop()` failure aborts the uninstall before anything is dropped.
- **C6:** both uninstall scripts narrow the handler to `undefined_function` / `invalid_schema_name` — empirically the sqlstates raised when pgque is not installed (42883 / 3F000) — keeping the scripts idempotent without swallowing real `stop()` failures. `sql/pgque_uninstall.sql` also performs the drop inside the same `do` block, so abort-before-drop holds regardless of client `ON_ERROR_STOP` settings.
- **C7:** the header (generated by `build/transform.sh`) now states the file is a psql script and points non-psql callers at `pgtle.install_extension()` with the `sql/pgque.sql` body. `ON_ERROR_STOP` behavior is kept.
- **C9:** `pgque.uninstall()` detects `pg_extension` membership first and raises a clear exception pointing to `drop extension pgque cascade` / `sql/pgque-tle-uninstall.sql`.

[...]

PR #289, opened 2026-06-10
fix(ts): back off on nack/ack failure in consumer

NikolayS:
## Bug

`Consumer.start()` in `clients/typescript/src/consumer.ts` slept `pollIntervalMs` only on receive error or empty batch. When a batch **was** received but finalization failed — a nack failed so ack was intentionally skipped, or `ack()` threw — the while loop re-polled immediately. Because the batch was never finished, `pgque.next_batch` returns the **same batch instantly**, so:

- a persistent nack/ack failure (e.g. partial grants where `receive` works but `nack`/`ack` don't) becomes a tight loop at full speed: re-receive → re-run ALL handlers (duplicate side effects) → fail → repeat, hammering both the application and Postgres;
- even a single transient ack failure re-executes the whole batch's handlers with zero delay.

The comment on the skip-ack path even claimed redelivery happens "on the next poll" — a poll interval that did not exist on that path.

## Fix

`await sleep(this.pollIntervalMs, signal)` before re-polling on both the `anyNackFailed` path and the ack-throw path. The existing `sleep` helper is abort-aware, so shutdown latency is unchanged. `ack` returning 0 without throwing stays warning-only with no sleep (the batch is finished in that case; the loop should keep draining).

Red/green TDD: two new mock-based tests (`pollInterval: 60_000`, `receive` always returns the same batch, nack/ack fail persistently) assert `receive` is called exactly once within a 300 ms observation window. Unfixed, the hot loop is so tight that the vitest worker dies of OOM recording mock calls — that was the red run. Green after the fix.

Second commit (informational note from the same review): `src/producer_bench.ts` spliced the table name from `pgque.current_event_table()` into `select count(*) from ${table}` via a template literal — not exploitable (self-generated value, dev-only script) but the only non-parameterized query in the repo. Now quoted part by part with pg's `escapeIdentifier`. Also fixed the result generic: the pgque pool parses int8 (OID 20) to `BigInt`, so `count(*)` arrives as `bigint`, not `string`; the old `<{ count: string }>` + `Number.parseInt` only worked by coercion.

## Verification

All run in `clients/typescript/` (deps via `bun install`):

```
bun run check
# tsc --noEmit -p tsconfig.json && tsc --noEmit -p tsconfig.test.json — clean

npx vitest run    # without PGQUE_TEST_DSN
# Test Files  7 passed | 1 skipped (8)
#      Tests  35 passed | 50 skipped (85)

[...]

PR #301, opened 2026-07-09
release: stamp v0.3.0-alpha.1

NikolayS:
## Summary

Stamps the **v0.3.0-alpha.1** prerelease. Bumps the devel `pgque.version()`
literal from `0.3.0-devel` to `0.3.0-alpha.1`, regenerates the single-file
installs from source via `build/transform.sh`, and updates the version
assertion in `test_pgque_config.sql`.

Also adds two terse alpha-tester notes to `devel/sql/README.md`:
- ticker idle-period latency (default `queue_ticker_idle_period` 60s — on a
  quiet queue an event can take up to that long to become receivable);
- a pointer to `blueprints/partition-keys/SPEC.md` for partition-keys testers
  until `docs/` coverage lands.

Scope note: this stamps the **devel** install only. Stable `sql/` stays at
0.2.0. A follow-up PR will restore the devel literal to `0.3.0-devel` after
this merges and the tag/release are cut.

## Changes

- `devel/sql/pgque-additions/lifecycle.sql` — `pgque.version()` → `0.3.0-alpha.1`
- `devel/sql/pgque.sql`, `devel/sql/pgque-tle.sql` — regenerated (version-only diff)
- `tests/test_pgque_config.sql` — version assertion → `0.3.0-alpha.1`
- `devel/sql/README.md` — two alpha-tester notes

## Verification

Built and tested against Postgres 17 in Docker, mirroring CI:

```bash
bash build/transform.sh          # clean regen, version 0.3.0-alpha.1, all checks pass
# install
psql ... -v ON_ERROR_STOP=1 -f devel/sql/pgque.sql
# regression suite
psql ... -v ON_ERROR_STOP=1 -f tests/run_all.sql             # 190 PASS, exit 0
# experimental (separate db)
psql ... -f devel/sql/experimental/{delayed,observability,config_api}.sql
psql ... -v ON_ERROR_STOP=1 -f tests/run_experimental.sql    # 18 PASS, exit 0
# acceptance
psql ... -v ON_ERROR_STOP=1 -f tests/acceptance/run_acceptance.sql  # 54 PASS, exit 0
# idempotency re-install
psql ... -f devel/sql/pgque.sql && psql ... -f tests/test_install_idempotency.sql  # PASS
# runtime version
psql ... -tAc "select pgque.version()"   # 0.3.0-alpha.1
```

Generated-file diff is version-only churn (header `Version:` line, `version()`
return, and the pg_tle registration strings); no incidental regeneration noise.

## Release notes draft

Posted as a PR comment for review. After merge: annotated tag `v0.3.0-alpha.1`
on the merge commit, GitHub **prerelease** with those notes, then the
devel-restore follow-up PR.

### DIFF
--- a/devel/sql/README.md
+++ b/devel/sql/README.md
@@ -7,6 +7,10 @@ churn and validate against a throwaway database, not production.
 The released, stable install lives in [`../../sql/`](../../sql/) — use that for
 anything real.
 
+Testing partition keys (ordered per-key processing via slot consumers)? Until
+`docs/` coverage lands, follow
+[`blueprints/partition-keys/SPEC.md`](../../blueprints/partition-keys/SPEC.md).
+
 Layout:
 
 - `pgque.sql`, `pgque-tle.sql` — generated single-file installs (built from the
@@ -40,6 +44,10 @@ To uninstall: `\i devel/sql/pgque_uninstall.sql`.
 
 PgQue does not deliver messages without a ticker: enqueueing works, but
 consumers see nothing until ticks are created.
+
+On a quiet queue, the ticker falls back to `queue_ticker_idle_period`
+(default 60s), so a newly enqueued event can take up to that long to become
+receivable. Run the ticker at your desired cadence to bound this latency.
 
 With `pg_cron` in the same database, one call sets up the ticker and
 maintenance jobs (10 ticks/sec by default):
--- a/devel/sql/pgque-additions/lifecycle.sql
+++ b/devel/sql/pgque-additions/lifecycle.sql
@@ -448,7 +448,7 @@ returns text as $$
 begin
     /* In-development build: the '-devel' suffix marks this as not a stable
        release. Bump to the release version (drop '-devel') at release time. */
-    return '0.3.0-devel';
+    return '0.3.0-alpha.1';
 end;
 $$ language plpgsql security definer set search_path = pgque, pg_catalog;
 
--- a/devel/sql/pgque-tle.sql
+++ b/devel/sql/pgque-tle.sql
@@ -70,8 +70,8 @@ begin
     from pgtle.available_extensions()
     where name = 'pgque';
 
-    if existing_version = '0.3.0-devel' then
-        raise notice 'pgque 0.3.0-devel already registered with pg_tle; skipping install_extension().';
+    if existing_version = '0.3.0-alpha.1' then
+        raise notice 'pgque 0.3.0-alpha.1 already registered with pg_tle; skipping install_extension().';
         return;
     end if;
 
@@ -80,16 +80,16 @@ begin
             'but this script registers version %. Run '
             'devel/sql/pgque-tle-uninstall.sql first to remove the existing '
             'registration, then re-run this script.',
-            existing_version, '0.3.0-devel';
+            existing_version, '0.3.0-alpha.1';
     end if;
 
     perform pgtle.install_extension(
         'pgque',
-        '0.3.0-devel',
+        '0.3.0-alpha.1',
         'PgQue — PgQ Universal Edition (zero-bloat Postgres queue)',
 $pgque_extension_body$
 -- pgque.sql -- PgQ Universal Edition
--- Version: 0.3.0-devel
+-- Version: 0.3.0-alpha.1
 -- Copyright 2026 Nikolay Samokhvalov. Apache-2.0 license.
 -- Includes code derived from PgQ (ISC license, Marko Kreen / Skype Technologies OU).
 --
@@ -4670,7 +4670,7 @@ returns text as $$
 begin
     /* In-development build: the '-devel' suffix marks this as not a stable
        release. Bump to the release version (drop '-devel') at release time. */
-    return '0.3.0-devel';
+    return '0.3.0-alpha.1';
 end;
 $$ language plpgsql security definer set search_path = pgque, pg_catalog;
 
@@ -8121,5 +8121,5 @@ $pgque_extension_body$
 end $wrapper$;
 
 \echo ''
-\echo 'PgQue 0.3.0-devel registered with pg_tle.'
+\echo 'PgQue 0.3.0-alpha.1 registered with pg_tle.'
 \echo 'Run create extension pgque; to materialise the schema in this database.'
--- a/devel/sql/pgque.sql
+++ b/devel/sql/pgque.sql
@@ -1,5 +1,5 @@
 -- pgque.sql -- PgQ Universal Edition
--- Version: 0.3.0-devel
+-- Version: 0.3.0-alpha.1
 -- Copyright 2026 Nikolay Samokhvalov. Apache-2.0 license.
 -- Includes code derived from PgQ (ISC license, Marko Kreen / Skype Technologies OU).
 --
@@ -4580,7 +4580,7 @@ returns text as $$
 begin
     /* In-development build: the '-devel' suffix marks this as not a stable
        release. Bump to the release version (drop '-devel') at release time. */
-    return '0.3.0-devel';
+    return '0.3.0-alpha.1';
 end;
 $$ language plpgsql security definer set search_path = pgque, pg_catalog;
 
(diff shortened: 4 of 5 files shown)


NikolayS:
Release notes draft for review (will become the GitHub prerelease body):

---

# PgQue v0.3.0-alpha.1

**Prerelease — for testing only.** This alpha ships the new partition-keys and
producer-idempotency work for early feedback. Install it from
`devel/sql/pgque.sql` at this tag. The stable install under `sql/` remains
**0.2.0** and is unchanged; use that for anything real.

This is an SQL-only alpha. The client libraries (gem / PyPI / npm / Go module)
are **not** being versioned or published in this stamp — client releases will
follow once the partition-keys surface stabilizes.

## Highlights

### Partition keys — ordered per-key processing (#295)

Consume a queue with strict per-key ordering while still processing different
keys in parallel. Events are distributed across a fixed set of *slots*; each
slot is consumed by one worker at a time, and ordering is preserved within a
slot. Coordination uses **batch-granularity leases** rather than advisory
locks, so it is safe behind a transaction pooler (e.g. PgBouncer in
transaction mode) and needs no session-level state.

There are no client-library helpers yet — drive it with the raw SQL described
in `blueprints/partition-keys/SPEC.md` (see §15). User-facing docs under
`docs/` are still pending; the SPEC is the reference for now.

### Producer idempotency — `pgque.send_idem()` (#295)

Deduplicate on the producer side: `pgque.send_idem()` claims an idempotency
key and appends the event atomically, so a retried producer call does not
enqueue a duplicate event.

### New: Ruby client (#234)

A `pgque` Ruby gem joins the client lineup. (Not published to RubyGems in this
alpha — see the SQL-only note above.)

## Fixes rolled in

- **`get_queue_info` is reader-callable** — consumers can read queue metadata
  without writer/admin grants (#267).
- **`dlq_replay` concurrent double-replay guarded** — two workers can no longer
  replay the same DLQ event twice (#287).
- **Uninstall hardening** — scheduler jobs are torn down, and the uninstall
  scripts abort loudly instead of silently swallowing errors (#290).
- **`transform.sh` build guards** — the build fails loud on integrity
  regressions instead of emitting a subtly wrong install file (#292).
- **Go client** — NULL type/payload columns scan correctly, and ack failures
  back off instead of hot-looping (#285, #284).
- **TypeScript client** — ack/nack now back off on failure; benchmark harness
  hardened (#289).
- **Python client** — consumer drains cleanly and survives per-event errors
  without aborting the loop (#291).
- **Experimental SQL** — throughput division-by-zero and negative
  `max_retries` fixed, with CI coverage added (#288).

## Continuous integration

PostgreSQL 19 beta 1 is now part of the continuously tested server matrix
(PG 14–19beta1).

## Breaking change (Python client)

`Event.extra` has been **removed**. It never functioned — data passed via
`extra` was silently dropped. Passing `extra=` now raises `TypeError` at the
call site instead of failing silently. Update callers to stop passing it.

## Known limitations — partition-keys alpha

Partition keys are alpha. Read `blueprints/partition-keys/SPEC.md` before
relying on any of this, and watch `pgque.partition_slot_status`.

- **No sticky-worker takeover (M=N).** When a sticky worker stops, its freed
  slot is not automatically picked up by another worker.
- **Stale subscriptions pin rotation silently.** A subscribed-but-idle slot
  holds back tick rotation with no error. Monitor
  `pgque.partition_slot_status`.
- **`i_max` caps a slot's batch slice.** On the partitioned path, events beyond
  `i_max` in a batch are skipped for that slot and picked up on the next batch.
- **Every subscribed slot must be polled.** If any subscribed slot is never
  consumed, rotation pins. Poll all subscribed slots.
- **Partial `unsubscribe_slot` orphans its key-class.** Unsubscribing only some
  slots of a key-class can leave that class orphaned.
- **Non-superuser-installer variant is not yet security-tested.**
- **No client-library helpers yet.** Use the raw SQL in the SPEC (§15).
- **User-facing docs pending.** `docs/` coverage is not in this alpha.

## Install

```sql
begin;
\i devel/sql/pgque.sql
commit;
```

Requires PostgreSQL 14+ and a periodic ticker. See `devel/sql/README.md` for
ticker setup and role grants.

NikolayS:
## Release review — v0.3.0-alpha.1 stamp

Independent review pass over the head commit (`6574783`), performed in a fresh clone with the build reproduced and the install exercised against Postgres 17 in Docker.

### 1. Scope discipline — PASS

Diff against the true base (current `main` tip `500882f`, which this branch is based on) touches exactly the five expected files. In the two generated files (`devel/sql/pgque.sql`, `devel/sql/pgque-tle.sql`), **every changed line contains the version string** — no incidental regeneration noise (verified mechanically by filtering the diff). The two `devel/sql/README.md` additions are the ticker idle-period note and the partition-keys SPEC pointer, nothing else.

### 2. Build reproduction — PASS

Fresh clone at the PR head, `git submodule update --init`, `bash build/transform.sh`:
all assembly checks pass, output reports `version 0.3.0-alpha.1`, and `git status` is **clean** afterward — the committed generated files match the build byte-exact.

### 3. Empirical checks (Postgres 17.10, Docker) — PASS

- `\i devel/sql/pgque.sql` on a fresh database, then `select pgque.version()` → `0.3.0-alpha.1`.
- `tests/run_all.sql` → 190 PASS, `=== ALL TESTS PASSED ===`, exit 0.
- `tests/acceptance/run_acceptance.sql` → 54 PASS, exit 0.
- Experimental (`delayed`, `observability`, `config_api` in a separate DB) + `tests/run_experimental.sql` → 18 PASS, exit 0.
- Re-install over an existing install + `tests/test_install_idempotency.sql` → PASS.
- `pgque-tle.sql` string check: `0.3.0-alpha.1` appears in all seven expected places (skip-guard, mismatch error, `install_extension` arg, embedded header, embedded `version()`, final `\echo`); zero `0.3.0-devel` leftovers.
- The release-notes install snippet was executed **verbatim** (`begin; \i devel/sql/pgque.sql commit;`) on a fresh DB — works, returns `0.3.0-alpha.1`.

### 4. Release-notes draft fact-check — PASS

- All 11 cited PRs (#295, #234, #267, #287, #290, #292, #285, #284, #289, #291, #288) are **MERGED** and contained in this branch's base.
- Breaking-change claim verified in code: `Event.extra` field removed in #291; `clients/python/tests/test_send.py` asserts `TypeError` on `Event(..., extra=...)`. The `extra1..extra4` fields (PgQ `ev_extra1..4`) are distinct and untouched — the wording correctly scopes the removal to `extra`.
- Known-limitations list is grounded in `blueprints/partition-keys/SPEC.md` (sticky/unclaimed-slot risk, non-superuser-owner security test still pending per T-security, §15 as the raw-SQL reference). Claims are conservative, no overclaiming; no benchmark numbers in the notes.
- Stable `sql/pgque.sql` confirmed at `0.2.0` (header + `version()` literal) and untouched by this PR; the "stable stays 0.2.0" claim is true.
- No private links, work-item references, or prospect names in the draft.
- Semver sanity: `0.3.0-alpha.1 < 0.3.0` per SemVer prerelease precedence — a later `0.3.0` tag will order correctly.
- README note accuracy: `queue_ticker_idle_period` default is `'1 minute'` in the source, matching the "default 60s" wording.

### 5. Version-string mechanics — PASS

- `grep -r '0.3.0-devel'` over the whole tree (excluding the `pgq/` submodule) → zero hits; nothing stale.
- `transform.sh` accepts the prerelease literal (build above proves it).
- `tests/test_pgque_config.sql` assertion updated to the exact new string.

### 6. CI — PASS

All 16 check runs green on head `6574783`, including the frozen-`sql/` smoke, both upgrade paths, pg_tle/pg_cron/pg_timetable paths, and all four client suites.

### Nits (non-blocking)

- The comment above `pgque.version()` in `devel/sql/pgque-additions/lifecycle.sql` still reads "In-development build: the '-devel' suffix marks this as not a stable release" while the literal is now `0.3.0-alpha.1`. Harmless and self-resolving when the follow-up restores `0.3.0-devel`; no action needed for this stamp.

### Verdict

**APPROVED — release review passed.** Safe to merge, tag `v0.3.0-alpha.1` on the merge commit, and publish the prerelease with the drafted notes. (Formal GitHub approval is blocked because the reviewing session is authenticated as the PR author; this comment records the approval.)

PR #311, opened 2026-07-10
fix(partition-keys): fence lease release

NikolayS:
## What changed

- refuse `pgque.release_slot()` when the owning worker still has an open slot batch;
- retain the lease so another worker cannot claim through an invalid mid-batch handoff;
- add regression coverage directly in the zombie/takeover scenario;
- regenerate both development install artifacts.

## Why

The lease is the process-to-ack fence. Previously an owner could release it after `receive_partitioned()` but before `ack_partitioned()`. A successor could then claim the slot and receive the same open batch while the original worker was still processing it, violating the single-processor guarantee.

Crash recovery remains unchanged: a genuinely abandoned batch is recovered after lease expiry, with the existing epoch fencing.

Fixes #304.

## Validation

Red/green:

- the new regression failed on current `main` with `fencing: owner release with an open batch must raise`;
- it passes after the fix and verifies the attempted release leaves the original lease intact.

Checks run:

- `bash build/transform.sh`
- `git diff --check`
- PostgreSQL 14: install + `tests/test_partition_keys.sql`
- PostgreSQL 18: install + `tests/test_partition_keys.sql` + full `tests/acceptance/run_acceptance.sql`
- PostgreSQL 19 beta 1: install + `tests/test_partition_keys.sql`

This is intentionally a draft and has not been merged.

### DIFF
--- a/devel/sql/pgque-api/partition_keys.sql
+++ b/devel/sql/pgque-api/partition_keys.sql
@@ -446,11 +446,16 @@ create or replace function pgque.release_slot(
 returns boolean as $$
 declare
     v_queue_id int4;
+    v_n int4;
     v_owner text;
+    v_batch_id bigint;
 begin
-    select ps.queue_id, ps.lease_owner into v_queue_id, v_owner
+    select ps.queue_id, ps.lease_owner, pc.n into v_queue_id, v_owner, v_n
     from pgque.partition_slot as ps
     join pgque.queue as q on q.queue_id = ps.queue_id
+    join pgque.partition_consumer as pc
+      on pc.queue_id = ps.queue_id
+     and pc.co_name = ps.co_name
     where q.queue_name = i_queue
       and ps.co_name = i_consumer
       and ps.slot = i_slot
@@ -462,6 +467,15 @@ begin
 
     if v_owner is null or v_owner <> i_worker then
         return false;
+    end if;
+
+    /* The lease is the process-to-ack fence. Clearing it with an open batch
+       would let a successor receive that same batch while this worker may
+       still be processing it. Crash recovery must go through lease expiry. */
+    v_batch_id := pgque._slot_batch(i_queue, i_consumer, i_slot, v_n);
+    if v_batch_id is not null then
+        raise exception 'cannot release slot % of consumer % on queue % while batch % is open; ack the batch first',
+            i_slot, i_consumer, i_queue, v_batch_id;
     end if;
 
     update pgque.partition_slot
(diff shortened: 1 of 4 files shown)


NikolayS (commented):
REV rubric review

This is a manual execution of the repository's five applicable REV rubrics because the automated ultrareview quota was unavailable. It is not bot output. SOC2 was omitted.

- Bug/security: the open-batch release fence preserves lease ownership and prevents concurrent handoff; no additional defect or security issue found.
- Tests: focused failure and lease-retention assertions cover the hazard; CI is 17/17 green. The initial SQL/test commit makes the original RED retrospective.
- Docs/guidelines: invariant comments are concise and generated artifacts match. The historical `test:` commit conflicts with the repo-specific CLAUDE type list, although generic rules allow it.

Blocking patch findings: none. Remaining constraints: preserve history, use a compliant squash/merge subject, and post exact real-user database evidence before merge. This COMMENT review is not an approval.

NikolayS:
Real-user verification evidence for head `73dc1d68c8be7f394fdcbe9e721996226d59abc6`.

I tested this from fresh disposable databases. `PGQUE_TEST_DSN` pointed at the fresh database in each PostgreSQL container. The focused replay was:

```bash
PAGER=cat psql --no-psqlrc "$PGQUE_TEST_DSN" --set=ON_ERROR_STOP=1 \
  --file=devel/sql/pgque.sql
PAGER=cat psql --no-psqlrc "$PGQUE_TEST_DSN" --set=ON_ERROR_STOP=1 \
  --file=tests/test_partition_keys.sql
bash build/transform.sh
git diff --check
```

The PR regression replayed against the pre-fix SQL failed at the intended assertion:

```text
ERROR: fencing: owner release with an open batch must raise
```

The same test against the PR head passed and continued through the full partition suite:

```text
PASS fencing: zombie ack raises after takeover; heir re-issued same batch
PASS guard: receive_partitioned renews the lease (heartbeat)
PASS: partition keys (Phase 1A)
```

This verifies that a mid-batch `release_slot()` is rejected, the original lease remains in place, a stale owner cannot ack after takeover, and the successor receives the same open batch. Fresh-install coverage was repeated on PostgreSQL 14, 18, and 19beta1; PostgreSQL 18 also ran the full acceptance suite.

RED/GREEN chronology: retrospective. The implementation was already present before the regression-test commit. I independently replayed that regression against the parent/pre-fix SQL to establish RED, then against this head to establish GREEN; this is not a chronological test-first commit sequence.

Full CI: [17/17 checks passed](https://github.com/NikolayS/PgQue/actions/runs/29155163781).

This comment records test evidence only; it is not a review or merge decision.

PR #322, opened 2026-07-10
fix(upgrade): preserve legacy hash consumers

NikolayS:
## What changed

- classify a consumer as a partition slot only when matching partition metadata exists, rather than from `#` in its name;
- preserve plain receive/ack/nack for ordinary consumers created before `#` became reserved;
- prevent `subscribe_slot` from silently reusing an exact generated name that belongs to a legacy ordinary consumer;
- retain rejection of new plain `#` registrations and idempotent re-subscribe for genuine slots;
- add a real v0.2.0→HEAD upgrade fixture/assertion set and CI lane;
- regenerate development install artifacts.

## Why

PgQue 0.2.0 allowed ordinary consumer names containing `#`. After upgrade, 0.3's string-only guard stranded those subscriptions even though their queue position and pending events survived. Exact slot-shaped names could also be silently reclassified.

Fixes #306.

## Validation

Red evidence:

```text
ERROR: consumer team#1 is a partition slot consumer;
       use pgque.receive_partitioned()
```

The collision regression also reproduced `subscribe_slot` silently reusing `workers#0/2` before the fix.

Green checks:

- real v0.2.0 install → pending legacy event → transactional HEAD upgrade;
- legacy `team#1` receive, ack, nack, and DLQ behavior as bare `pgque_reader`;
- exact generated-name collision is rejected with ordinary subscription and rollback-clean partition metadata preserved;
- genuine slot re-subscribe remains idempotent;
- exact new CI sequence including grants, roles, security, reinstall, and idempotency checks;
- existing v0.1.0→HEAD upgrade fixture;
- complete `tests/run_all.sql`;
- `bash build/transform.sh`, YAML parse, shell syntax, and `git diff --check`.

This is intentionally a draft and has not been merged.

PR #321, opened 2026-07-10
fix(partition-keys): make claims nonblocking

NikolayS:
## Stack

This PR is intentionally stacked on #311 because both changes touch the partition lease implementation. Review only the commit above `agent/fix-partition-lease-boundary`.

## What changed

- use a non-blocking row lock in `claim_slot`;
- return `NULL` when another transaction currently owns the slot row, allowing a first-free scan to continue;
- keep the distinct “slot is not subscribed” error when no row exists;
- reject non-finite lease TTLs;
- extend the real two-session lease harness with an uncommitted-row-lock contender;
- run that harness in every PostgreSQL matrix lane;
- add PostgreSQL 19 infinite-lease coverage and regenerate development artifacts.

## Why

The API contract says a busy claim returns `NULL`; it previously blocked on `SELECT ... FOR UPDATE`. A slow claim/renew/receive transaction could therefore head-of-line-block a worker scanning for any free slot.

Fixes #309.

## Validation

Red evidence on the parent/current behavior:

```text
ERROR: canceling statement due to statement timeout
CONTEXT: while locking tuple ... in relation "partition_slot"
```

Green checks:

- real two-session harness: contender returns `NULL` behind an uncommitted lease row, existing lease-expiry/epoch recovery still passes;
- PostgreSQL 14: complete partition regression;
- PostgreSQL 18: complete partition regression + two-session harness;
- PostgreSQL 19 beta 1: complete partition regression including infinite-TTL rejection;
- `bash build/transform.sh`;
- workflow YAML parse, shell syntax, and `git diff --check`.

This is intentionally a draft and has not been merged.

PR #336, opened 2026-07-10
fix(partition-keys): make slot setup atomic

NikolayS:
## Summary

- add `subscribe_partitioned(queue, consumer, n)` as the atomic whole-consumer setup path
- create every slot subscription at one shared tick in one transaction
- reject pre-existing partial setup instead of implying lost history can be recovered
- expose `partition_slot_status.subscribed` and report unknown lag as `NULL`
- retain and document `subscribe_slot` only as an explicit repair path — now explicitly forward-only (the missed window is permanently lost) in the error hint, comments, SPEC, and README

Review follow-ups folded in (from the design+empirical review):

- **n-cap upgrade guard (finding 1):** `pgque._partition_n_cap_guard()` runs before the idempotent constraint re-add; a pre-cap consumer with `n > 256` now aborts the install with an actionable error naming the queue/consumer/n and the recreate path, instead of a bare check violation
- **whole-consumer teardown (finding 3):** `pgque.unsubscribe_partitioned(queue, consumer)` — one transaction drops every slot subscription and the pinned-N row; succeeds on partial setups (the escape hatch the incomplete-setup error points to); absent consumer is a notice-level no-op; granted to `pgque_reader`. `unsubscribe_slot` on a complete consumer now emits a WARNING that it creates incomplete setup
- **canonical alert (finding 2):** documented as `pending_events > X or not subscribed` in the view header, SPEC (R7 / §15 / US-12.6), and `devel/sql/README.md` — a threshold-only alert skips the `NULL`-lag rows and never sees incomplete setup
- **repair ≠ recovery wording (finding 4):** see `subscribe_slot` bullet above
- **SPEC contract sync for the base PRs:** `release_slot` raises on an open owner batch (ack first); `claim_slot` NULL means "not claimable right now" (busy or foreign-owned — including the same worker's in-flight receive/ack renewal; never infer lease loss from NULL); claim TTL must be finite (`'infinity'` rejected)
- **test gate fix:** the infinite-TTL guard test now runs on PG 17+ (`'infinity'::interval` is PG 17+ syntax, not PG 19+), so the 17/18 CI lanes exercise it
- **catalog-driven slot classification (from the PR #322 adversarial review):** a legacy ordinary consumer literally named like a slot (`workers#0/2`, creatable on older installs) is never classified as a partition slot by name shape. `partition_slot_status` now reports `subscribed` only when both the `partition_slot` row and the engine subscription exist (no more contradictory view-vs-API signals); `unsubscribe_slot`'s completeness warning and last-slot cleanup use the same rule, so a legacy subscription cannot keep the pinned-N row alive; `unsubscribe_partitioned` drops only catalog-registered slots and leaves a legacy same-named consumer (and its cursor history) intact. Pinned-n hints now point at `unsubscribe_partitioned`; the slot-name collision error names plain `pgque.unsubscribe()` as the legacy-removal path. (The `subscribe_slot`-side collision error lives on the #322 branch, not here.)

Fixes #307.

## Stack

- base: #321 (`agent/fix-claim-slot-nonblocking`)
- #321 is stacked on #311

## Validation

- focused atomic setup regression: PostgreSQL 14, 17, and 19beta1
- old partial-state upgrade/status check passed
- pre-cap (`n=300`) re-install simulation: guard raises the actionable error; `--single-transaction` re-install rolls back cleanly (PG 18.3)
- teardown regression (complete / partial / absent / re-setup with a new n) in `tests/test_partition_setup.sql` passed
- partial-unsubscribe WARNING emission + teardown-of-partial asserted in `tests/two_session_slot_claim.sh`
- infinite-TTL guard exercised on PG 18.3 under the lowered gate
- existing partition-key suite passed
- US-12 acceptance passed
- full regression suite passed (`tests/run_all.sql`)
- full acceptance suite passed (`tests/acceptance/run_acceptance.sql`)
- stacked two-session nonblocking/TTL/fencing harness passed
- deterministic generator/assembly (`build/transform.sh`) and `git diff --check` passed
- legacy name-shaped consumer regression (view classification, teardown survival, last-slot cleanup, plain-unsubscribe unblock) in `tests/test_partition_setup.sql` passed
- idempotent re-install of the regenerated `devel/sql/pgque.sql` on a populated database passed

### DIFF
--- a/blueprints/partition-keys/SPEC.md
+++ b/blueprints/partition-keys/SPEC.md
@@ -260,11 +260,17 @@ optimization is future (R6).
   `subscription` (`partition_block.sub_id → subscription.sub_consumer =
   dead_letter.dl_consumer_id`) — `sub_id` and `co_id` are different ID spaces
   (`dlq.sql:24,75-85`, `pgque.sql:170-183`); do not compare them directly.
-- **Teardown:** `unsubscribe_slot` removes the slot subscription (the
-  `partition_slot` and `partition_block` FKs cascade). Note `unregister_consumer`
-  cascades `dead_letter` (`dlq.sql:24`), so dropping a slot drops its DLQ audit —
+- **Teardown:** `unsubscribe_partitioned(queue, consumer)` is the whole-consumer
+  inverse of `subscribe_partitioned`: one transaction drops every slot
+  subscription and the pinned-N row (lease rows cascade via FK). It succeeds on
+  partial setups — it is the recreate path the incomplete-setup error names —
+  and an absent consumer is a notice-level no-op. `unsubscribe_slot` removes one
+  slot (the `partition_slot` and `partition_block` FKs cascade); on a complete
+  consumer it emits a WARNING, because it creates exactly the incomplete-setup
+  state `subscribe_partitioned` rejects. Note `unregister_consumer` cascades
+  `dead_letter` (`dlq.sql:24`), so dropping a slot drops its DLQ audit —
   documented.
-- **Grants:** producer → `pgque_writer`; `subscribe_partitioned`/`subscribe_slot`/`unsubscribe_slot`/
+- **Grants:** producer → `pgque_writer`; `subscribe_partitioned`/`subscribe_slot`/`unsubscribe_slot`/`unsubscribe_partitioned`/
   `receive_partitioned`/`ack_partitioned`/`nack_partitioned`/`claim_slot`/
   `release_slot` and `select` on `partition_slot_status` → `pgque_reader`;
   `partition_consumer`/`partition_slot`/`partition_block` revoked from all app
--- a/blueprints/partition-keys/SPEC.md
+++ b/blueprints/partition-keys/SPEC.md
@@ -147,7 +147,7 @@ to slot 0 — G1, so no event is dropped), assembled only from the validated int
 **SECURITY DEFINER ownership (round 3 — corrected).** `get_batch_cursor`'s
 `extra_where` is a trusted-SQL sink, revoked from `public/pgque_reader/
 pgque_writer`, admin-only (`pgque.sql:2221`, `:4852`). `receive_partitioned` and
-`subscribe_slot` reach it **because they are owned by the same role that owns
+`subscribe_partitioned`/`subscribe_slot` reach it **because they are owned by the same role that owns
 `get_batch_cursor` (the install owner) — a function owner may execute its own
 functions regardless of grants.** This is *not* the `receive`/`nack` pattern
 (those never call `get_batch_cursor`; they call reader-granted internals), and it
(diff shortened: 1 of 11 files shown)


NikolayS (commented):
REV rubric review

This is a manual execution of the repository's five applicable REV rubrics because the automated ultrareview quota was unavailable. It is not bot output. SOC2 was omitted.

- Bug/security: the previously reproduced unbounded reader-role work is resolved. `n` is now constrained to 1–256 in table, setup, and repair paths, with current-state capacity guidance.
- Tests: `d05fbac` adds the boundary contract before `6db7b5b` implements it; rollback, 256 acceptance, 257/large rejection, and generated artifacts are covered. CI is 17/17 green.
- Docs/guidelines: shell synchronization and current base are clean. Historical `style:` is repo-list-inconsistent and 52 characters.

Blocking patch findings: none. Merge-order constraint: this PR is stacked on #321, which is stacked on #311; merge in that order and rerun if a parent changes. Exact database evidence remains required. This COMMENT review is not an approval.

NikolayS:
Real-user verification evidence for head `6db7b5b0495398417b7623455142198a6e3f0f1f`, tested with the full #311 -> #321 stack.

Fresh-install functional commands:

```bash
PAGER=cat psql --no-psqlrc "$PGQUE_TEST_DSN" --set=ON_ERROR_STOP=1 \
  --file=devel/sql/pgque.sql
PAGER=cat psql --no-psqlrc "$PGQUE_TEST_DSN" --set=ON_ERROR_STOP=1 \
  --file=tests/test_partition_setup.sql
PAGER=cat psql --no-psqlrc "$PGQUE_TEST_DSN" --set=ON_ERROR_STOP=1 \
  --file=tests/test_partition_slot_limit.sql
PAGER=cat psql --no-psqlrc "$PGQUE_TEST_DSN" --set=ON_ERROR_STOP=1 \
  --file=tests/test_partition_keys.sql
PAGER=cat psql --no-psqlrc "$PGQUE_TEST_DSN" --set=ON_ERROR_STOP=1 \
  --file=tests/run_all.sql
PAGER=cat psql --no-psqlrc "$PGQUE_TEST_DSN" --set=ON_ERROR_STOP=1 \
  --file=tests/acceptance/run_acceptance.sql
PGQUE_TEST_DSN="$PGQUE_TEST_DSN" tests/two_session_slot_claim.sh
```

The original atomic-setup regression was RED on the parent (`ERROR: column "subscribed" does not exist`) and GREEN here:

```text
PASS: partial partition setup is explicit and rejected by atomic setup
PASS: atomic setup preserves all post-setup slot histories
PASS: failed atomic setup rolls back all partial state
```

During release audit I also reproduced an unbounded-input abuse through APIs available to `pgque_reader`:

```sql
set statement_timeout = '250ms';
select pgque.subscribe_partitioned('bound_audit', 'huge', 100000);
-- ERROR: canceling statement due to statement timeout

select pgque.subscribe_slot('bound_audit', 'huge_slot', 0, 100000000);
-- succeeded before the ceiling
select * from pgque.partition_slot_status
where queue_name = 'bound_audit' and consumer = 'huge_slot';
-- ERROR: canceling statement due to statement timeout
```

The first transaction rolled back after timing out inside the per-slot setup loop. The second call persisted `n=100000000`; the status view then attempted its `generate_series` expansion and timed out. This established a resource-exhaustion risk rather than only an input-validation preference.

The 256 ceiling is a safety boundary, not a throughput target. On a local PostgreSQL 18 disposable instance, using one queue and consumer per sample and polling 10,000 events, `/usr/bin/time -p` recorded:

```text
N      setup   status   poll 10k
32     0.02s   0.02s    0.06s
128    0.02s   0.01s    0.14s
256    0.02s   0.01s    0.34s
512    0.03s   0.01s    1.17s
1024   0.04s   0.02s    2.69s
```

The benchmark command shape was:

```bash
/usr/bin/time -p psql --no-psqlrc "$PGQUE_TEST_DSN" --set=ON_ERROR_STOP=1 \
  --command="select pgque.subscribe_partitioned('bench_q', 'bench_n', N)"
/usr/bin/time -p psql --no-psqlrc "$PGQUE_TEST_DSN" --set=ON_ERROR_STOP=1 \
  --command="select count(*) from pgque.partition_slot_status where queue_name='bench_q' and consumer='bench_n'"
/usr/bin/time -p psql --no-psqlrc "$PGQUE_TEST_DSN" --set=ON_ERROR_STOP=1 \
  --file=/tmp/poll-10000-events.sql
```

The new boundary test was committed before the ceiling implementation and failed as intended:

```text
ERROR: expected subscribe_partitioned slot-limit rejection
```

It then passed after the implementation:

```text
PASS: partition slot count ceiling
```

That test proves exact `n=256` success, `n=257` rejection through both `subscribe_partitioned` and `subscribe_slot`, the exact diagnostic, and no partial `partition_consumer`, slot, subscription, or consumer state after rejection. The persisted table CHECK enforces the same range. Guidance is deliberately to choose the minimum N needed for actual worker parallelism; 256 is an emergency ceiling, not a recommendation to provision 256 slots.

RED/GREEN chronology: mixed. The original atomic-setup change is retrospective (independent parent RED/head GREEN replay, not test-first history). The later ceiling is chronological: RED commit `d05fbac` followed by GREEN commit `6db7b5b`.

Full CI: [17/17 checks passed](https://github.com/NikolayS/PgQue/actions/runs/29155331297), including PostgreSQL 14-19beta1, the stacked two-session harness, full regression, and acceptance coverage.

This comment records test evidence only; it is not a review or merge decision.

NikolayS:
# Review — design + empirical verification

Verified against fresh installs of `devel/sql/pgque.sql` (head `6db7b5b`) on PG 18.3: the #307 repro on both paths, concurrency at the subscription boundary, idempotency/cursor behavior, the n-cap constraint on simulated pre-cap data, alert semantics of the NULL lag, the PR's own test suites, and teardown behavior. **Verdict: the fix is correct and merge-worthy after one decision (finding 1). Findings 2–4 are follow-up material.**

## What's right

- **Correct fix, not a patch.** `subscribe_partitioned` registers all N slots at one shared tick in one transaction via the engine's existing `register_consumer_at` — engine untouched, failure rolls back the whole consumer. Confirmed: the #307 repro passes on the atomic path, and the sharpest case holds — a producer sending *during* an open `subscribe_partitioned` transaction loses nothing (5 concurrent keyed events delivered exactly once across the slot union).
- **Honesty over magic, consistently.** Partial setup raises instead of implying lost history is recoverable; the view reports `subscribed=false` + `pending_events=NULL` instead of the deceptive `0`; a repeated complete setup is a verified no-op that never repositions cursors (checked across intervening ticks); wrong N still hits the pinned-N error.
- Spec (D3/D7, §8, §17), the dedicated `test_partition_setup.sql` regression suite (including rollback-of-partial-state), acceptance, and error-message pointers are all updated coherently. All suites pass; CI green.

## Findings

**1. Pre-merge decision — the bundled `n ≤ 256` cap is an upgrade landmine.**
The install re-adds `partition_consumer_n_check` on every idempotent run. Simulated a pre-cap install holding a consumer with `n=300`: the re-install **aborts** at the constraint re-add (`check constraint … is violated by some row`, psql exit 3). Loud, no corruption — but it blocks the upgrade with no migration or guidance. Since this PR introduces the cap, it should own the consequence: a pre-flight guard (raise a clear "fix consumer X (n=300) before upgrading" message), or at minimum an explicit release-note line. Also a process note: the cap is a second logical change (own commits, own test file) riding in an "atomic setup" PR — fine at draft stage, but it's the part most likely to need independent revisiting.

**2. The NULL lag trades a lie for silence — alert recipes must change.**
Confirmed: with one slot unsubscribed, the naive threshold alert `… where pending_events > X` silently drops the NULL row — the incomplete slot doesn't appear at all. The #307 goal was "unmistakable"; for anyone monitoring by threshold it's now merely invisible-in-a-different-way. Docs/monitoring guidance should pin the canonical alert as `where pending_events > X or not subscribed`.

**3. Teardown asymmetry, now sharper.**
`unsubscribe_slot` on one slot of a complete consumer succeeds **silently** (no warning), leaving `partition_consumer` at n=4 with a hole — exactly the "incomplete setup" state `subscribe_partitioned` rejects. There is no `unsubscribe_partitioned`. Atomic create + silent partial destroy is a half-finished symmetry; the whole-consumer teardown is the natural sibling PR.

**4. "Repair" oversells — repair ≠ recovery.**
Confirmed on the retained `subscribe_slot` path: a late-subscribed slot starts at the current tick; the historical event stays unreachable. The error text's "or recreate it before producing" carries the real meaning, but user-facing docs should say it plainly: repair closes the gap going forward only; the missed window is gone.

## Stack note

Stacked on #321 ← #311 (both draft), so this result is conditional on those landing underneath as-is; nothing here conflicts with them.

*Methodology: scratch worktree at `6db7b5b`, fresh `createdb` installs, two-session concurrency via separate backends, constraint-hazard simulation by dropping/violating the check then re-running the full install file. All probes scripted and reproducible.*

PR #338, opened 2026-07-10
fix(clients): define nullable message behavior

NikolayS:
## Summary

- preserve SQL NULL honestly in Python, TypeScript, and Ruby message values/types
- retain Go's existing NULL-to-empty-string compatibility contract
- prevent NULL from dispatching to an illicit null/empty named handler
- define wildcard/unknown-policy behavior explicitly
- add real normal and cooperative receive coverage across all four clients

Fixes #316.

## Validation

- Go: `go test -race -count=1 ./...` passed
- Python: `81 passed`
- TypeScript: `89 passed`; type-check and build passed
- Ruby: `80 runs, 205 assertions, 0 failures, 0 errors`
- focused red/green tests covered public types, per-type dispatch, wildcard dispatch, and DB round trips
- `git diff --check` and language syntax checks passed

### DIFF
--- a/clients/go/README.md
+++ b/clients/go/README.md
@@ -97,6 +97,12 @@ func main() {
 | `WithUnknownHandlerPolicy(p)`           | `NackUnknown`  | `AckUnknown` logs and skips messages with no registered handler.      |
 | `WithRetryAfter(d time.Duration)`       | `60s`          | Retry delay for Consumer-issued `Nack` calls on handler failure or unknown type. |
 
+PgQ permits SQL `NULL` in `ev_type` and `ev_data` for direct SQL and trigger
+producers. To preserve the existing Go API, `Receive` and `ReceiveCoop`
+normalize either value to the empty string. An empty `Message.Type` never
+matches a handler (including one registered with `Handle("", ...)`); the
+high-level Consumer applies `UnknownHandlerPolicy` instead.
+
 ## Manual ticking
 
 For tests, demos, or manual operation without `pg_cron`, use
--- a/clients/go/consumer.go
+++ b/clients/go/consumer.go
@@ -152,7 +152,13 @@ func (c *Consumer) Start(ctx context.Context) error {
 		nackFailed := false
 		for _, msg := range msgs {
 			batchID = msg.BatchID
-			handler, ok := c.handlers[msg.Type]
+			// SQL NULL is normalized to "" by Receive. Keep it on the
+			// unknown-type path instead of dispatching an empty-string key.
+			var handler HandlerFunc
+			var ok bool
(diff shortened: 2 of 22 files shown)


NikolayS:
Real-user verification evidence

The original RED reproductions are retrospective because the first commit mixed implementation and tests. Final focused replays against base produced:

- Python: 2 failed, 3 passed; head: 5 passed.
- Ruby null dispatch: 4 runs, 11 assertions, 1 failure; head: 4 runs, 13 assertions, 0 failures.
- Go: the Consumer null-type dispatch test called the illicit empty-name handler on base; head routed through unknown/catch-all behavior.
- TypeScript: nullable type checks and Consumer null dispatch failed on base; head passed.

Head commands:

```sh
go test -race -count=1 ./...
PGQUE_TEST_DSN=postgresql:///pgque_pr_rules_audit python -m pytest -q
bun install --frozen-lockfile && bun run check && bun run test && bun run build
PGQUE_TEST_DSN=postgresql:///pgque_pr_rules_audit ruby -S rake test
```

Results: Go race suite passed; Python 81 passed; TypeScript 89 passed with typecheck/build; Ruby 80 runs, 205 assertions, 0 failures, 0 errors. Normal and cooperative database receive paths were exercised across all four clients. Head CI: 17/17 checks passed.

NikolayS (commented):
## REV rubric review

Automated ultrareview quota was unavailable, so I executed the five official
REV rubric prompts manually. SOC2 findings were omitted as instructed. This is
a comment-only review, not an approval.

### Blocking findings

None. I found no unresolved functional, security, documentation, or test
blocker on head `866e24ad8146c64aa6edd0eb9e177054bcc27eb9`.

### Rubric results

- Bug hunter: Python, Ruby, and TypeScript preserve SQL NULL explicitly. Go's
  NULL-to-empty-string behavior is an intentional compatibility constraint;
  empty types take the unknown-handler path consistently.
- Security reviewer: no SQL interpolation, privilege, deserialization, or
  secret-handling regression is introduced.
- Docs reviewer: every client documents its public nullable representation and
  unknown/catch-all behavior, including the intentional Go normalization.
- Guidelines checker: the cross-client scope is justified by one shared public
  contract and all CI checks pass. Immutable history: `test(python): run
  nullable consumer units without PostgreSQL` uses a disallowed type and its
  60-character subject exceeds the current limit. Do not amend or force-push.
- Test analyzer: normal and cooperative receives, public types, illicit
  null-handler registration, catch-all routing, and all four client behaviors
  are covered with live-Postgres and focused unit tests.

### Nonblocking findings

Only the immutable commit-type/length history noted above. No current code
change is requested.

